PR #1, opened 2019-02-11
WIP Neptune Client API

lukasz-walkiewicz:


neptune-client-api.md
```diff
@@ -0,0 +1,56 @@
...
+```python
+exp.send_metric("ROC_AUC", 0.75)
+exp.send_metric("ROC_AUC_STD", 0.002)
+exp.send_text("log", "done")
+exp.send_image("false", "/tmp/failed_class.png")
+```
+
+
+```python
+exp.set_tags(exp.get_tags().append("new-tag")) # update tags
```
jakubczakon: @lukasz-walkiewicz Could we go even simpler here?

```python
exp.tags.append('new-tag')
```

neptune-client-api.md
```diff
@@ -0,0 +1,56 @@
...
+    handle_uncaught_exceptions=False # default = True
+)
+```
+
+
+```python
+exp.send_metric("ROC_AUC", 0.75)
+exp.send_metric("ROC_AUC_STD", 0.002)
+exp.send_text("log", "done")
+exp.send_image("false", "/tmp/failed_class.png")
```
jakubczakon: @lukasz-walkiewicz 
Those are awesome, I really like it.

neptune-client-api.md
```diff
@@ -0,0 +1,56 @@
...
+exp = project.create_experiment( # create experiment, with params and tags
+    params={
+    "lr__C": 1000,
+    "epochs": 10,
+    "shuffle": true
+}, 
+    tags=["solution-6", "stacking"],
+    send_hardware_metrics=False, # default = True
+    run_monitoring_thread=False, # default = True
+    upload_sources_files=["main.py"], # default = this file
```
jakubczakon: @lukasz-walkiewicz I am wondering, whether some kind of magic auto discovery could be possible here.
In case of a more elaborate project, listing everything would not really be feasible.
Sacred has something like this https://sacred.readthedocs.io/en/latest/collected_information.html#source-code

neptune-client-api.md
```diff
@@ -0,0 +1,56 @@
+
+
+```python
+from neptune.session import Session
+
+session = Session("token=")
```
jakubczakon: @lukasz-walkiewicz I think, from the user perspective if we could join the session/project/exp into a single constructor it would gain some extra usability points. 

I see the point though. With this I can query this project for things (like in the neptune-lib now) so I see why we need this object. 

How do you feel about having an ugly wrapper function:

```python
exp = create_experiment(token=,
                                         project=,
                                         params=,
                                         ...
                                         )
```

that under the hood contructs session and project objects?

neptune-client-api.md
lukasz-walkiewicz: Yes, we could but I think it would be a bit misleading for the user that `append` method makes a HTTP request to Neptune under the hood. Furthermore, we would have to implement the rest of list manipulating methods such as: `sort` and `reverse` which, IMO, makes little sense.

neptune-client-api.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+```python
+project = session.get_project("neptune-ml/Ships")
+```
+
+
+```python
+exp = project.create_experiment( # create experiment, with params and tags
+    params={
```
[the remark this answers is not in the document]
lukasz-walkiewicz: Currently: no. Experiment's parameters are immutable.

neptune-client-api.md
lukasz-walkiewicz: We've changed that part recently. Would you please take another look?

neptune-client-api.md
lukasz-walkiewicz: Looks interesting, I'll look into it.

PR #3, opened 2019-02-13
Send hardware metrics

dominik-jastrzebski:


neptune/utils.py
```diff
@@ -15,10 +15,14 @@
 #
 
 import functools
+import sys
 
 import numpy as np
 import pandas as pd
 
+# http://stackoverflow.com/questions/1325581/how-do-i-check-if-im-running-on-windows-in-python
```
lukasz-walkiewicz: Is that comment necessary?

PR #9, opened 2019-02-15
Terminate experiment when it's aborted in UI

dominik-jastrzebski:


neptune/internal/abort.py
```diff
@@ -0,0 +1,88 @@
...
+    def abort(self):
+        self.__runnable()
+
+
+class DefaultAbortImpl(object):
+    KILL_TIMEOUT = 5
+
+    def __init__(self, pid):
+        self._pid = pid
+        self._logger = logging.getLogger(__name__)
```
lukasz-walkiewicz: Remove all `logging` package usages

neptune/internal/websockets/reconnecting_websocket.py
```diff
@@ -0,0 +1,118 @@
...
+            except Exception:
+                if self._is_active():
+                    self._handle_lost_connection()
+
+    def _is_active(self):
+        return not self._shutdown_event.is_set()
+
+    def _on_successful_connect(self):
+        if self._reconnect_counter.retries >= 1:
+            print(CONNECTION_RESTORED_MESSAGE)
```
lukasz-walkiewicz: I think we shouldn't print anything in these kind of situations

PR #12, opened 2019-02-19
Simplify API

lukasz-walkiewicz:


neptune/__init__.py
```diff
@@ -13,3 +13,87 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from neptune import projects, experiments
+from neptune.internal.structs.stack import Stack
+from neptune.sessions import Session
+
+session = None
```
pitercl: Locks, please.

neptune/experiments.py
```diff
@@ -40,7 +41,7 @@ class Experiment(object):
         Instantiate a session.
 
         >>> from neptune.session import Session
-        >>> session = Session()
+        >>> current_session = Session()
```
pitercl: Why make it more complex?

neptune/experiments.py
```diff
@@ -439,3 +442,26 @@ def _create_channel(self, channel_name, channel_type):
         channel = self._client.create_channel(self.internal_id, channel_name, channel_type)
         self._leaderboard_entry.add_channel(channel)
         return channel
+
+
+current_experiment = None
+
+_run_stack = Stack()
```
pitercl: `_experiment_stack` maybe?

neptune/internal/structs/stack.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+class Stack(object):
```
pitercl: What's wrong with using a list? Why do we need this wrapper?

neptune/experiments.py
```diff
@@ -439,3 +442,26 @@ def _create_channel(self, channel_name, channel_type):
...
+
+
+def push_new_experiment(new_experiment):
```
pitercl: Locks, please.

neptune/projects.py
```diff
@@ -267,7 +267,7 @@ def get_experiment_groups(self):
 
     # pylint:disable=unused-argument
     def create_experiment(self,
-                          name,
+                          name="Untitled",
```
pitercl: It seems that this default name should be in a single place.

neptune/__init__.py
```diff
@@ -13,3 +13,87 @@
...
+                      properties=None,
+                      tags=None,
+                      upload_source_files=None,
+                      abort_callback=None,
+                      send_hardware_metrics=True,
+                      run_monitoring_thread=True,
+                      handle_uncaught_exceptions=True):
+    # pylint: disable=global-statement
+    global project
+    return experiments.push_new_experiment(
```
pitercl: `push_new_experiment` is also called in `project.create_exp`.

neptune/sessions.py
```diff
@@ -64,6 +65,15 @@ def get_project(self, project_qualified_name):
             name=project.name
         )
 
+    def get_default_project(self, organization_name=None):
```
pitercl: Why do we need this in `Session`'s API? I'd drop it.

neptune/__init__.py
pitercl: I know that assignment is atomic, but you don't just assign:

```
if session is None:
        session = init(project_qualified_name=project_qualified_name)
```

neptune/sessions.py
lukasz-walkiewicz: `get_default_project` is called when user initializes session without `project_qualified_name` argument. I can drop it but it that case we'll need to find another solution to support flow like:
```python
import neptune

neptune.init()
neptune.create_experiment(...)
```

neptune/internal/structs/stack.py
lukasz-walkiewicz: IMO it looks better this way but I can use plain list if you want me to.

pitercl: I'd prefer less code than more.

neptune/sessions.py
pitercl: I think it's better to inline it in `init()`. We don't want to expose such functionality, it's here only to facilitate the simplest-possible API.

PR #15, opened 2019-02-20
Add generic API exceptions mappings

dominik-jastrzebski:


neptune/exceptions.py
```diff
@@ -14,5 +14,22 @@
 # limitations under the License.
 #
 
-class InvalidApiToken(Exception):
-    pass
+
+class ConnectionLost(Exception):
```
lukasz-walkiewicz: I think that it would be more convenient to have common parent for all of these exceptions, something a bit more specific than `Exception`

PR #21, opened 2019-02-21
Map experiment creation api exceptions

dominik-jastrzebski:


neptune/client.py
```diff
@@ -195,8 +196,17 @@ def create_experiment(self, project, name, description, params, properties, tags
...
+            error_type = error_response.get('type')
+            if error_type == 'DUPLICATE_PARAMETER':
+                raise DuplicateParameter()
+            elif error_type == 'INVALID_TAG':
+                raise InvalidTag(error_response.get('message'))
+            else:
+                raise
         except HTTPUnprocessableEntity as e:
-            if e.response.json()['type'] == 'LIMIT_OF_EXPERIMENTS_IN_PROJECT_REACHED':
+            if e.response.json().get('type') == 'LIMIT_OF_EXPERIMENTS_IN_PROJECT_REACHED':
```
hubertjaworskineptune: e.response.json() might fail when response is not json (might happen during service downtime), add a private function that extracts type from a response objects (we might want to double check that http code is 4XX)

PR #25, opened 2019-02-25
Implement update_tags

lukasz-walkiewicz:


neptune/client.py
```diff
@@ -214,6 +214,24 @@ def create_experiment(self, project, name, description, params, properties, tags
...
+        try:
+            self.backend_swagger_client.api.updateTags(
+                updateTagsParams=UpdateTagsParams(
+                    experimentIds=[experiment.internal_id],
+                    groupsIds=[],
+                    tagsToAdd=tags_to_add,
+                    tagsToDelete=tags_to_delete
+                )
+            ).response().result
+        except HTTPNotFound:
```
dominik-jastrzebski: I think you should map BadRequest (400). Tags are (or should be) validated before updating. You can use code from experiment creation to provide the same error message.

PR #30, opened 2019-02-26
Add tags and properties handling

lukasz-walkiewicz:
* Implement properties setting during experiment execution
* Change tag manipulation methods
* Send exception value in traceback for failed experiment

### DIFF
--- a/neptune/client.py
+++ b/neptune/client.py
@@ -197,7 +197,7 @@ class Client(object):
             experiment = self.backend_swagger_client.api.createExperiment(
                 experimentCreationParams=params).response().result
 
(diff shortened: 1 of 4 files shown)


neptune/client.py
```diff
@@ -558,35 +548,36 @@ def _convert_to_api_properties(self, raw_properties):
             ) for key, value in raw_properties.items()
         ]
 
-    def _convert_experiment_to_leaderboard_entry(self, experiment):
+    def _convert_to_experiment(self, experiment):
```
dominik-jastrzebski: Maybe rename it to `api_experiment`? `convert_to_experiment(experiment)` may be confusing.

PR #32, opened 2019-03-01
Don't cache mutable data in experiment object

lukasz-walkiewicz:


neptune/client.py
```diff
@@ -214,24 +217,24 @@ def create_experiment(self, project, name, description, params, properties, tags
             else:
                 raise
 
+    @with_api_exceptions_handler
+    def get_experiment(self, experiment_id):
+        return self.backend_swagger_client.api.getExperiment(experimentId=experiment_id).response().result
+
     @with_api_exceptions_handler
     def update_experiment(self, experiment, properties):
```
aniezurawski: 1. `experiment` argument is used only to get its `internal_id` and as return value. Let's just pass `experiment_id` instead.
2. How do we update tags, name, description, etc.?
3. Shouldn't this method be called `update_properties`?

neptune/experiments.py
```diff
@@ -61,9 +61,11 @@ class Experiment(object):
         Column sorting
     """
 
-    def __init__(self, client, leaderboard_entry):
+    def __init__(self, client, _id, internal_id, project_full_id):
```
aniezurawski: Why is there an underscore in `_id` argument name?

aniezurawski: Project name can change. Shouldn't we store `uuid` instead?

neptune/experiments.py
```diff
@@ -129,36 +130,43 @@ def system_properties(self):
...
+        experiment = self._client.get_experiment(self._internal_id)
+        return self._simple_dict_to_dataframe({
+            'id': experiment.shortId,
+            'name': experiment.name,
+            'created': experiment.timeOfCreation,
+            'finished': experiment.timeOfCompletion,
+            'running_time': experiment.runningTime,
+            'owner': experiment.owner,
+            'size': experiment.storageSize,
+            'tags': experiment.tags,
```
aniezurawski: Why are tags here but properties are not? What's `system_property` and what is not?

neptune/experiments.py
```diff
@@ -129,36 +130,43 @@ def system_properties(self):
...
+            'notes': experiment.description
+        })
+
+    def get_tags(self):
+        return self._client.get_experiment(self._internal_id).tags
 
     def append_tag(self, tag):
```
aniezurawski: It would be nice if we could pass a list of tags here.

neptune/experiments.py
```diff
@@ -129,36 +130,43 @@ def system_properties(self):
...
         self._client.update_tags(experiment=self,
                                  tags_to_add=[tag],
                                  tags_to_delete=[])
-        self._leaderboard_entry.tags.append(tag)
 
     def remove_tag(self, tag):
```
aniezurawski: Same here.

neptune/experiments.py
```diff
@@ -294,13 +311,15 @@ def parameters(self):
 
             Get experiment parameters.
 
-            >>> experiment.parameters
+            >>> experiment.get_parameters()
 
         """
-        return self._simple_dict_to_dataframe(self._leaderboard_entry.parameters)
+        experiment = self._client.get_experiment(self.internal_id)
+        return dict(
```
aniezurawski: Should we return dict or DataFrame here?

neptune/experiments.py
```diff
@@ -323,21 +342,24 @@ def properties(self):
 
             Get experiment properties.
 
-            >>> experiment.properties
+            >>> experiment.get_properties
 
         """
-        return self._simple_dict_to_dataframe(self._leaderboard_entry.properties)
+        experiment = self._client.get_experiment(self.internal_id)
+        return dict(
```
aniezurawski: Should we return dict or DataFrame here?

neptune/model.py
```diff
@@ -78,12 +78,10 @@ def system_properties(self):
             'created': entry.timeOfCreation,
             'finished': entry.timeOfCompletion,
             'running_time': entry.runningTime,
-            'source_code_size': entry.sourceSize,
             'owner': entry.owner,
             'size': entry.size,
             'tags': entry.tags,
-            'notes': entry.description,
-            'git_hash': entry.commitId
```
aniezurawski: Why was `sourceSize` and `commitId` removed?

neptune/client.py
lukasz-walkiewicz: 1. It's not. `experiment.project_full_id` is used in 404 (NotFound Error) handling.
2. Tags: there is a separate method below, `update_tags`. Name, description: currently, we don't.
3. It's a wrapper for `updateExperiment` api method so I think that current method name looks better than `updateProperties`.

neptune/experiments.py
lukasz-walkiewicz: `_id` is a built-in function in python: https://docs.python.org/3/library/functions.html#id

neptune/experiments.py
lukasz-walkiewicz: List of fields included in system properties was previously defined in: https://github.com/neptune-ml/neptune-client/blob/master/neptune/model.py#L73 and this change use the same set of fields.

neptune/experiments.py
lukasz-walkiewicz: It's an interface defined by Jakub. He said that it's more convenient for a typical user to add and remove tags one by one.

neptune/experiments.py
lukasz-walkiewicz: Fixed

neptune/experiments.py
lukasz-walkiewicz: Fixed

neptune/client.py
aniezurawski: 3. But it updates only properties.

neptune/experiments.py
lukasz-walkiewicz: `project_full_id` is used only during api errors handling and project name will rarely change so I don't think it's worth it.

neptune/experiments.py
lukasz-walkiewicz: https://github.com/neptune-ml/neptune-client/pull/32#discussion_r262002105

neptune/model.py
lukasz-walkiewicz: It was more complicated to obtain these two fields from backend api than it was from leaderboard, so we decided to drop it, at least for now.

neptune/experiments.py
aniezurawski: So `def project_full_id(self):` property should not be exposed in Experiment.

PR #38, opened 2019-03-05
Send channels values in a separate thread

lukasz-walkiewicz:


neptune/experiments.py
```diff
@@ -474,6 +477,11 @@ def get_numeric_channels_values(self, *channel_names):
         return align_channels_on_x(pd.concat(channels_data.values(), axis=1, sort=False))
 
     def stop(self, exc_tb=None):
+        if self._channels_values_sending_thread:
+            self._channels_values_sending_thread.interrupt()
```
aniezurawski: I'm not sure if thread per experiment is a good idea.

neptune/experiments.py
```diff
@@ -500,6 +508,8 @@ def stop(self, exc_tb=None):
         if self._stderr_uploader:
             self._stderr_uploader.close()
 
+        sys.excepthook = sys.__excepthook__
```
aniezurawski: `pop_stopped_experiment` should be responsible for setting this variable when last experiment was popped, shouldn't it?

neptune/internal/channels/channels.py
```diff
@@ -0,0 +1,83 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import time
+
+
+class ChannelWithValue(object):
```
aniezurawski: This class seems to be useless since it's used only in `Experiment` implementation to pass value from one method to another. Moreover, its name is misleading. It does not contain a channel and a value. It's some tuple containing a name, a type and some other variables.
Just pass `channel_name`, `channel_type` and `value` of type `ChannelValue` separately in `Experiment` implementation.

neptune/internal/channels/channels.py
```diff
@@ -0,0 +1,83 @@
...
+    @property
+    def y(self):
+        return self._y
+
+    @property
+    def t(self):
+        return self._t
+
+
+class ChannelWithValues(object):
```
aniezurawski: `ChannelIdWithValues`

neptune/internal/channels/channels.py
```diff
@@ -0,0 +1,83 @@
...
+
+class ChannelValue(object):
+
+    def __init__(self, t, x, y):
+        self._t = t
+        self._x = x
+        self._y = y
+
+    @property
+    def t(self):
```
aniezurawski: Please. Call it `ts` at least if `timestamp` is too long.

neptune/projects.py
```diff
@@ -286,6 +288,14 @@ def exception_handler(exc_type, exc_val, exc_tb):
         if handle_uncaught_exceptions:
             sys.excepthook = exception_handler
 
+        # pylint:disable=protected-access
+        experiment._channels_values_queue = queue.Queue()
+        # pylint:disable=protected-access
+        experiment._channels_values_sending_thread = ChannelsValuesSendingThread(
+            experiment, experiment._channels_values_queue)
+        # pylint:disable=protected-access
```
aniezurawski: You know something is going wrong if this line appears too many times.

PR #42, opened 2019-03-06
Disable stdout & stderr upload in notebook

lukasz-walkiewicz:


neptune/projects.py
```diff
@@ -297,11 +296,11 @@ def exception_handler(exc_type, exc_val, exc_tb):
                 websocket_factory=websocket_factory, abort_impl=abort_impl, experiment_id=experiment.internal_id)
             experiment._aborting_thread.start()
 
-        if upload_stdout:
+        if upload_stdout and not is_notebook():
```
hubertjaworskineptune: Since is_notebook handled by exception, it would be best if we didn't call it multiple times

neptune/utils.py
```diff
@@ -78,6 +78,13 @@ def in_docker():
     cgroup_file = '/proc/self/cgroup'
     return os.path.exists('./dockerenv') or (os.path.exists(cgroup_file) and file_contains(cgroup_file, text='docker'))
 
+def is_notebook():
```
hubertjaworskineptune: is_run_in_notebook?
Experiment is not a notebook, it may be <in> notebook

PR #44, opened 2019-03-07
Improve channels values sending thread

lukasz-walkiewicz:
* Sleep appropriate time when values queue was empty
  * Stop gracefully on `join()`
  * Log errors using logging module (on DEBUG level)

### DIFF
--- a/neptune/internal/channels/channels_values_sender.py
+++ b/neptune/internal/channels/channels_values_sender.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import logging
 import threading
 import time
 from collections import namedtuple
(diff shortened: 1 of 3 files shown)


neptune/internal/channels/channels_values_sender.py
```diff
@@ -76,27 +79,24 @@ def __init__(self, experiment, values_queue):
...
 
     def join(self, timeout=None):
         self.interrupt()
-        while not self._values_queue.empty():
-            self._values_batch.append(self._values_queue.get())
-            self._values_queue.task_done()
-        self._process_batch()
+        super(ChannelsValuesSendingThread, self).join(timeout)
 
     def _process_batch(self):
```
pitercl: This can be called from `run` and `join` and is not synchronized.

PR #49, opened 2019-03-07
Fix comments in Experiment

aniezurawski:


lukasz-walkiewicz:
I think you should delete also `_simple_dict_to_dataframe` method from `Experiment`

PR #59, opened 2019-03-13
Allow user to ignore certificate validation

hubertjaworskineptune:


neptune/sessions.py
```diff
@@ -50,11 +50,11 @@ class Session(object):
         >>> session = Session()
     """
 
-    def __init__(self, api_token=None):
+    def __init__(self, api_token=None, verify_ssl=False):
```
aniezurawski: This method is exposed to user do let's call this argument `insecure` (it seems to be very common) with a default value `None` and let's check for existence of `NEPTUNE_INSECURE` environment variable if the value is `None`.
Moreover, we should add such argument to `neptune.init` too.

PR #62, opened 2019-03-13
Send image channel values in smaller batches

lukasz-walkiewicz:


neptune/internal/channels/channels_values_sender.py
```diff
@@ -87,7 +88,10 @@ def run(self):
             except queue.Empty:
                 self._sleep_time = 0
 
-            if self._sleep_time <= 0 or len(self._values_batch) >= self._MAX_VALUES_BATCH_LENGTH:
+            if self._sleep_time <= 0 \
+                    or len(self._values_batch) >= self._MAX_VALUES_BATCH_LENGTH \
+                    or len([v for v in self._values_batch if
+                            v.channel_type == 'image']) >= self._MAX_IMAGE_VALUES_BATCH_LENGTH:  # pylint:disable=line-too-long
```
dominik-jastrzebski: Is the pylint suppression really necessary? We should have 120 chars line limit.

lukasz-walkiewicz: It was hard to break that line without loosing readability.

PR #65, opened 2019-03-15
Test ChannelsValuesSender

lukasz-walkiewicz:


neptune/internal/channels/channels.py
```diff
@@ -15,11 +15,18 @@
 #
 import time
 from collections import namedtuple
+from enum import Enum
 
 ChannelNameWithType = namedtuple("ChannelNameWithType", ['channel_name', 'channel_type'])
 ChannelIdWithValues = namedtuple('ChannelIdWithValues', ['channel_id', 'channel_values'])
 
 
+class ChannelType(Enum):
```
dominik-jastrzebski: `enum` is in Python 3.4+ only. Can we break Python 2.x compatibility?

tests/neptune/internal/channels/test_channels_values_sender.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def setUp(self):
+        # pylint: disable=protected-access
+        self._EXPERIMENT._get_channels.return_value = self._CHANNELS
+
+    def tearDown(self):
+        self._EXPERIMENT.reset_mock()
+
+    def test_send_values_on_join(self):
+        # GIVEN
```
dominik-jastrzebski: Do we really need to use uppercase GIVEN/WHEN/THEN?

PR #71, opened 2019-03-26
Allow proxies usage.

zmonoid:
Allow to use proxies. Experimented pass.

Simple example to verify:

1. Create local dynamic port forwarding:
```
ssh -D 9999 user@localhost
```
2. Run the following:
```python
if __name__ == '__main__':
    proxies = {"http": "socks5h://127.0.0.1:9999", 'https': 'socks5h://127.0.0.1:9999'}
    neptune.init(project_qualified_name='namespace/project', proxies=proxies)
    neptune.create_experiment()
    neptune.send_metric('Test', 0)


```

### DIFF
--- a/neptune/__init__.py
+++ b/neptune/__init__.py
@@ -26,7 +26,7 @@ project = None
 __lock = threading.RLock()
 
 
-def init(project_qualified_name=None, api_token=None):
+def init(project_qualified_name=None, api_token=None, proxies={}):
     if project_qualified_name is None:
         project_qualified_name = os.getenv(envs.PROJECT_ENV_NAME)
 
@@ -34,7 +34,7 @@ def init(project_qualified_name=None, api_token=None):
     with __lock:
         global session, project
 
-        session = Session(api_token=api_token)
+        session = Session(api_token=api_token, proxies=proxies)
 
         if project_qualified_name is None:
             raise MissingProjectQualifiedName()
--- a/neptune/__init__.py
+++ b/neptune/__init__.py
@@ -26,7 +26,7 @@ project = None
 __lock = threading.RLock()
 
 
-def init(project_qualified_name=None, api_token=None, proxies={}):
+def init(project_qualified_name=None, api_token=None, proxies=None):
     if project_qualified_name is None:
         project_qualified_name = os.getenv(envs.PROJECT_ENV_NAME)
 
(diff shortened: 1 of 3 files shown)


zmonoid:
Update string formating

neptune/__init__.py
```diff
@@ -26,15 +26,15 @@
 __lock = threading.RLock()
 
 
-def init(project_qualified_name=None, api_token=None):
+def init(project_qualified_name=None, api_token=None, proxies={}):
```
pitercl: Please, go with `proxies=None` and in the method body do
```python
if proxies is None:
    proxies = {}
```

Using a dictionary as default parameter value is a [no-no](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html) in Python :)

neptune/__init__.py
```diff
@@ -26,15 +26,15 @@
...
     if project_qualified_name is None:
         project_qualified_name = os.getenv(envs.PROJECT_ENV_NAME)
 
     # pylint: disable=global-statement
     with __lock:
         global session, project
 
-        session = Session(api_token=api_token)
+        session = Session(api_token=api_token, proxies=proxies)
```
pitercl: Same here.

neptune/client.py
```diff
@@ -74,16 +74,17 @@ def wrapper(*args, **kwargs):
 class Client(object):
 
     @with_api_exceptions_handler
-    def __init__(self, api_address, api_token):
+    def __init__(self, api_address, api_token, proxies={}):
```
pitercl: Same here.

neptune/sessions.py
```diff
@@ -50,11 +50,13 @@ class Session(object):
         >>> session = Session()
     """
 
-    def __init__(self, api_token=None):
+    def __init__(self, api_token=None, proxies={}):
```
pitercl: Same here.

neptune/client.py
```diff
@@ -110,6 +111,15 @@ def __init__(self, api_address, api_token):
         )
         self._http_client.authenticator = self.authenticator
 
+    @with_api_exceptions_handler
```
pitercl: I don't think we need this decorator here, since the method doesn't make any calls.

neptune/__init__.py
zmonoid: I put instead under Client.__init__():
```python
if proxies is not None:
    self._update_proxies()
```

pitercl:
@zmonoid Thanks for the PR!

zmonoid:
@pitercl Glad to help >.<

PR #74, opened 2019-03-28
Add ability to send git commit info

hubertjaworskineptune:


neptune/internal/utils/git_info.py
```diff
@@ -0,0 +1,68 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import sys
+import subprocess
+
+
+class GitInfo(object):
```
pitercl: This is not an internal/util class - it's our API.

neptune/__init__.py
```diff
@@ -73,12 +75,21 @@ def create_experiment(name=None,
     global project
     if project is None:
         raise Uninitialized()
+
+    # pylint: disable=bare-except
+    if git_info is None:
+        try:
```
pitercl: This try/except is unnecessary - there's already one in the implementation. If the implementation guarantees that it doesn't throw, we should rely on this guarantee in the rest of the code.

neptune/projects.py
```diff
@@ -242,6 +244,12 @@ def create_experiment(self,
         if tags is None:
             tags = []
 
+        if git_info is None:
+            try:
```
pitercl: Unnecessary try/except.

neptune/internal/utils/git_info.py
pitercl: Perhaps a params validation is in order? Will the user be able to tell what went wrong after putting wrong values in this object?

neptune/internal/utils/git_info.py
```diff
@@ -0,0 +1,68 @@
...
+            try:
+                with open(os.devnull, 'w') as devnull:
+                    return subprocess.check_output(['git', '--version'], stderr=devnull).decode("utf-8").strip()
+            except:  # pylint: disable=bare-except
+                return None
+
+        if not get_git_version():
+            return None
+
+        if hasattr(sys, 'getwindowsversion') and r'GIT_PYTHON_GIT_EXECUTABLE' not in os.environ:
```
pitercl: Remove this, please. If our users want, they can overwrite the env themselves.

neptune/internal/utils/git_info.py
```diff
@@ -0,0 +1,68 @@
...
+        import git
+
+        def get_git_version():
+            try:
+                with open(os.devnull, 'w') as devnull:
+                    return subprocess.check_output(['git', '--version'], stderr=devnull).decode("utf-8").strip()
+            except:  # pylint: disable=bare-except
+                return None
+
+        if not get_git_version():
```
pitercl: What is the purpose of this? We already catch every exception.

neptune/internal/utils/git_info.py
```diff
@@ -0,0 +1,68 @@
...
+            return None
+
+        if hasattr(sys, 'getwindowsversion') and r'GIT_PYTHON_GIT_EXECUTABLE' not in os.environ:
+            os.environ[r'GIT_PYTHON_GIT_EXECUTABLE'] = os.popen("where git").read().strip()
+
+        repository_path = os.getcwd()
+
+        try:
```
pitercl: Why nest try/except??

neptune/internal/utils/git_info.py
```diff
@@ -0,0 +1,68 @@
...
+            except:  # pylint: disable=bare-except
+                return None
+
+        if not get_git_version():
+            return None
+
+        if hasattr(sys, 'getwindowsversion') and r'GIT_PYTHON_GIT_EXECUTABLE' not in os.environ:
+            os.environ[r'GIT_PYTHON_GIT_EXECUTABLE'] = os.popen("where git").read().strip()
+
+        repository_path = os.getcwd()
```
pitercl: Is this always the case? the CWD doesn't have to have anything to do with where the repo is.

neptune/projects.py
```diff
@@ -214,6 +215,7 @@ def create_experiment(self,
                           params=None,
                           properties=None,
                           tags=None,
+                          git_info=None,
```
pitercl: Move to the end of param list

tests/neptune/test_git_info.py
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+#
+
+import unittest
+
+from neptune.internal.utils.git_info import get_git_info
+
+
+class TestGitInfo(unittest.TestCase):
+    def test_git_info(self):
```
pitercl: This is one possible test. Another should mock the `git` module.

neptune/__init__.py
pitercl: Also, why would we call `get_git_info` in two places?

neptune/internal/utils/git_info.py
hubertjaworskineptune: What better option do you suggest? please keep in mind that this is an internal implementation for default git info - user is not supposed to call it manually

PR #98, opened 2019-05-15
Add download_artifact method to experiment

aniezurawski:


tests/neptune/test_experiment.py
```diff
@@ -39,7 +39,13 @@ class TestExperiment(unittest.TestCase):
     def test_send_metric(self, ChannelsValuesSender):
         # given
         channels_values_sender = ChannelsValuesSender.return_value
-        experiment = Experiment(mock.MagicMock(), an_experiment_id(), a_uuid_string(), a_project_qualified_name())
+        experiment = Experiment(
+            mock.MagicMock(),
+            mock.MagicMock(),
```
lukasz-walkiewicz: Use (and create if necessary) `a_project` method. We should avoid using mocks is such cases.

lukasz-walkiewicz: x N

PR #108, opened 2019-06-04
Add reset_log method to experiment

aniezurawski:


neptune/experiments.py
```diff
@@ -359,6 +374,12 @@ def send_graph(self, graph_id, value):
 
         self._client.put_tensorflow_graph(self, graph_id, value)
 
+    def reset_log(self, log_name):
+        channel = self._find_channel(log_name, ChannelNamespace.USER)
+        if channel is None:
+            raise ChannelDoesNotExist(self.id, log_name)
+        self._client.reset_channel(channel.id)
```
hubertjaworskineptune: You can add a try-catch block to change exception to a more user-friendly exception when backend does not resolve channel id

neptune/client.py
```diff
@@ -455,6 +455,17 @@ def create_channel(self, experiment, name, channel_type):
         except HTTPConflict:
             raise ChannelAlreadyExists(channel_name=name, experiment_short_id=experiment.id)
 
+    @with_api_exceptions_handler
+    def reset_channel(self, channel_id):
```
hubertjaworskineptune: We should try to let users know wchich channel is "bad", and User never operates on UUIDs

aniezurawski: `Client` is an internal class. User should not user it directly.

neptune/experiments.py
aniezurawski: It should never be a case since this id was just fetched from backend.

PR #144, opened 2019-08-01
Increase max retry count

aniezurawski:


neptune/client.py
```diff
@@ -49,16 +50,19 @@
 from neptune.utils import is_float
 
 
+_logger = logging.getLogger(__name__)
+
 def with_api_exceptions_handler(func):
     def wrapper(*args, **kwargs):
-        for retry in range(0, 10):
+        for retry in range(0, 11):
```
pitercl: @aniezurawski The retries should handle lack of connection as well as http errors.

PR #169, opened 2019-09-09
Raise ValueError in case of invalid argument types

aniezurawski:


neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -190,6 +190,17 @@ def create_experiment(self,
                           git_info,
                           hostname,
                           notebook_id):
+        if not isinstance(name, six.string_types):
+            raise ValueError("name should be string")
```
lukasz-walkiewicz: Invalid name `{}`, should be a string.

neptune/internal/utils/image.py
```diff
@@ -40,6 +40,7 @@ def get_image_content(image):
                 return _get_pil_image_data(Image.fromarray(array2d.astype(numpy.uint8)))
             if shape[2] in (3, 4):
                 return _get_pil_image_data(Image.fromarray(image.astype(numpy.uint8)))
+        raise ValueError("Incorrect size of numpy.ndarray.")
```
lukasz-walkiewicz: Not enough debug information

neptune/internal/backends/hosted_neptune_backend.py
lukasz-walkiewicz: Same for all value errors below.

PR #172, opened 2019-09-12
Create an empty checkpoint on `create_experiment`

lukasz-walkiewicz:


neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -315,20 +315,29 @@ def create_notebook(self, project):
...
+                                       }) as response:
+                if response.status_code == NOT_FOUND:
+                    raise NotebookNotFound(notebook_id=notebook_id)
+                else:
+                    response.raise_for_status()
+                    CheckpointDTO = self.leaderboard_swagger_client.get_model('CheckpointDTO')
+                    return CheckpointDTO.unmarshal(response.json())
+        else:
+            NewCheckpointDTO = self.leaderboard_swagger_client.get_model('NewCheckpointDTO')
+            return self.leaderboard_swagger_client.api.createCheckpoint2(
```
pitercl: Can we please rename this to something more meaningful than `..2`? `createIncorporealCheckpoint` (as in "no body") maybe?

pitercl: I'm actually wondering if we shouldn't leave the old method there as well for the long run..

neptune/projects.py
```diff
@@ -414,6 +426,24 @@ def _get_experiment_link(self, experiment):
             exp_id=experiment.id
         )
 
+    def _create_checkpoint(self, experiment_id, notebook_id, notebook_path):
+        try:
+            from ipykernel.comm import Comm
+
+            checkpoint = self._backend.create_checkpoint(notebook_id, notebook_path)
+
+            neptune_comm = Comm(target_name='neptune_comm')
```
pitercl: Does calling this multiple times work?

neptune/projects.py
```diff
@@ -414,6 +426,24 @@ def _get_experiment_link(self, experiment):
...
+            neptune_comm.send(data=dict(
+                message_type="CHECKPOINT_CREATED",
```
pitercl: Please extract the protocol somewhere.

neptune/projects.py
```diff
@@ -414,6 +426,24 @@ def _get_experiment_link(self, experiment):
             exp_id=experiment.id
         )
 
+    def _create_checkpoint(self, experiment_id, notebook_id, notebook_path):
+        try:
+            from ipykernel.comm import Comm
+
+            checkpoint = self._backend.create_checkpoint(notebook_id, notebook_path)
```
pitercl: What about not creating multiple checkpoints in a single notebook cell?

neptune/internal/comm/comm.py
```diff
@@ -0,0 +1,45 @@
+#
```
pitercl: Why isn't this file in internal/notebooks?

neptune/internal/comm/comm.py
```diff
@@ -0,0 +1,45 @@
...
+
+from neptune.internal.comm.message_type import MessageType
+
+_logger = logging.getLogger(__name__)
+
+_neptune_comm = None
+_neptune_comm_lock = threading.Lock()
+
+
+def send_checkpoint_created(experiment_id, notebook_id, notebook_path, checkpoint_id):
```
pitercl: Please comment here that this throws if ipykernel is not available.

neptune/projects.py
```diff
@@ -388,6 +395,12 @@ def create_experiment(self,
             notebook_id=notebook_id
         )
 
+        if notebook_id is not None and notebook_path is not None:
+            create_checkpoint(backend=self._backend,
```
pitercl: I just noticed this, but it should be the other way around - it's the experiment that should get the checkpoint_id, not checkpoint getting experiment_id.

I'd treat it similarly to source files.

PR #173, opened 2019-09-16
Support wildcards when uploading experiment source files

aniezurawski:


neptune/projects.py
```diff
@@ -336,6 +340,21 @@ def create_experiment(self,
                                           description='neural net trained on MNIST',
                                           upload_source_files=[])
 
+                # Send all python source files
```
pitercl: Those are just py files in CWD.

pitercl: Also, this probably won't see files starting with a dot, will it?

neptune/projects.py
```diff
@@ -336,6 +340,21 @@ def create_experiment(self,
...
+                neptune.create_experiment(upload_source_files=['*.py'])
+
+                # Send all files in the current directory (excluding hidden files with names beginning with a dot)
```
pitercl: Both files and directories

neptune/projects.py
```diff
@@ -372,6 +391,10 @@ def create_experiment(self,
         if notebook_id is None and os.getenv(NOTEBOOK_ID_ENV_NAME, None) is not None:
             notebook_id = os.environ[NOTEBOOK_ID_ENV_NAME]
 
+        expanded_source_files = []
```
pitercl: Use set here, please

PR #190, opened 2019-11-18
Improve upload files performance

szymon-kuklewicz:


neptune/internal/storage/storage_utils.py
```diff
@@ -95,8 +101,22 @@ def __repr__(self):
...
-        size = 0 if os.path.isdir(entry.source_path) else os.path.getsize(entry.source_path)
+        if os.path.isdir(entry.source_path):
+            for root, _, files in os.walk(entry.source_path):
+                path_relative_to_entry_source = os.path.relpath(root, entry.source_path)
+                if path_relative_to_entry_source == ".":
+                    target_root = entry.target_path
+                else:
+                    target_root = entry.target_path + "/" + path_relative_to_entry_source
+                for file in files:
+                    walked_entries.add(UploadEntry(root + "/" + file, target_root + "/" + file))
```
hubertjaworskineptune: constructing paths like this is dangerous, use os.path.join. Even so, this should be tested on windows

neptune/internal/storage/storage_utils.py
```diff
@@ -95,8 +101,22 @@ def __repr__(self):
 
 def split_upload_files(upload_entries, max_package_size=1 * 1024 * 1024, max_files=500):
     current_package = UploadPackage()
+    walked_entries = set()
     for entry in upload_entries:
-        size = 0 if os.path.isdir(entry.source_path) else os.path.getsize(entry.source_path)
+        if os.path.isdir(entry.source_path):
```
hubertjaworskineptune: given a list of directories, we'll upload it's contents 1 level deep with this code.
Questions:
1. why don't we upload entire recursive dir?
2. if this is a desired behaviour for a use-case, this logic should be in the method that calls split_upload_files - it should pass a list of files without directories

neptune/internal/storage/storage_utils.py
```diff
@@ -95,8 +101,22 @@ def __repr__(self):
...
+            for root, _, files in os.walk(entry.source_path):
+                path_relative_to_entry_source = os.path.relpath(root, entry.source_path)
+                if path_relative_to_entry_source == ".":
```
pitercl: We can replace this if/else with
```py
target_root = os.path.join(entry.target_path, path_relative_to_entry_source)
normalized_root = os.path.normpath(target_root)
```

This should be less sensitive to different platforms.

neptune/internal/storage/storage_utils.py
pitercl: 1. `os.walk` is recursive.
2. I agree, that this method coult now be split into 2 for readability.

neptune/internal/storage/storage_utils.py
pitercl: Also, please use some name other than `file`. I think it'd bad form in Python, since this is shadows a built-in.

PR #192, opened 2019-11-26
Fix text logging to restored experiment

### LINKED ISSUE #105: Experiment state and stdout channel do not update when resuming an experiment
Hi,

Sending metrics when resuming an experiment works fine. However the experiment state does not change from "succeeded" to "running". And neither does it log system information (cpu/gpu information) or stdout when resuming an experiment. Code example:

```
import time
import neptune
from neptune.sessions import Session

# Initialize experiment
neptune.init(project_qualified_name='Test/project')
experiment = neptune.create_experiment(name='test')
experiment_id = experiment.id

# Send metrics
for i in range(10):
    time.sleep(2)
    print('logging this to STDOUT channel.')
    experiment.send_metric('iter', i)
    
experiment.stop()

# Resume experiment
session = Session()
project = session.get_project(project_qualified_name='Test/project')
experiment = project.get_experiments(id=experiment_id)[0]

# Send metrics
for i in range(10):
    time.sleep(2)
    print('Unable to log this to STDOUT channel.')
    experiment.send_metric('iter', 3*i)

experiment.stop()
```

wjaskowski:
Now when the experiment is restored from neptune server, you can log text
(stdout, stderr, logger) without the fear that the x values will be
the same as the old ones and will be discarded. (see the rationale:
https://spectrum.chat/neptune-community/general/continuing-an-interrupted-experiment~e9fff5f8-c322-48fc-a84f-2ece82e5c630)

Fixes #105

### DIFF
--- a/neptune/backend.py
+++ b/neptune/backend.py
@@ -26,6 +26,10 @@ class Backend(object):
     def api_address(self):
         pass
 
+    @abstractproperty
+    def display_address(self):
+        pass
+
     @abstractmethod
     def get_project(self, project_qualified_name):
         pass
--- a/neptune/exceptions.py
+++ b/neptune/exceptions.py
@@ -104,3 +104,13 @@ class InvalidNeptuneBackend(NeptuneException):
             'Use this environment variable to modify neptune-client behaviour at runtime, '
             'e.g. using {}=offline allows you to run your code without logging anything to Neptune'
             .format(envs.BACKEND, provided_backend_name, envs.BACKEND))
+
+class DeprecatedApiToken(NeptuneException):
+    def __init__(self, app_url):
+        super(DeprecatedApiToken, self).__init__(
+            "Your API token is deprecated. Please visit {} to get a new one.".format(app_url))
+
+class CannotResolveHostname(NeptuneException):
+    def __init__(self, host):
+        super(CannotResolveHostname, self).__init__(
+            "Cannot resolve hostname {}. Please contact Neptune support.".format(host))
--- a/neptune/experiments.py
+++ b/neptune/experiments.py
@@ -150,7 +150,7 @@ class Experiment(object):
     def get_system_properties(self):
         """Retrieve experiment properties.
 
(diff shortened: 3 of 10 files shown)


kamil-kaczmarek:
Hey @wjaskowski thanks for the PR

@pitercl - can you take a look?

pitercl:
Hi @wjaskowski! Thank you for this PR. I understand the problem you're solving here.

I have some reservations about calling backend every time this class is instantiated, but I think the impact is negligible.

Please, fix the implementation for Python 2.7 (which, unfortunately, we still need to support), and I'll merge it :)

kamil-kaczmarek:
Hey @wjaskowski once again thanks for the PR.

As I understand form @pitercl he is willing to merge it. However, some tests are failing. Would you mind to fix them, so that we can merge it? If you need any help please let us know here.

Thanks!
Kamil

wjaskowski:
Sorry, I have had no time recently. Could you point me how to run the tests
to see what is failing?

pitercl:
That's no problem.

First of all, you can see the tests that fail on Travis here: 
https://travis-ci.org/neptune-ml/neptune-client/builds/621092359?utm_source=github_status&utm_medium=notification

But if you want to run them locally, create a virtual env and run:
```sh
pip install -r requirements.txt
pip install -r test_requirements.txt
make tests
```

Local tests are executed on Python 2.7, 3.5, 3.6.

wjaskowski:
I have found some workarounds for python 2.7 (https://stackoverflow.com/questions/51913210/python-script-with-timezone-fails-when-back-ported-to-python-2-7). Which one do you prefer?

pitercl:
If I understand the problem correctly, we may get away with the following approach:
```py
self._time_started = experiment.get_system_properties()['created']
...

x=(datetime.now(tz=self._time_started.tzinfo) - self._time_started).total_seconds() * 1000
```

What do you think?

wjaskowski:
Some magic in tests doesn't work now...

pitercl:
Ok, thanks for taking it this far. Leave the rest to us - we'll fix it this week.

PR #194, opened 2019-11-28
Support new tokens

aniezurawski:


neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -66,8 +68,12 @@ def __init__(self, api_token=None, proxies=None):
...
-                                                               .format(self.api_address))
+        self.backend_swagger_client = self._get_swagger_client('{}/api/backend/swagger.json'.format(api_url))
+        self._client_config = self.get_client_config(api_token)
+
+        if self._client_config.api_url != api_url:
+            self.backend_swagger_client = self._get_swagger_client('{}/api/backend/swagger.json'
+                                                                   .format(self.api_address))
 
         self.leaderboard_swagger_client = self._get_swagger_client('{}/api/leaderboard/swagger.json'
                                                                    .format(self.api_address))
```
pitercl: Extract this to a method, please, and maybe avoid using `self.api_address` here, since its correctness depends on something that happens as part of this operation.

neptune/client_config.py
```diff
@@ -0,0 +1,30 @@
+#
```
pitercl: This is an internal implementation detail. It should be in `neptune/internal/backends`

PR #201, opened 2020-01-13
Migrate dependency from nvidia-ml-py3 to py3nvml

szymon-kuklewicz:


neptune/internal/hardware/gpu/gpu_monitor.py
```diff
@@ -23,14 +23,17 @@ def get_card_count(self):
         return self.__nvml_get_or_else(nvmlDeviceGetCount, default=0)
 
     def get_card_usage_percent(self, card_index):
+        # pylint: disable=no-member
```
hubertjaworskineptune: add a comment why this is necessary (x3)

PR #206, opened 2020-01-23
Retry api call on 502 bad gateway error

aniezurawski:


neptune/utils.py
```diff
@@ -201,7 +201,7 @@ def wrapper(*args, **kwargs):
                 raise SSLError()
             except (BravadoConnectionError, BravadoTimeoutError,
                     requests.exceptions.ConnectionError, requests.exceptions.Timeout,
-                    HTTPRequestTimeout, HTTPServiceUnavailable, HTTPGatewayTimeout):
+                    HTTPRequestTimeout, HTTPServiceUnavailable, HTTPGatewayTimeout, HTTPBadGateway):
```
hubertjaworskineptune: You should also catch basic 502 (non-bravado exceptionn)

PR #207, opened 2020-01-28
Add delete_artifacts

hubertjaworskineptune:


neptune/__init__.py
```diff
@@ -338,6 +338,14 @@ def send_artifact(artifact, destination=None):
     return get_experiment().log_artifact(artifact, destination)
 
 
+def delete_artifacts(path):
```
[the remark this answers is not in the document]
hubertjaworskineptune: We decided to use delete_aftifacts as the name, since you can delete entire directory of artifacts with it

neptune/experiments.py
```diff
@@ -603,6 +603,36 @@ def log_artifact(self, artifact, destination=None):
...
+        """
+        if path is not None:
+            paths = path
+            if not isinstance(path, list):
+                paths = [path]
+            paths = [p[1:] if p.startswith['/'] else p for p in paths]
+            project_storage_path = ["/{experimentId}/output/{file}".format(experimentId=self.id, file=p) for p in
+                                    paths]
+            try:
+                self._backend.rm_data(project=self._project, project_storage_path=project_storage_path, recursive=True)
```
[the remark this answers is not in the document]
hubertjaworskineptune: This is outdated, in newest version I implemented failsafes for this case

tests/neptune/test_experiment.py
```diff
@@ -150,6 +150,46 @@ def build_call(tags_list):
...
+                recursive=True
+            )
+
+        # when
+        experiment.delete_artifacts('/an_abs_path_in_exp_output')
+        experiment.delete_artifacts('/../an_abs_path_in_exp')
+        experiment.delete_artifacts('/../../an_abs_path_in_prj')
+        experiment.delete_artifacts('a_path_in_exp_output')
+        experiment.delete_artifacts('../a_path_in_exp')
+        experiment.delete_artifacts('../../a_path_in_prj')
```
aniezurawski: Why all those `../` paths do not throw exceptions?

hubertjaworskineptune: because we prepend user paths with shortid/output client-side before sending to server

neptune/experiments.py
```diff
@@ -603,6 +603,39 @@ def log_artifact(self, artifact, destination=None):
                           upload_tar_api_fun=self._backend.extract_experiment_output,
                           experiment=self)
 
+    def delete_artifacts(self, path):
+        """Removes an artifact(s) (file/directory) from the experiment storage.
+
+        Args:
+            path (:obj:`list`): List of paths to remove from the experiment's output
```
aniezurawski: or string

neptune/experiments.py
```diff
@@ -603,6 +603,39 @@ def log_artifact(self, artifact, destination=None):
...
+        Raises:
+            `FileNotFound`: If a path in experiment artifacts does not exist.
+
+        Examples:
+            Assuming that `experiment` is an instance of :class:`~neptune.experiments.Experiment`.
+
+            .. code:: python3
+
+                experiment.delete_artifacts('forest_results.pkl')
+                experiment.delete_artifacts('')
```
aniezurawski: Example for list of paths.

neptune/experiments.py
```diff
@@ -603,6 +603,39 @@ def log_artifact(self, artifact, destination=None):
...
+                if path.startswith(".."):
+                    raise ValueError("path to delete must be within project's directory")
+                if path == "." or path == "/":
+                    raise ValueError("path must be different than project's directory")
+            try:
+                self._backend.rm_data(experiment=self, paths=paths)
+                return
+            except PathInProjectNotFound:
+                raise FileNotFound(path)
+        raise ValueError("path argument must not be None")
```
aniezurawski: It's easier to read if you write
```
if path is None:
    raise ValueError("path argument must not be None")
```

neptune/experiments.py
```diff
@@ -603,6 +603,39 @@ def log_artifact(self, artifact, destination=None):
...
+                experiment.delete_artifacts('forest_results.pkl')
+                experiment.delete_artifacts('')
+        """
+        if path is not None:
+            paths = path
+            if not isinstance(path, list):
+                paths = [path]
+            for path in paths:
+                if path.startswith(".."):
+                    raise ValueError("path to delete must be within project's directory")
```
aniezurawski: Remove this whole condition and add
```
path = os.path.abspath(os.sep + path.strip(os.sep))[1:]
```

neptune/experiments.py
```diff
@@ -603,6 +603,39 @@ def log_artifact(self, artifact, destination=None):
...
+                if path == "." or path == "/":
```
aniezurawski: `or not path`

neptune/experiments.py
```diff
@@ -603,6 +603,39 @@ def log_artifact(self, artifact, destination=None):
...
+                    raise ValueError("path must be different than project's directory")
```
aniezurawski:
```
    raise ValueError("Cannot delete whole artifacts directory")
```

neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -682,6 +682,17 @@ def extract_experiment_output(self, experiment, data):
                 raise StorageLimitReached()
             raise
 
+    @with_api_exceptions_handler
+    def rm_data(self, experiment, paths):
+        try:
+            return [
```
aniezurawski: Loop should be in Experiment instead of Client since our api endpoint does not accept list of paths.

tests/neptune/test_experiment.py
```diff
@@ -150,6 +150,40 @@ def build_call(tags_list):
...
+        def build_call(paths):
+            return call(
+                experiment=experiment,
+                paths=paths
+            )
+
+        # when
+        experiment.delete_artifacts('/an_abs_path_in_exp_output')
+        experiment.delete_artifacts('/../an_abs_path_in_exp')
+        experiment.delete_artifacts('/../../an_abs_path_in_prj')
```
aniezurawski: Should fail.

tests/neptune/test_experiment.py
```diff
@@ -150,6 +150,40 @@ def build_call(tags_list):
...
+        # and
+        def build_call(paths):
+            return call(
+                experiment=experiment,
+                paths=paths
+            )
+
+        # when
+        experiment.delete_artifacts('/an_abs_path_in_exp_output')
+        experiment.delete_artifacts('/../an_abs_path_in_exp')
```
aniezurawski: Should fail.

neptune/experiments.py
hubertjaworskineptune: I disagree. Your way changes path that goes to server to output-relative and may cause user dot incidentally delete a file he might not have wanted

tests/neptune/test_experiment.py
hubertjaworskineptune: actually no, it shouldn't . If follows the behaviour of linux standard commands.
execute following script:
```
touch /tmp/test
rm /../tmp/test
```
It will successfully remove the file

tests/neptune/test_experiment.py
hubertjaworskineptune: explanation as above

neptune/experiments.py
aniezurawski: Could you give me an example? I don't see such possibility.

tests/neptune/test_experiment.py
aniezurawski: Right.

tests/neptune/test_experiment.py
```diff
@@ -150,6 +150,40 @@ def build_call(tags_list):
...
+                path=path
+            )
+
+        # when
+        experiment.delete_artifacts('/an_abs_path_in_exp_output')
+        experiment.delete_artifacts('/../an_abs_path_in_exp')
+        experiment.delete_artifacts('/../../an_abs_path_in_prj')
+        experiment.delete_artifacts('a_path_in_exp_output')
+        self.assertRaises(ValueError, experiment.delete_artifacts, '../a_path_outside_exp')
+        self.assertRaises(ValueError, experiment.delete_artifacts, "..")
```
aniezurawski: How about `dir/../../a_path_outside_exp'`?

neptune/experiments.py
hubertjaworskineptune: passing `somthing to '../file'` would delete `/file` in artifact output, which is confusing and in most cases wrong. I added this check as a simple pre-send validation of a request that would fail anyway due to server

aniezurawski: Yeah. This is exactly what executing `rm ../tmp/file` from root directroy would do - remove `/tmp/file`. Ask @kamil-kaczmarek

hubertjaworskineptune: as agreed, we'll use os.path.normpath to check if path starts with `..`

PR #209, opened 2020-02-03
Support for absolute/relative paths with .. while uploading sources

szymon-kuklewicz:


neptune/projects.py
```diff
@@ -412,14 +414,25 @@ def create_experiment(self,
             if os.path.isfile(main_file):
                 entrypoint = normalize_file_name(os.path.basename(main_file))
                 upload_source_entries = [
-                    UploadEntry(os.path.abspath(main_file), normalize_file_name(os.path.basename(main_file)))
+                    UploadEntry(abspath(main_file), normalize_file_name(os.path.basename(main_file)))
```
hubertjaworskineptune: You imported abspath, but not basename, I think it's confusing and it's better to leave os.path.abspath for reading clarity

neptune/projects.py
```diff
@@ -412,14 +414,25 @@ def create_experiment(self,
...
+                absolute_paths.add(abspath(filepath))
+            try:
+                common_source_root = commonpath(absolute_paths)
+                if os.path.isfile(common_source_root):
+                    common_source_root = os.path.dirname(common_source_root)
+                for absolute_path in absolute_paths:
+                    upload_source_entries.append(UploadEntry(absolute_path, normalize_file_name(os.path.relpath(absolute_path, common_source_root))))
+            except ValueError:
+                for absolute_path in absolute_paths:
+                    upload_source_entries.append(UploadEntry(absolute_path, normalize_file_name(absolute_path)))
```
hubertjaworskineptune: 1. When error occurs on last file you re-upload them all
2.Where exactly does the ValueError come from?

neptune/internal/storage/storage_utils.py
```diff
@@ -134,7 +134,7 @@ def split_upload_files(upload_entries, max_package_size=1 * 1024 * 1024, max_fil
 
 
 def normalize_file_name(name):
-    return name.replace(os.sep, '/')
+    return name.replace(sep, '/')
```
hubertjaworskineptune: I think this code is confusing long-term - sep looks like a normal variable and I need to look up imports to understand what it is

neptune/projects.py
szymon-kuklewicz: This only collects UploadEntry in the memory, does not send files anywhere. Actual sending is done later, out of scope of this change.

PR #211, opened 2020-02-06
Change max image size to 15MB

hubertjaworskineptune:


neptune/internal/channels/channels_values_sender.py
```diff
@@ -116,8 +116,8 @@ def run(self):
 
             if self._sleep_time <= 0 \
                     or len(self._values_batch) >= self._MAX_VALUES_BATCH_LENGTH \
-                    or len([v for v in self._values_batch if
-                            v.channel_type == ChannelType.IMAGE.value]) >= self._MAX_IMAGE_VALUES_BATCH_LENGTH:  # pylint:disable=line-too-long
+                    or sum([len(v.channel_value.y) for v in self._values_batch if
```
aniezurawski: Is `v.channel_value.y` a dict, isn't it? Isn't `v.channel_value.y['image_value']` what you meant?

hubertjaworskineptune: I thoght so too, but testing/debugging clearly shows that Y is a base64-encoded string containing the image

hubertjaworskineptune: We had invalid fixtures in tests

PR #213, opened 2020-02-24
Fix starting experiment outside of notebook with env set

hubertjaworskineptune:


neptune/internal/notebooks/notebooks.py
```diff
@@ -29,17 +29,20 @@ def create_checkpoint(backend, notebook_id, notebook_path):
...
         import IPython
         ipython = IPython.core.getipython.get_ipython()
-        execution_count = ipython.kernel.execution_count
+        execution_count = -1
+        if ipython is not None and ipython.kernel is not None:
+            execution_count = ipython.kernel.execution_count
         with _checkpoints_lock:
 
-            if execution_count in _checkpoints:
+            if execution_count in _checkpoints and execution_count != -1:
```
aniezurawski: Is second condition necessary? `execution_count in _checkpoints` implies `execution_count != -1`, right?

hubertjaworskineptune: Agreed, changed

PR #215, opened 2020-03-02
Fix uploading wrong sources from Jupyter Lab

szymon-kuklewicz:


neptune/utils.py
```diff
@@ -256,3 +256,11 @@ def glob(pathname):
         return globlib.glob(pathname)
     else:
         return globlib.glob(pathname, recursive=True)
+
+def is_ipython():
+    try:
+        import IPython
+        ipython = IPython.core.getipython.get_ipython()
```
hubertjaworskineptune: I think we have similar code elsewhere - if you add this utility method, we should refernce it everywhere, otherwise lgtm

szymon-kuklewicz: Yes we have similar code in one other place, but it can't be easily deduplicated because it is more complex and uses intermediate internal variables that are returned from this function.

PR #220, opened 2020-03-31
Remove double warning about missing psutil on experiment start

pitercl:


neptune/internal/abort.py
```diff
@@ -34,56 +37,36 @@ def __init__(self, pid):
 
     @staticmethod
     def requirements_installed():
-        # pylint:disable=unused-import,unused-variable,bad-option-value,import-outside-toplevel
-        try:
-            import psutil
-            return True
-        except ImportError:
-            _logger.warning('psutil is not installed. The experiment will not be abortable.')
```
aniezurawski, line 42: Nie widzę jak zmiany mają się do commit message'a. Z tego co rozumiem były wcześniej dwa różne komunikaty o braku psutil (dla aborta i dla metryk) i wciąż tak będzie, przy czym trochę lepszy kod.

neptune/internal/abort.py
```diff
@@ -34,56 +37,36 @@ def __init__(self, pid):
...
             process.terminate()
         except psutil.NoSuchProcess:
             pass
 
     def _kill(self, process):
-        # pylint:disable=bad-option-value,import-outside-toplevel
-        import psutil
-
-        for process in self._get_processes(process):
+        for process in self._get_process_with_children(process):
```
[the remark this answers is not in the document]
pitercl: I think the author's idea might have been that since then some new children could've popped up. This, of course, is far from bulletproof. Just a bit better.

neptune/internal/abort.py
pitercl: Yes, there will be 2, instead of 3 (right now we print the message about abort twice) :)
`requirements_installed` is called twice so I decided to move the error log to a place that should be executed onece per experiment.

PR #222, opened 2020-03-31
Fix sending stdout/stderr on Windows

pitercl:


neptune/internal/streams/channel_writer.py
```diff
@@ -53,3 +53,26 @@ def write(self, data):
             )
 
         self._data = lines[-1]
+
+
+class TimeOffset(object):
```
hubertjaworskineptune, line 58: Maybe TimeOffsetGenerator name would be better? would explain the .next() call

pitercl: Done

PR #223, opened 2020-04-01
Fix stopping daemon threads in Python >= 3.7

pitercl:
The goal here is to make the following script exit on Python 3.7 and later:
```python
import neptune
neptune.init()
neptune.create_experiment()
```

Before 3.7, when the main thread exited, a `join()` method was called on every daemonic thread. This changed and so we're now checking manually if the main thread still exists.

### DIFF
--- a/neptune/internal/threads/neptune_thread.py
+++ b/neptune/internal/threads/neptune_thread.py
@@ -25,6 +25,8 @@ class NeptuneThread(threading.Thread):
         self._interrupted = threading.Event()
 
     def should_continue_running(self):
+        # TODO: remove this pylint exception once we stop supporting Python 2
+        # pylint: disable=no-member
         if utils.is_python_2():
             all_threads = threading.enumerate()
 
--- a/neptune/internal/threads/neptune_thread.py
+++ b/neptune/internal/threads/neptune_thread.py
@@ -15,6 +15,8 @@
 #
 import threading
 
+from neptune import utils
+
 
 class NeptuneThread(threading.Thread):
     def __init__(self, is_daemon):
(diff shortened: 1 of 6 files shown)


neptune/utils.py
```diff
@@ -266,3 +266,7 @@ def is_ipython():
         return ipython is not None
     except ImportError:
         return False
+
+
+def is_python_2():
```
hubertjaworskineptune, line 271: this function can be replaced with `six.PY2`

neptune/internal/threads/neptune_thread.py
```diff
@@ -15,15 +15,27 @@
...
-    def is_interrupted(self):
-        return self._interrupted.is_set()
+    def should_continue_running(self):
+        # TODO: remove this pylint exception once we stop supporting Python 2
+        # pylint: disable=no-member
+        if utils.is_python_2():
+            all_threads = threading.enumerate()
+
+            # pylint:disable=protected-access
+            main_thread_is_alive = any(t.__class__ is threading._MainThread for t in all_threads)
```
hubertjaworskineptune, line 34: I think this might actually not work.

In py3's threading there is a comment that main thread is marked as not alive only very close to interpreter's death, which leads me to conclusion that we might not know when exactly main thread is removed from `threading.enumerate()`.

However, even if that is true, we're not making it worse and we're close to ending support for python 2.7 anyway

neptune/utils.py
pitercl: Done

neptune/internal/threads/neptune_thread.py
pitercl: In all Pythons (at least 2.7 - 3.8), there is a `threading._shutdown` function. That's where the the main thread is stopped - it happens just before waiting for all non-daemon threads.

However, it was a good catch with `enumerate()`. I added checking for alive-ness of the thread.

PR #224, opened 2020-04-14
Added support streamed uploading

szymon-kuklewicz:


neptune/internal/storage/storage_utils.py
```diff
@@ -107,28 +110,35 @@ def scan_unique_upload_entries(upload_entries):
...
-
-        if (size + current_package.size > max_package_size or current_package.len > max_files) \
-                and not current_package.is_empty():
+        if entry.is_stream():
             yield current_package
             current_package.reset()
-        current_package.update(entry, size)
+            current_package.update(entry, 0)
+            yield current_package
+            current_package.reset()
```
hubertjaworskineptune, line 134: Since user can provide an open file/stream, is it possible that resetting will fail?

On certain server response codes we repeat the upload, won't this invalidate the non-resettable steam and cause uploaded file to be corrupted?

PR #226, opened 2020-04-16
Inform about updates

aniezurawski:


neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -915,7 +920,53 @@ def _create_authenticator(self, api_token, ssl_verify, proxies):
     @with_api_exceptions_handler
     def _create_client_config(self, api_token, backend_client):
         config = backend_client.api.getClientConfig(X_Neptune_Api_Token=api_token).response().result
-        return ClientConfig(api_url=config.apiUrl, display_url=config.applicationUrl)
+        
+        min_recommended = None
+        min_compatible = None
+        max_compatible = None
+
+        if hasattr(config, "versions"):
```
hubertjaworskineptune, line 928: else section with a warning would be nice - non-SaaS users would get a warning with information that this client is newer than they (propably) support

neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -79,15 +82,17 @@ def __init__(self, api_token=None, proxies=None):
         backend_client = self._get_swagger_client('{}/api/backend/swagger.json'.format(config_api_url))
         self._client_config = self._create_client_config(self.credentials.api_token, backend_client)
```
hubertjaworskineptune, line 83: I'm thinking of refactoring this code to be the first call (before getting backend swagger.json) with a static url.

This might allow us to move swagger file at some point, but will still force us to support config endpoint with static path

aniezurawski: I just changed url of this endpoint in backend and swagger path was never changed, so... I'm not so sure about this. Let's leave it as it is.

neptune/internal/backends/hosted_neptune_backend.py
hubertjaworskineptune: I think this key changed

PR #236, opened 2020-06-11
Fixes "neptune.init() throws exceptions with combination of ipython and offline-mode"

PiotrJander:
The fix was to replace `__getattribute__` with `__getattr__` in `NoopBackend` (difference explained [here](https://stackoverflow.com/questions/3278077/difference-between-getattr-vs-getattribute)).

Fixes https://neptune-labs.atlassian.net/browse/NPT-8209 , not sure how to link JIRA issues

### DIFF
--- a/neptune/internal/backends/offline_backend.py
+++ b/neptune/internal/backends/offline_backend.py
@@ -131,14 +131,5 @@ class OfflineBackend(Backend):
 
 class NoopObject(object):
 
-    def __getattribute__(self, item):
-        return NoopObject()
-
-    def __call__(self, *args, **kwargs):
-        return NoopObject()
-
-    def __enter__(self):
-        return NoopObject()
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        pass
+    def __getattr__(self, item):
+        return None
--- a/neptune/internal/backends/offline_backend.py
+++ b/neptune/internal/backends/offline_backend.py
@@ -132,4 +132,13 @@ class OfflineBackend(Backend):
 class NoopObject(object):
 
     def __getattr__(self, item):
-        return None
+        return self
+
+    def __call__(self, *args, **kwargs):
+        return self
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
--- a/tests/neptune/internal/backends/test_noop_backend.py
+++ b/tests/neptune/internal/backends/test_noop_backend.py
@@ -1,0 +1,25 @@
+#
+# Copyright (c) 2019, Neptune Labs Sp. z o.o.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


tests/neptune/internal/backends/test_noop_backend.py
```diff
@@ -0,0 +1,25 @@
...
+
+import unittest
+
+from neptune.internal.backends.offline_backend import NoopObject
+
+class TestHostedNeptuneBackend(unittest.TestCase):
+
+    def test_builtin_fields_not_overriden(self):
+        objectUnderTest = NoopObject()
+        self.assertIsInstance(objectUnderTest.__class__, type)
```
PiotrJander, line 25: Regression test. The bug was caused by `__getattribute__` overriding `__class__`.

neptune/internal/backends/offline_backend.py
```diff
@@ -131,14 +131,5 @@ def download_data(self, project, path, destination):
...
-        return NoopObject()
-
-    def __call__(self, *args, **kwargs):
-        return NoopObject()
-
-    def __enter__(self):
-        return NoopObject()
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        pass
```
PiotrJander, line 144: I tentatively removed `__call__`, `__enter__`, and `__exit__`. I understand that `NoopObject` serves as a placeholder for `Project`, and `Project` doesn't seem to implement any of those three methods. I might be mistaken or be missing something though.

neptune/internal/backends/offline_backend.py
```diff
@@ -131,14 +131,5 @@ def download_data(self, project, path, destination):
...
+    def __getattr__(self, item):
+        return None
```
PiotrJander, line 135: I didn't see the point of `NoopObject` being returned from `__getattr__` so I made it return `None`. I might be mistaken or be missing something though.

tests/neptune/internal/backends/test_noop_object.py
```diff
@@ -0,0 +1,59 @@
...
+#
+
+import unittest
+
+from neptune.internal.backends.offline_backend import NoopObject
+
+class TestHostedNeptuneObject(unittest.TestCase):
+
+    # given
+    objectUnderTest = NoopObject()
```
pitercl, line 24: While it's true that this object should not be mutable, I think it would be prudent to keep it scoped to a single test.
Right now, tests cases can interact with each other.

tests/neptune/internal/backends/test_noop_object.py
```diff
@@ -0,0 +1,59 @@
...
+    some_value = 42
+
+    def test_builtin_fields_not_overriden(self):
```
pitercl, line 27: Given-when-then can sometimes be shortened to given-then :) 

In this case, I'd go with
```py
def test_builtin_fields_not_overriden(self):
       # given
       objectUnderTest = NoopObject()

       # then
       self.assertIsInstance(objectUnderTest.__class__, type)
```

PR #253, opened 2020-08-18
WIP: Wip/waj/new client mvp

aniezurawski:
Work in progress


neptune/handler.py
```diff
@@ -0,0 +1,79 @@
...
+        self._experiment = _experiment
+        self._path = path
+
+    def __getitem__(self, path: str) -> 'Handler':
+        return Handler(self._experiment, self._path + parse_path(path))
+
+    def __setitem__(self, key: str, value) -> None:
+        self[key].assign(value)
+
+    def __getattr__(self, attr):
```
PiotrJander, line 39: So we also support attribute syntax, not just dictionary lookup syntax?

How about paths which are not valid Python attribute names?

neptune/internal/experiment_structure.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def get_structure(self) -> Dict[str, Any]:
+        return self._structure
+
+    def get(self, path: List[str]) -> Optional[T]:
+        ref = self._structure
+
+        for part in path:
+            if not isinstance(ref, dict):
+                raise MetadataInconsistency("Cannot access path '{}': '{}' is already defined as a variable, "
```
PiotrJander, line 36: So we don't support getting subspaces?

neptune/internal/experiment_structure.py
```diff
@@ -0,0 +1,82 @@
...
+        ref[variable_name] = var
+
+    def pop(self, path: List[str]) -> None:
+        self._pop_impl(self._structure, path, path)
+
+    def _pop_impl(self, ref: dict, sub_path: List[str], var_path: List[str]):
+        if not sub_path:
+            return
+
+        head, tail = sub_path[0], sub_path[1:]
```
PiotrJander, line 71: `head, *tail = sub_path` :)

neptune/internal/experiment_structure.py
```diff
@@ -0,0 +1,82 @@
...
+        if head not in ref:
+            raise MetadataInconsistency("Cannot delete {}. Variable not found.".format(var_path))
+
+        if not tail:
+            if isinstance(ref[head], dict):
+                raise MetadataInconsistency("Cannot delete {}. It's a namespace, not a variable.")
```
PiotrJander, line 77: Looks like we cannot delete whole namespaces either?

neptune/handler.py
```diff
@@ -0,0 +1,79 @@
...
+            if isinstance(value, (float, int)):
+                var = FloatVariable(self._experiment, self._path)
+                self._experiment._structure.set(self._path, var)
+            if isinstance(value, str):
+                var = StringVariable(self._experiment, self._path)
+                self._experiment._structure.set(self._path, var)
+        var.assign(value)
+
+    def log(self, value, step=None, timestamp=None) -> None:
+        var = self._experiment._structure.get(self._path)
```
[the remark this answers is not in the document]
aniezurawski: It have to be done on server side since it could be changed by another process.

neptune/handler.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __getattr__(self, attr):
+        var = self._experiment._structure.get(self._path)
+        if var:
+            return getattr(var, attr)
+        else:
+            raise AttributeError()
+
+    def assign(self, value) -> None:
+        var = self._experiment._structure.get(self._path)
```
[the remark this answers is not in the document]
aniezurawski: It have to be done on server side since it could be changed by another process.

neptune/handler.py
aniezurawski: It is not used like that. It proxies method call to underlying variable. So if FloatVariable has a `get` mathod we can call `exp['var'].get()` on a `Handler` object.

neptune/internal/experiment_structure.py
aniezurawski: I cannot see any reason for supporting it for now. Notice that it is just internal implementation of experiment structure. It's not a `Handler`. We still support accessing namespace via `Handler` object like this: `exp['some/ns']`.

neptune/internal/experiment_structure.py
aniezurawski: Oh, cool. :)

neptune/internal/experiment_structure.py
aniezurawski: Yup. No requirement for such operation in specification. Or am I wrong?

PiotrJander: Not sure, we should clarify

PR #254, opened 2020-08-21
Create operation queue with disk storage

aniezurawski:


neptune/internal/collections/disk_queue.py
```diff
@@ -0,0 +1,94 @@
...
+
+        try:
+            os.mkdir(self._dir_path)
+        except FileExistsError:
+            pass
+
+        self._read_file_idx = 0
+        self._write_file_idx = 0
+        self._writer = open(self._current_write_log_file(), "a")
+        self._reader = JsonFileSplitter(self._current_read_log_file())
```
PiotrJander, line 50: I would consider making this a function which always returns the current file (no need to manually maintain the invariant). Admittedly then we couldn't close files manually, but files get closed by GC anyway.

Same for self._writer

neptune/internal/collections/disk_queue.py
```diff
@@ -0,0 +1,94 @@
...
+import os
+from typing import TypeVar, List, Callable, Optional
+
+from neptune.internal.collections.queue import Queue
+from neptune.internal.utils.json_file_splitter import JsonFileSplitter
+
+T = TypeVar('T')
+
+
+class DiskQueue(Queue[T]):
```
PiotrJander, line 26: Consider exposing two interfaces, one for writing and one for reading. No strong feelings about this one though.

neptune/internal/utils/json_file_splitter.py
```diff
@@ -0,0 +1,56 @@
...
+#
+import json
+from json import JSONDecodeError
+from typing import Optional
+
+
+class JsonFileSplitter:
+
+    # TODO: experiment with larger buffer sizes
+    BUFFER_SIZE = 4096
```
PiotrJander, line 24: Out of curiosity: why this size?

neptune/internal/utils/json_file_splitter.py
```diff
@@ -0,0 +1,56 @@
...
+    def _read_data(self):
+        data = self._file.read(self.BUFFER_SIZE)
+        if not data:
+            return
+        self._buffer = self._buffer[self._start_pos:] + data
+        self._start_pos = 0
+
+    def _decode(self):
+        self._start_pos = self._buffer.index("{", self._start_pos)
+        data, size = self._decoder.raw_decode(self._buffer, self._start_pos)
```
PiotrJander, line 54: The name `size` is misleading, how about `end`?

PiotrJander (approved):

neptune/internal/utils/json_file_splitter.py
aniezurawski: Ok.

neptune/internal/utils/json_file_splitter.py
aniezurawski: It is default internal buffer size in splitstream library (https://github.com/rickardp/splitstream) which eventually I dicided not to use. But 4kb seems to be good guess.

PR #261, opened 2020-09-01
Implement reporting hardware metrics

aniezurawski:


neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -121,6 +127,69 @@ def execute_operations(self, operations: List[Operation]) -> None:
...
+    def send_hardware_metric_reports(
+            self,
+            experiment_uuid: UUID,
+            metrics: List[Metric],
+            metric_reports: List[MetricReport]) -> None:
+        verify_type("experiment_uuid", experiment_uuid, uuid.UUID)
+        verify_type("metrics", metrics, list)
+        verify_type("metric_reports", metric_reports, list)
+
+        SystemMetricValues = self.backend_client.get_model('SystemMetricValues')
```
hubertjaworskineptune, line 140: This locks us out from changing the name of the model later in backend. Do we want that? what are we getting in return ("compile" time - wise)?

Seems that `verify_type` would assert correct structure and then we can send dicts to bravado api calls

neptune/internal/hardware/hardware_metric_reporting_job.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(self, exp_uuid: UUID, backend: NeptuneBackend, period: float = 10):
+        self._exp_uuid = exp_uuid
+        self._backend = backend
+        self._period = period
+        self._random_execution_id = "{}-{}".format(socket.gethostname(), str(uuid.uuid4()))
+        self._thread = None
+        self._started = False
+
+    def start(self):
+        if not SystemMonitor.requirements_installed():
```
hubertjaworskineptune, line 55: Is there a reason why you dont set those variables in constructor?

aniezurawski: Just because they are needed only by a thraed and we do not want to create w thread if psutils is not installed.

hubertjaworskineptune (approved):

PR #262, opened 2020-09-02
Support notebooks in new client

hubertjaworskineptune:


neptune/internal/notebook.py
```diff
@@ -0,0 +1,72 @@
...
+                 backend: NeptuneBackend,
+                 notebook_id: uuid.UUID,
+                 notebook_path: str
+                 ):
+        self._notebook_id = notebook_id
+        self._notebook_path = notebook_path
+        self._backend = backend
+
+    def create_checkpoint(self) -> uuid.UUID:
+        if is_ipython():
```
aniezurawski, line 43: This method returns `None` if it is false. So returned type should be `Optional[uuid.UUID]`.

neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -209,6 +216,50 @@ def _get_client_config(self, backend_client: SwaggerClient) -> ClientConfig:
             max_compatible_version=version.parse(max_compatible) if max_compatible else None
         )
 
+    @with_api_exceptions_handler
+    def create_checkpoint(self, notebook_id: uuid.UUID, jupyter_path: str, _file: Any = None) -> Optional[uuid.UUID]:
+        # TODO maybe write an endpoint in backend that accepts FormData to be swagger-compatible? shouldn't hurt much
+        #  and we would be able to call this normally
+        if _file is not None:
```
aniezurawski, line 223: You don't use it. Jupyter extension is responsible for uploading content.

neptune/internal/notebook.py
```diff
@@ -0,0 +1,72 @@
...
+from neptune.internal.backends.neptune_backend import NeptuneBackend
+from neptune_old.utils import is_ipython
+
+_logger = logging.getLogger(__name__)
+
+_checkpoints_lock = threading.Lock()
+_checkpoints = dict()
+
+
+class Notebook:
```
aniezurawski, line 31: This module should be called global_notebook_context or something like that. And it should expose only global function. Notebook class will probably something else in future.

neptune/internal/init_impl.py
```diff
@@ -61,7 +62,23 @@ def init(
...
+    notebook_id = None
+    notebook_path = None
+    checkpoint_id = None
+
+    if os.getenv(NOTEBOOK_ID_ENV_NAME, None) is not None:
+        notebook_id = uuid.UUID(os.environ[NOTEBOOK_ID_ENV_NAME])
+    if os.getenv(NOTEBOOK_PATH_ENV_NAME, None) is not None:
+        notebook_path = os.environ[NOTEBOOK_PATH_ENV_NAME]
+    if notebook_id is not None and notebook_path is not None:
+        checkpoint_id = Notebook(backend, notebook_id, notebook_path).create_checkpoint()
```
aniezurawski, line 75: Check for None.

aniezurawski (changes requested):

neptune/internal/init_impl.py
hubertjaworskineptune: I don't need - I pass this downstream to experiment creation where it's sent to backend as None

neptune/internal/backends/hosted_neptune_backend.py
hubertjaworskineptune: Actually, this is not true. Creating checkpoint (as in database entry) is separated from uploading it's content, which, in fact, is done by Jupyter

aniezurawski (approved):

PR #263, opened 2020-09-02
Make experiment thread safe

aniezurawski:


neptune/internal/utils/paths.py
```diff
@@ -13,12 +13,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from typing import List
+from typing import List, Iterable
+
+
+def _remove_empty_paths(paths: Iterable[str]) -> List[str]:
+    return [path for path in paths if path]
```
PiotrJander, line 20: Could be `filter(bool, paths)`

neptune/variables/series/float_series.py
```diff
@@ -15,32 +15,39 @@
...
-        if not timestamp:
-            timestamp = time.time()
-        self._next_step = step + 1
-
-        self._experiment._op_processor.enqueue_operation(LogFloats(self._experiment._uuid, self._path, [value]), wait)
+    def assign(self, value: FloatSeriesVal, wait: bool = False):
+        verify_type("value", value, FloatSeriesVal)
+        with self._experiment.lock():
+            self.clear()
+            # TODO: Avoid loop
```
PiotrJander, line 36: How can we avoid a loop here?

neptune/variables/series/float_series.py
```diff
@@ -15,32 +15,39 @@
...
-    def log(self, value: float, step: float = None, timestamp: float = None, wait: bool = False):
-        # TODO: Support steps and timestamps
-        if not step:
-            step = self._next_step
-        if not timestamp:
-            timestamp = time.time()
-        self._next_step = step + 1
-
-        self._experiment._op_processor.enqueue_operation(LogFloats(self._experiment._uuid, self._path, [value]), wait)
+    def assign(self, value: FloatSeriesVal, wait: bool = False):
```
PiotrJander, line 32: Same code for both float and string series, any chance to do it generically?

PiotrJander (approved):

neptune/variables/series/float_series.py
aniezurawski: Using Operation with list as a value instead of logging single value.

neptune/variables/atoms/float.py
```diff
@@ -24,8 +27,11 @@
 
 class Float(Atom):
 
-    def assign(self, value: float, wait: bool = False):
-        self._experiment._op_processor.enqueue_operation(AssignFloat(self._experiment._uuid, self._path, value), wait)
+    def assign(self, value: Union[float, int], wait: bool = False):
```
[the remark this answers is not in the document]
aniezurawski: Not worth it. Type verification is different. It would result in more complicated code actually (implementing `asign` in parent and calling `super` in childs).

PR #276, opened 2020-09-28
Combine operations and send them to server

aniezurawski:


neptune/alpha/internal/backends/operations_preprocessor.py
```diff
@@ -0,0 +1,190 @@
...
+    FLOAT = "Float",
+    STRING = "String",
+    FILE = "File"
+    FLOAT_SERIES = "Float Series",
+    STRING_SERIES = "String Series",
+    IMAGE_SERIES = "Image Series",
+    STRING_SET = "String Set"
+
+
+class _OperationsAccumulator(OperationVisitor[None]):
```
PiotrJander, line 53: This algorithm is nice and generic (I'll certainly borrow some ideas for the Java version) but I can see some potential problems: there is a property which I'll sketch below which an accumulating/compressing algorithm should respect, and I'm not such this algorithm respects this property.

Namely, compressed operations should be equivalent to the original operations, regardless of the state of the attribute they act on. But the client does not know the experiment's state on the server, so it needs to compress in a way which works for all possible server states.

For example, suppose that an experiment has a `FloatSeries([1.0, 1.1])` attribute defined at path `path`.

Now suppose that we have an uncompressed list of operations `[AssignFloat(path, 1.2), LogFloats(path, [1.3, 1.4])]`.

If you send the original list to the server, then the server will reject the `AssignFloat` operation with an error, and process the `LogFloats` operation ending up in a state with `path` set to `FloatSeries([1.0, 1.1, 1.3, 1.4])`.

But your aggregation algorithm will compress those original operations to just `[AssignFloat(path, 1.2)]`, rejecting the `LogFloats` operation.

If you send `[AssignFloat(path, 1.2)]` to the server, then the server will reject that operation with an error and end up in a state with `path` set to `FloatSeries([1.0, 1.1])`, which is a different state.

An example of a client-side aggregation which works for all possible server states is compressing

`[AssignFloat(path, 1.2), LogFloats(path, [1.3, 1.4]), LogFloats(path, [1.5, 1.6])]`

to 

`[AssignFloat(path, 1.2), LogFloats(path, [1.3, 1.4, 1.5, 1.6])]`

neptune/alpha/internal/backends/operation_api_object_converter.py
```diff
@@ -0,0 +1,95 @@
...
+from neptune.alpha.internal.operation_visitor import OperationVisitor
+
+
+class OperationApiObjectConverter(OperationVisitor[dict]):
+
+    def convert(self, op: Operation) -> dict:
+        return op.accept(self)
+
+    def visit_assign_float(self, op: AssignFloat) -> dict:
+        return {
```
PiotrJander, line 30: Dlaczego nie uzywamy klas wygenerowanych przez bravado jak w https://bravado.readthedocs.io/en/stable/quickstart.html#lets-try-a-post-call ?

neptune/alpha/internal/backends/operations_preprocessor.py
```diff
@@ -140,7 +158,7 @@ def visit_delete_attribute(self, op: DeleteAttribute) -> None:
             else:
                 # This case is tricky. There was no delete operation, but some modifications was performed.
                 # We do not know if this attribute exists on server side and we do not want a delete op to fail.
-                # So we need to send a single modification before delete so be sure a delete op is valid.
+                # So we need to send a single modification before delete to be sure a delete op is valid.
```
PiotrJander, line 158: Or we could make the delete operation idempotent and define its meaning in terms of the end result (after processing a delete operation, the attribute does not exist, whether it existed previously or not) rather than "delete operation deletes an existing attribute"

neptune/alpha/internal/operation_processors/async_operation_processor.py
```diff
@@ -67,6 +71,9 @@ def stop(self):
 
     class ConsumerThread(Daemon):
 
+        RETRIES = 10
+        RETRY_WAIT = 30
```
PiotrJander, line 75: Tak na przyszlosc, ten pr mozna bylo rozbic na dwa

PiotrJander (approved):

PR #281, opened 2020-10-03
Added docs links to error messages

jakubczakon:
Added:
* links to relevant docs in error messages
* changed things to make the copy more friendly
* added colouring

### DIFF
--- a/neptune/exceptions.py
+++ b/neptune/exceptions.py
@@ -157,8 +157,8 @@ You may also want to check the following docs pages:
     - https://docs.neptune.ai/getting-started/quick-starts/log_first_experiment.html
 
 {correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help.html
-""".format(**STYLES)
-        super(NeptuneNoExperimentContextException, self).__init__(message)
+"""
+        super(NeptuneNoExperimentContextException, self).__init__(message.format(**STYLES))
 
 
 class NeptuneMissingApiTokenException(NeptuneException):
@@ -199,8 +199,9 @@ You may also want to check the following docs pages:
     - https://docs.neptune.ai/getting-started/quick-starts/log_first_experiment.html
 
 {correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help.html
-""".format(**{'env_api_token': envs.API_TOKEN_ENV_NAME, **STYLES})
-        super(NeptuneMissingApiTokenException, self).__init__(message)
+"""
+        inputs = dict(list({'env_api_token': envs.API_TOKEN_ENV_NAME}.items()) + list(STYLES.items()))
+        super(NeptuneMissingApiTokenException, self).__init__(message.format(**inputs))
 
 
 class NeptuneMissingProjectQualifiedNameException(NeptuneException):
@@ -238,8 +239,9 @@ You may also want to check the following docs pages:
     - https://docs.neptune.ai/getting-started/quick-starts/log_first_experiment.html
     
 {correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help.html
-""".format(**{'env_project': envs.PROJECT_ENV_NAME, **STYLES})
-        super(NeptuneMissingProjectQualifiedNameException, self).__init__(message)
+"""
+        inputs = dict(list({'env_project': envs.PROJECT_ENV_NAME}.items()) + list(STYLES.items()))
+        super(NeptuneMissingProjectQualifiedNameException, self).__init__(message.format(**inputs))
 
 
 class NeptuneIncorrectProjectQualifiedNameException(NeptuneException):
@@ -266,8 +268,9 @@ You may also want to check the following docs pages:
     - https://docs.neptune.ai/getting-started/quick-starts/log_first_experiment.html
 
 {correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help.html
-""".format(**{'project_qualified_name': project_qualified_name, **STYLES})
-        super(NeptuneIncorrectProjectQualifiedNameException, self).__init__(message)
+"""
+        inputs = dict(list({'project_qualified_name': project_qualified_name}.items()) + list(STYLES.items()))
+        super(NeptuneIncorrectProjectQualifiedNameException, self).__init__(message.format(**inputs))
 
 
 class InvalidNeptuneBackend(NeptuneException):
--- a/neptune/exceptions.py
+++ b/neptune/exceptions.py
@@ -19,7 +19,7 @@ from neptune import envs
 class NeptuneException(Exception):
     BASE_MESSAGE = """\n
 Neptune team here, seems like there are some problems, sorry.
-No worries thought, in most cases you can fix them by:
+In most cases you can get it to work by:
     - Finding an answer in the docs -> https://docs.neptune.ai
     - Getting help from the team (we are friendly) -> https://docs.neptune.ai/getting-started/getting-help.html
     """
--- a/neptune/exceptions.py
+++ b/neptune/exceptions.py
@@ -50,7 +50,7 @@ before you ran:
 You may also want to check the following docs pages:
     - https://docs.neptune.ai/getting-started/quick-starts/log_first_experiment.html
     
-{correct}Get help{end}: https://docs.neptune.ai/getting-started/getting-help.html
+{correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help.html
 """.format(**STYLES)
         super(NeptuneUninitializedException, self).__init__(message)
 
@@ -109,7 +109,7 @@ To install run:
 You may also want to check the following docs pages:
     - https://docs.neptune.ai/getting-started/installation/index.html
     
-{correct}Get help{end}: https://docs.neptune.ai/getting-started/getting-help.html
+{correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help.html
 """.format(**{'library': library, **STYLES})
         super(NeptuneLibraryNotInstalledException, self).__init__(message)
 
@@ -143,7 +143,7 @@ Looks like you forgot to run:
 You may also want to check the following docs pages:
     - https://docs.neptune.ai/getting-started/quick-starts/log_first_experiment.html
 
-{correct}Get help{end}: https://docs.neptune.ai/getting-started/getting-help.html
+{correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help.html
 """.format(**STYLES)
         super(NeptuneNoExperimentContextException, self).__init__(message)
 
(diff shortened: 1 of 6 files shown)


neptune/exceptions.py
```diff
@@ -13,18 +13,54 @@
...
+    
+You may also want to check the following docs pages:
+    - https://docs.neptune.ai/getting-started/quick-starts/log_first_experiment.html
+    
+{correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help.html
+""".format(**STYLES)
+        super(NeptuneUninitializedException, self).__init__(message)
+
+
+def Uninitialized():
```
pitercl, line 58: This won't work, since you can't do this:
```py3
try:                                                                                                                                                                                                               
    raise exceptions.NeptuneUninitializedException()                                                                                                                                                               
except exceptions.Uninitialized as e:                                                                                                                                                                              
    print(e)                                                                                                                                                                                                       
```
because of
```
TypeError: catching classes that do not inherit from BaseException is not allowed
```

Also, we're using `Uninitialized` in our code, so every time we raise it, user gets a warning they can do nothing about.

If you feel we need the rename, please change the usages in our code as well, but still, `Uninitialized` cannot be left as a function.

I think the most we can do here is something like `Uninitialized = NeptuneUninitializedException`, but I'm not sure how to make our users aware of the deprecation. Maybe it's not required? How many people are actually catching this particular exception?

Basically, those exceptions are thrown only by our lib. The only way our users should use them is to catch them. And I'm not sure anyone would bother to catch exceptions that you're changing here.

neptune/exceptions.py
```diff
@@ -13,18 +13,54 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import warnings
+
 from neptune import envs
 
+STYLES = {'h1': '\033[95m',
```
pitercl, line 20: Here is a nice comment on how to detect if a terminal supports ANSI color codes. 
https://stackoverflow.com/questions/7445658/how-to-detect-if-the-console-does-support-ansi-escape-codes-in-python

Also, as we talked offline, an ENV-based switch would be nice.

For what it's worth, I think it will work in most cases (I checked my own terminal and Jupyter notebook/lab). I'm mostly worried about Windows, TBH - haven't check this yet.

pitercl (changes requested):

neptune/exceptions.py
jakubczakon: As I am using this name in the error message I'd rather have it mention Neptune.
That is why I wanted to rename.

I went with the rename in code also option.
Is this ok?

neptune/exceptions.py
pitercl: Hm, ok, so unfortunately, I checked and on Windows I get get this (both in the standard CMD and in PowerShell):
image

So it seems that we need to do some detecting before using colors.

neptune/exceptions.py
```diff
@@ -13,18 +13,45 @@
...
-            "You must initialize neptune-client first. "
-            "For more information, please visit: https://github.com/neptune-ai/neptune-client#initialize-neptune")
+        message = """
+{h1}     
+----NeptuneUninitializedException---------------------------------------------------------------------------------------
+{end}
+You must initialize neptune-client before you create an experiment.
+
+Looks like you forgot to add:
+    {python}neptune.init(api_token='YOUR_LONG_TOKEN', project_qualified_name='WORKSPACE_NAME/PROJECT_NAME'){end}
```
pitercl, line 44: Here, assuming we're going with proposing to put the token in the code (which is somewhat unorthodox, even though I understand you strive for simplicity), I'd change:
1. the order of the params - if the user doesn't use named params, they'll get another error
2. the `YOUR_LONG_TOKEN` string. I think I might half-wonder if there's something actually called LONG_TOKEN. Let's use our standard name - `YOUR_API_TOKEN`

neptune/exceptions.py
```diff
@@ -59,9 +86,23 @@ def __init__(self):
         super(NoChannelValue, self).__init__('No channel value provided.')
 
 
-class LibraryNotInstalled(NeptuneException):
+class NeptuneLibraryNotInstalledException(NeptuneException):
```
pitercl, line 89: Oook, so I checked and this actually is used [here](https://github.com/neptune-ai/neptune-contrib/blob/c0eb668d6e04f90b51172748e2aa6849c18cc29b/neptunecontrib/monitoring/keras.py#L18). Very weird dependency, but changing this name will break contrib. So we probably need a backward compatible rename, but also removing this dependency from contrib seems like a good idea.

neptune/exceptions.py
```diff
@@ -59,9 +86,23 @@ def __init__(self):
...
     def __init__(self, library):
-        super(LibraryNotInstalled, self).__init__("Library {} is not installed".format(library))
+        message = """
+{h1}     
+----NeptuneLibraryNotInstalledException---------------------------------------------------------------------------------
+{end}
+Looks like library {library} wasn't installed.
+
+To install run:
+    {bash}pip install{library}{end}
```
pitercl, line 98: When I look at the usage in contrib, this won't work. It'll be `pip installKeras` (without a space even).

neptune/exceptions.py
```diff
@@ -71,32 +112,132 @@ def __init__(self, expected_type, actual_type):
...
+        message = """
+{h1}
+----NeptuneMissingApiTokenException-------------------------------------------------------------------------------------
+{end}
+Neptune client couldn't find your API token.
+
+Learn how to get it in this docs page:
+https://docs.neptune.ai/security-and-privacy/api-tokens/how-to-find-and-set-neptune-api-token.html
+
+There are two options two add it:
```
pitercl, line 145: `options two add` -> `options to add`

neptune/exceptions.py
```diff
@@ -71,32 +112,132 @@ def __init__(self, expected_type, actual_type):
...
+    - specify it in your code 
+    - set an environment variable in your operating system.
+
+{h2}CODE{end}
+Pass the token to {bold}neptune.init(){end} via {bold}api_token{end} argument:
+    {python}neptune.init(api_token='YOUR_LONG_TOKEN', project_qualified_name='WORKSPACE_NAME/PROJECT_NAME'){end}
```
pitercl, line 151: Same comments here as in `NeptuneUninitializedException`.

neptune/exceptions.py
```diff
@@ -71,32 +112,132 @@ def __init__(self, expected_type, actual_type):
...
+{end}
+Neptune client couldn't find your project name.
+
+There are two options two add it:
+    - specify it in your code 
+    - set an environment variable in your operating system.
+
+{h2}CODE{end}
+Pass it to {bold}neptune.init(){end} via {bold}project_qualified_name{end} argument:
+    {python}neptune.init(api_token='YOUR_LONG_TOKEN', project_qualified_name='WORKSPACE_NAME/PROJECT_NAME'){end}
```
pitercl, line 190: Same comments here as in NeptuneUninitializedException.

neptune/exceptions.py
```diff
@@ -71,32 +112,132 @@ def __init__(self, expected_type, actual_type):
...
+    {correct}Linux/Unix{end}
+    In your terminal run:
+       {bash}export {env_project}=WORKSPACE_NAME/PROJECT_NAME{end}
 
+    {correct}Windows{end}
+    In your CMD run:
+       {bash}set {env_project}=WORKSPACE_NAME/PROJECT_NAME{end}
 
-class IncorrectProjectQualifiedName(NeptuneException):
+and skip the {bold}project_qualified_name{end} argument of {bold}neptune.init(){end}:
```
pitercl, line 203: I'd surround the given project name with `"` in case it's empty, or a whitespace.

neptune/exceptions.py
```diff
@@ -71,32 +112,132 @@ def __init__(self, expected_type, actual_type):
...
+The correct project qualified name should look like this {correct}WORKSPACE/PROJECT_NAME{end}.
+It has two parts:
+    - {correct}WORKSPACE{end}: which can be your username or your organization name
+    - {correct}PROJECT_NAME{end}: which is the actual project name you chose 
+
+For example, a project {correct}jakub-czakon/blog-hpo{end} parts are:
+    - {correct}jakub-czakon{end}: a {underline}WORKSPACE{end} my username
+    - {correct}blog-hpo{end}: a {underline}PROJECT_NAME{end} a project name
+    
+The URL to this project looks like this: https://ui.neptune.ai/jakub-czakon/blog-hpo
```
pitercl, line 232: Just a quick reminder about our on-prem deployments (in their cases the link will be different) - but I'm ok with that if you are.

Also, I don't have a very strong opinion on using your particular username, but keep in mind that `jakub-czakon` might not be an obvious `firstname-lastname` situation for everyone.

pitercl (changes requested):

neptune/exceptions.py
jakubczakon: Good point, changed to neptune-ai/credit-default-prediction

neptune/exceptions.py
jakubczakon: Done

neptune/exceptions.py
jakubczakon: Done

neptune/exceptions.py
jakubczakon: Done

neptune/exceptions.py
jakubczakon: Fixed

neptune/exceptions.py
jakubczakon: Dropped this usage from contrib (and removed it there with `ModuleNotFoundError`. 
I don't think it made much sense to use custom NeptuneException for such usecase by the way.

neptune/exceptions.py
jakubczakon: Dropped it from contrib see this PR https://github.com/neptune-ai/neptune-contrib/pull/107/files

neptune/exceptions.py
jakubczakon: Done

neptune/exceptions.py
jakubczakon: Done.

For now it just gives no collors to Windows users but if we figure it out we can change it later.

neptune/exceptions.py
```diff
@@ -59,9 +116,23 @@ def __init__(self):
...
+Looks like library {library} wasn't installed.
+
+To install run:
+    {bash}pip install {library}{end}
+
+You may also want to check the following docs pages:
+    - https://docs.neptune.ai/getting-started/installation/index.html
+    
+{correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help.html
+""".format(**{'library': library, **STYLES})
```
pitercl, line 134: I believe, that this is the reason for failing checks - on Python 2.7 this syntax is not supported. That's why this module cannot be parsed and  subsequently imported anywhere.

pitercl (approved):

PR #284, opened 2020-10-08
Require bravado>=10.6.2

PiotrJander:


PiotrJander:
Pipelines fail on Python 3.6, not sure why https://travis-ci.org/github/neptune-ai/neptune-client/jobs/733934706

PR #287, opened 2020-10-22
Added Experiment.sync() and support for datetimes

szymon-kuklewicz:


neptune/alpha/attributes/atoms/datetime.py
```diff
@@ -0,0 +1,40 @@
+#
```
aniezurawski, line 1: Add some tests.

neptune/alpha/experiment.py
```diff
@@ -95,17 +101,59 @@ def close(self):
             self._bg_job.stop()
             self._op_processor.stop()
 
+    def sync(self, wait: bool = True):
```
aniezurawski, line 104: Let's support python type `datetime` in `experiment.define()`.
And add value type verification there because it's missing.

aniezurawski (changes requested):

neptune/alpha/types/atoms/datetime.py
```diff
@@ -27,7 +27,7 @@
 class Datetime(Atom):
 
     def __init__(self, value: datetime):
-        self.value = value
+        self.value = value.replace(microsecond=1000*int(value.microsecond/1000))
```
aniezurawski, line 30: We could make this conversion at the very late point in api layer, couldn't we?

neptune/alpha/internal/operation.py
```diff
@@ -142,19 +142,19 @@ class AssignDatetime(Operation):
 
     def __init__(self, path: List[str], value: datetime):
         super().__init__(path)
-        self.value = value
+        self.value = datetime.fromtimestamp(int(1000 * value.timestamp()) / 1000)
```
aniezurawski: We could make this conversion at the very late point in api layer, couldn't we?

aniezurawski (changes requested):

neptune/alpha/types/atoms/datetime.py
szymon-kuklewicz: I think the fact that we use millisecond resolution is part of our domain and should be enforced as fast as possible. It is not responsibility of api layer to make such conversions. If user runs script in offline mode and retrieves values after assignment, should he get different results? Should I put an extra conversion to cover this case ?

Also, it makes less sense IMO in tests. When we send a value to the server and then retrieve this value, we would expect to get the same exact value. By coding conversion here, we can have less tests to be aware of such conversion.

aniezurawski (approved):

PR #294, opened 2020-11-03
Merge master into new client

hubertjaworskineptune:


neptune/alpha/internal/backends/hosted_neptune_backend.py
```diff
@@ -64,18 +64,28 @@ def __init__(self, credentials: Credentials, proxies: Optional[Dict[str, str]] =
         if proxies is None:
             verify_host_resolution(config_api_url)
 
-        self.backend_client = create_swagger_client(config_api_url + self.BACKEND_SWAGGER_PATH, self._http_client)
-        self._client_config = self._get_client_config(self.backend_client)
+        token_http_client = self._create_http_client(ssl_verify, proxies)
```
aniezurawski, line 67: Why do we need different http client?

hubertjaworskineptune: YES. otherwise authenticator (attached to http client) will try to make a http call using it and it tries to attach authentication, hence stack overflow (tested)

aniezurawski (approved):

PR #295, opened 2020-11-03
Sets User-Agent before calling any endpoint

PiotrJander:


neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -87,6 +87,12 @@ def __init__(self, api_token=None, proxies=None):
         # for session re-creation we need to keep an authenticator-free version of http client
         self._http_client_for_token = RequestsClient(ssl_verify=ssl_verify)
 
+        user_agent = 'neptune-client/{lib_version} ({system}, python {python_version})'.format(
```
hubertjaworskineptune, line 90: This is not sufficient. You need to set user-agent to second https client before it requests config as well

hubertjaworskineptune (changes requested):

hubertjaworskineptune (dismissed):

hubertjaworskineptune (approved):

PR #296, opened 2020-11-04
Enables Mac and Windows on CI

PiotrJander:


tests/neptune/internal/channels/test_channels_values_sender.py
```diff
@@ -220,7 +220,7 @@ def test_send_when_waiting_for_next_value_timed_out(self):
             channels_values_sender.send(self._NUMERIC_CHANNEL.name, self._NUMERIC_CHANNEL.channelType, channel_value)
 
         # and
-        time.sleep(self.__TIMEOUT * 2)
+        time.sleep(self.__TIMEOUT * 10)
```
hubertjaworskineptune, line 223: Arbitrarly increasing timeout doesn't solve this issue in general, just hides the problem. Find the real reason why more sleep is required & fix instead of this

hubertjaworskineptune (changes requested):

tests/neptune/internal/channels/test_channels_values_sender.py
```diff
@@ -212,18 +213,21 @@ def test_send_when_waiting_for_next_value_timed_out(self):
...
         # when
         for channel_value in numeric_values:
             channels_values_sender.send(self._NUMERIC_CHANNEL.name, self._NUMERIC_CHANNEL.channelType, channel_value)
 
-        # and
-        time.sleep(self.__TIMEOUT * 2)
-
         # then
         # pylint: disable=protected-access
+        semaphore.acquire()
```
hubertjaworskineptune, line 230: nice solution, is there a way to make it so that after some time the test fails? otherwise "failure" will mean indefinite blocking

hubertjaworskineptune (approved):

PR #299, opened 2020-11-08
Adds pinging

PiotrJander:


neptune/alpha/internal/init_impl.py
```diff
@@ -103,7 +104,7 @@ def init(
     else:
         raise ValueError('connection_mode should be on of ["async", "sync", "offline", "debug"]')
 
-    background_jobs = []
+    background_jobs = [PingBackgroundJob()]
```
aniezurawski, line 107: Let's do it only if hardware metrics are disables.

aniezurawski (changes requested):

aniezurawski (approved):

PR #300, opened 2020-11-09
Waj/float series config

aniezurawski:


neptune/alpha/internal/operation.py
```diff
@@ -351,6 +351,39 @@ def from_dict(data: dict) -> 'ClearImageLog':
         return ClearImageLog(data["path"])
 
 
+class ConfigFloatSeries(Operation):
```
PiotrJander, line 354: Use a `@dataclass` here to eliminate boilerplate?

PiotrJander (approved):

PR #303, opened 2020-11-17
Add save_files()

szymon-kuklewicz:


aniezurawski (approved):

neptune/alpha/internal/backends/hosted_file_operations.py
```diff
@@ -63,7 +63,12 @@ def upload_file_attributes(swagger_client: SwaggerClient,
 
     expanded_files = set()
     for file_glob in file_globs:
-        expanded_files |= set(glob(file_glob))
+        expanded_file_glob = set(glob(file_glob))
+        if expanded_file_glob:
+            expanded_files |= expanded_file_glob
+        else:
+            result.append(FileUploadError(file_glob, "No files found for a path."))
```
aniezurawski, line 70: Shouldn't be an error if single glob is empty. I should be able to do `save_files(["*.py", "*.txt"])` if there are not `txt` files.

Moreover, it probably should create an empty file set if all globs are empty or empty list is given. It it possible to create empty archive? Will it work on backend side?

aniezurawski (approved):

PR #304, opened 2020-11-18
Implement graceful shutdown of experiment

aniezurawski:


neptune/alpha/experiment.py
```diff
@@ -61,10 +64,24 @@ def __init__(
...
+
+    def stop(self, seconds: Optional[float] = None):
+        if not self._started:
+            return
+        self._started = False
+        ts = time.time()
+        self._bg_job.stop()
+        self._bg_job.join(seconds)
+        with self._lock:
+            sec_left = None if not seconds else seconds - (time.time() - ts)
```
szymon-kuklewicz, line 83: i don't know what is further expected semantic, but it would be more consistent to use if seconds is None instead of not seconds.

aniezurawski: Done.

neptune/alpha/internal/backgroud_job_list.py
```diff
@@ -34,6 +35,9 @@ def stop(self):
         for job in self._jobs:
             job.stop()
 
-    def join(self):
+    def join(self, seconds: Optional[float] = None):
+        ts = time.time()
+        sec_left = seconds
```
szymon-kuklewicz, line 40: This line is not necessary

neptune/alpha/internal/containers/disk_queue.py
```diff
@@ -40,6 +41,7 @@ def __init__(
         self._to_dict = to_dict
         self._from_dict = from_dict
         self._max_file_size = max_file_size
+        self._event_empty = Event()
```
szymon-kuklewicz, line 44: By default event_empty is false (cleared). Should be set to true on start?

neptune/alpha/internal/threading/daemon.py
```diff
@@ -32,12 +32,16 @@ def interrupt(self):
...
+
     def run(self):
-        while True:
+        while True and not self._interrupted:
             self.work()
             if self._sleep_time > 0 and not self._interrupted:
                 self._event.wait(timeout=self._sleep_time)
                 self._event.clear()
+
             if self._interrupted:
```
szymon-kuklewicz, line 45: This condition is not necessary since You added negation to while condition

szymon-kuklewicz (approved):

PR #308, opened 2020-11-19
Create unique data directroy for every init call

aniezurawski:


neptune/alpha/internal/init_impl.py
```diff
@@ -99,21 +101,27 @@ def init(
...
             backend,
-            SyncOffsetFile(Path(experiment_path)),
+            SyncOffsetFile(Path(execution_path)),
             sleep_time=flush_period)
     elif connection_mode == "sync":
         operation_processor = SyncOperationProcessor(exp.uuid, backend)
     elif connection_mode == "debug":
         operation_processor = SyncOperationProcessor(exp.uuid, backend)
     elif connection_mode == "offline":
+        # Experiment was returned by mocked backend and has some random UUID.
```
PiotrJander, line 124: Now that we raise `ExperimentNotFound` in mock_backend's get_experiment, this is no longer true, is it?

neptune/alpha/sync.py
```diff
@@ -182,34 +185,38 @@ def list_experiments(path: Path, synced_experiments: Sequence[Experiment],
...
 
-    disk_queue = DiskQueue(str(path), OPERATIONS_DISK_QUEUE_PREFIX,
+def sync_execution(execution_path: Path, experiment_uuid: uuid.UUID, qualified_experiment_name: str) -> None:
+    disk_queue = DiskQueue(str(execution_path), OPERATIONS_DISK_QUEUE_PREFIX,
                            VersionedOperation.to_dict, VersionedOperation.from_dict)
-    sync_offset_file = SyncOffsetFile(path)
+    sync_offset_file = SyncOffsetFile(execution_path)
     sync_offset = sync_offset_file.read() or 0
 
     while True:
```
PiotrJander, line 222: Printing "Synchronization of experiment..." below is no longer accurate, need to fix.

PiotrJander (changes requested):
Also tests need to be fixed

neptune/alpha/internal/init_impl.py
aniezurawski: I belive we use `create_experiment` method in happy path in offline mode.

PiotrJander (approved):

PR #312, opened 2020-11-23
Fix stopping experiment when async thread failed

aniezurawski:


neptune/alpha/internal/threading/daemon.py
```diff
@@ -35,12 +36,19 @@ def wake_up(self):
...
+
     def run(self):
-        while True and not self._interrupted:
-            self.work()
-            if self._sleep_time > 0 and not self._interrupted:
-                self._event.wait(timeout=self._sleep_time)
-                self._event.clear()
+        self._is_running = True
+        try:
+            while True and not self._interrupted:
```
szymon-kuklewicz, line 45: you can skip "and True"

neptune/alpha/internal/threading/daemon.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, sleep_time: float):
         self._sleep_time = sleep_time
         self._interrupted = False
         self._event = threading.Event()
+        self._is_running = False
```
szymon-kuklewicz, line 27: I would created 3-state enum instead 2 booleans interrupted and is_running but won't force it if you don't have time for it

neptune/alpha/internal/operation_processors/async_operation_processor.py
```diff
@@ -71,10 +71,11 @@ def start(self):
...
-        self._consumer.wake_up()
-        self._queue.wait_for_empty(seconds)
-        self._consumer.interrupt()
+        if self._consumer.is_running():
+            self._consumer.disable_sleep()
+            self._consumer.wake_up()
+            self._queue.wait_for_empty(seconds)
+            self._consumer.interrupt()
         sec_left = None if seconds is None else seconds - (time() - ts)
         self._consumer.join(sec_left)
```
szymon-kuklewicz, line 80: Why isn't _consumer.join inside is_running() condition above?

szymon-kuklewicz (approved):

PR #314, opened 2020-11-23
Fix logging into image series

aniezurawski:


neptune/alpha/internal/utils/json_file_splitter.py
```diff
@@ -43,14 +44,20 @@ def get(self) -> Optional[dict]:
                 return None
 
     def _read_data(self):
+        if self._start_pos > 0:
+            new_buffer = StringIO()
+            new_buffer.write(self._buffer.getvalue()[self._start_pos:])
```
szymon-kuklewicz, line 49: I don't know why are you reading every time whole buffer into str and then select part of it. You are reading here into str even this parts that you already read. This is very suboptimal.

neptune/alpha/internal/utils/json_file_splitter.py
```diff
@@ -43,14 +44,20 @@ def get(self) -> Optional[dict]:
...
-        self._start_pos = 0
+        while data:
+            self._buffer.write(data)
+            data = self._file.read(self.BUFFER_SIZE)
 
     def _decode(self):
-        self._start_pos = self._buffer.index("{", self._start_pos)
-        data, end = self._decoder.raw_decode(self._buffer, self._start_pos)
+        self._start_pos = self._buffer.getvalue().index("{", self._start_pos)
+        data, end = self._decoder.raw_decode(self._buffer.getvalue(), self._start_pos)
```
szymon-kuklewicz, line 61: Same here - even twice.

szymon-kuklewicz (approved):
Conditionally - rewrite reading/loading json into queue in next MR

PR #315, opened 2020-11-24
Improve json splitter

aniezurawski:


neptune/alpha/internal/utils/json_file_splitter.py
```diff
@@ -14,50 +14,75 @@
...
-        data = self._file.read(self.BUFFER_SIZE)
-        while data:
-            self._buffer.write(data)
+        if self._part_read < self.MAX_PART_READ:
+            data = self._file.read(self.BUFFER_SIZE)
+            if not data:
+                return
+            if self._part_read > 0:
+                data = self._reset_part_buffer() + data
+            self._decode(data)
```
szymon-kuklewicz, line 54: There is a bug here I think. You release the date from the buffer, concatenate with new loaded data and store it in data variable. Then you pass it to the _decode method, which I think it is supposed to store the rest of unparsed data in the buffer since in 57-th line you overwrite data variable and previous data would be lost. However as I analyze _decode method, it does not always store rest of unparsed data, in case when first _json_start returns len(data).

neptune/alpha/internal/utils/json_file_splitter.py
```diff
@@ -14,50 +14,75 @@
...
+                json_data, start = self._decoder.raw_decode(data, start)
+                start = self._json_start(data, start)
+                self._parsed_queue.append(json_data)
+            except JSONDecodeError:
+                self._part_read = len(data) - start
+                self._part_buffer.write(data[start:])
+                break
+
+    @staticmethod
+    def _json_start(data: str, start: int = 0) -> int:
```
szymon-kuklewicz, line 77: I think it would be more clear if this method returned Optional[int]

neptune/alpha/internal/utils/json_file_splitter.py
```diff
@@ -14,50 +14,75 @@
...
+
+    def _decode(self, data: str):
+        start = self._json_start(data)
+        while start < len(data):
+            try:
+                json_data, start = self._decoder.raw_decode(data, start)
+                start = self._json_start(data, start)
+                self._parsed_queue.append(json_data)
+            except JSONDecodeError:
+                self._part_read = len(data) - start
```
szymon-kuklewicz, line 72: I know that = is enough but to be future proof += would better

neptune/alpha/internal/utils/json_file_splitter.py
```diff
@@ -14,50 +14,75 @@
...
-    # TODO: experiment with larger buffer sizes
     BUFFER_SIZE = 64 * 1024
+    MAX_PART_READ = 8 * 1024
 
     def __init__(self, file_path: str):
         self._file = open(file_path, "r")
         self._decoder = json.JSONDecoder(strict=False)
-        self._buffer = StringIO()
+        self._part_buffer = StringIO()
+        self._part_read = 0
```
szymon-kuklewicz, line 32: Why do you have such a variable instead of using self._buffer.len ?

neptune/alpha/internal/utils/json_file_splitter.py
aniezurawski: Why?

neptune/alpha/internal/utils/json_file_splitter.py
aniezurawski: It is very convinient to return string's length from `indexof` method. I think only python behave wierdly here. But, ok.

neptune/alpha/internal/utils/json_file_splitter.py
aniezurawski: Built-in `len` function does not work on StringIO and cannot found any documentation of `len` attribute.

szymon-kuklewicz (approved):

PR #317, opened 2020-11-25
Upload source files

aniezurawski:


neptune/alpha/attributes/file_set.py
```diff
@@ -37,10 +36,16 @@ def assign(self, value: Union[FileSetVal, str, Sequence[str]], wait: bool = Fals
...
 
     def save_files(self, globs: Union[str, Sequence[str]], wait: bool = False) -> None:
-        self.assign(FileSetVal(globs), wait)
+        if isinstance(globs, str):
+            globs = [globs]
+        else:
+            verify_collection_type("globs", globs, str)
+        with self._experiment.lock():
+            abs_file_globs = list(os.path.abspath(file_glob) for file_glob in globs)
+            self._enqueue_operation(UploadFileSet(self._path, abs_file_globs, reset=False), wait)
```
szymon-kuklewicz, line 48: I would extract a private method to be called from both places.

neptune/alpha/internal/operation.py
```diff
@@ -133,18 +133,20 @@ def from_dict(data: dict) -> 'UploadFile':
 class UploadFileSet(Operation):
 
     file_globs: Sequence[str]
+    reset: bool
```
szymon-kuklewicz, line 136: If you add reset to UploadFileSet, you should handle these parameter in backend mock to add file_globs instead replacing.

szymon-kuklewicz (approved):

PR #322, opened 2020-11-27
Implement missing getters for attributes

aniezurawski:


neptune/alpha/internal/backends/api_model.py
```diff
@@ -91,3 +92,33 @@ class Attribute:
     def __init__(self, _path: str, _type: AttributeType):
         self.path = _path
         self.type = _type
+
+
+@dataclass
+class FloatAttribute:
```
szymon-kuklewicz, line 98: Shouldn't it be named as FloatAttributeValue or something? Name suggest it is subclass of Attribute

aniezurawski: We can add name and type later if we need to, because it's returned from backend. But it was not necessary for now.

aniezurawski: There can be more then just value as well. For example flaot series' configuration.

szymon-kuklewicz (approved):

PR #325, opened 2020-12-01
Prints warning from server in domain 2

PiotrJander:


neptune/alpha/internal/backends/hosted_neptune_backend.py
```diff
@@ -102,7 +103,11 @@ def get_project(self, project_id: str) -> Project:
         verify_type("project_id", project_id, str)
 
         try:
-            project = self.backend_client.api.getProject(projectIdentifier=project_id).response().result
+            response = self.backend_client.api.getProject(projectIdentifier=project_id).response()
+            warning = response.metadata.headers.get('X-Server-Warning')
```
hubertjaworskineptune, line 107: Can we use API that can handle list of `X-Server-Wargning` s?

hubertjaworskineptune (changes requested):

PR #331, opened 2020-12-03
Add custom_experiment_id

szymon-kuklewicz:


neptune/alpha/internal/init_impl.py
```diff
@@ -122,7 +124,12 @@ def init(
         exp = backend.get_experiment(project + '/' + experiment)
     else:
         git_ref = get_git_info(discover_git_repo_location())
-        exp = backend.create_experiment(project_obj.uuid, git_ref)
+        if not custom_experiment_id:
+            custom_experiment_id = os.getenv(CUSTOM_EXP_ID_ENV_NAME)
```
aniezurawski, line 128: `custom_experiment_id = custom_experiment_id or os.getenv(CUSTOM_EXP_ID_ENV_NAME)`

neptune/alpha/internal/init_impl.py
```diff
@@ -122,7 +124,12 @@ def init(
...
+        if custom_experiment_id and len(custom_experiment_id) > 32:
+            _logger.warning('Given custom_experiment_id exceeds 32 characters and it will be ignored.')
```
aniezurawski, line 130: What is a type of column in db? I would like to have at least 36 characters so users can use uuids.

aniezurawski (changes requested):

aniezurawski (approved):

PR #336, opened 2020-12-04
Add custom monitoring namespaces

szymon-kuklewicz:
Tested scenario with monitoring to one namespace and in second run (with resume) capturing to second namespace. Works well.


neptune/alpha/internal/init_impl.py
```diff
@@ -74,6 +74,7 @@ def init(
         capture_stdout: bool = True,
         capture_stderr: bool = True,
         capture_hardware_metrics: bool = True,
+        capture_namespace: str = "monitoring",
```
aniezurawski, line 77: `monitoring_namespace` would be better name, whouldn't it? What do you think?

aniezurawski (approved):

PR #345, opened 2020-12-09
Support collections in log and single value in add/remove

aniezurawski:


neptune/alpha/experiment.py
```diff
@@ -133,6 +133,9 @@ def set_attribute(self, path: str, attribute: Attribute) -> Optional[Attribute]:
         with self._lock:
             return self._structure.set(parse_path(path), attribute)
 
+    def exists(self, path: str) -> bool:
+        return self.get_attribute(path) is not None
```
szymon-kuklewicz, line 137: Suggestion to define `__bool__` instead.

neptune/alpha/handler.py
```diff
@@ -73,26 +75,39 @@ def save_files(self, value: Union[str, Iterable[str]], wait: bool = False) -> No
...
                     attr = ImageSeries(self._experiment, parse_path(self._path))
                 attr.log(value, step=step, timestamp=timestamp, wait=wait)
                 self._experiment.set_attribute(self._path, attr)
             else:
                 attr.log(value, step=step, timestamp=timestamp, wait=wait)
 
-    def add(self, values: Iterable[str], wait: bool = False) -> None:
-        verify_collection_type("values", values, str)
+    def add(self, values: Union[str, Iterable[str]], wait: bool = False) -> None:
+        verify_type("values", values, (str, Iterable))
```
szymon-kuklewicz, line 110: verify_collection_type should be called for a collection

neptune/alpha/experiment.py
aniezurawski: __bool__ don't take arguments and is less intuitive for users.

neptune/alpha/handler.py
aniezurawski: It is verified in attr.add()

szymon-kuklewicz (approved):

PR #348, opened 2020-12-15
Add handling of entrypoint

szymon-kuklewicz:


neptune/alpha/internal/init_impl.py
```diff
@@ -203,10 +203,14 @@ def init(
     if capture_stderr and not _experiment.exists(stderr_path):
         _experiment.define(stderr_path, StringSeries([]))
 
-    if source_files is None:
-        if not is_ipython() and os.path.isfile(sys.argv[0]):
-            _experiment["source_code/files"].save_files(sys.argv[0])
-    else:
+    entrypoint = sys.argv[0] or None if not is_ipython() else None
```
aniezurawski, line 206: If source_files is not None and entrypoint file is in source_files then it should be path relative to source_files root (not to cwd).

neptune/alpha/internal/utils/__init__.py
```diff
@@ -53,3 +55,22 @@ def is_collection(var) -> bool:
...
+def get_common_root(absolute_paths: List[str]) -> Optional[str]:
+    try:
+        common_root = os.path.commonpath(absolute_paths)
+        if os.path.isfile(common_root):
+            common_root = os.path.dirname(common_root)
+        if common_root.startswith(os.getcwd() + os.sep):
+            common_root = os.getcwd()
+        return common_root
+    except ValueError:
+        return None
```
aniezurawski, line 76: Could we just return `/`? It is common root for everything and should let us avoid some if-statements.

neptune/alpha/internal/init_impl.py
```diff
@@ -203,13 +198,18 @@ def init(
...
+    if not is_ipython() and os.path.isfile(sys.argv[0]):
+        if source_files is None:
+            entrypoint = os.path.basename(sys.argv[0])
+            source_files = sys.argv[0]
+        else:
+            common_root = get_common_root(get_absolute_paths(source_files))
+            if common_root is not None:
+                entrypoint = normalize_file_name(os.path.relpath(os.path.abspath(sys.argv[0]), common_root))
+            else:
+                entrypoint = os.path.basename(sys.argv[0])
```
aniezurawski, line 210: Entrypoint may still be in sources even if no common root can be found so should probably use absolute path as we do for all source paths, shouldn't we?

tests/neptune/alpha/test_client.py
```diff
@@ -87,6 +87,44 @@ def test_resume(self):
...
+    @patch('neptune.alpha.internal.utils.os.getcwd', new=lambda: "/home/user/main_dir")
+    def test_entrypoint(self):
+        exp = init(connection_mode='debug')
+        self.assertEqual(exp["source_code/entrypoint"].get(), "main.py")
+
+        exp = init(connection_mode='debug', source_files=[])
+        self.assertEqual(exp["source_code/entrypoint"].get(), "main.py")
+
+        exp = init(connection_mode='debug', source_files=["../*"])
+        self.assertEqual(exp["source_code/entrypoint"].get(), "main_dir/main.py")
```
aniezurawski, line 104: If I'm correct there will be one absolute path in sources: `/home/user/main_dir/file.txt` with common root `/home/user/main_dir/`. So entrypoint would be just `main.py`. Where am I wrong?

tests/neptune/alpha/test_client.py
```diff
@@ -87,6 +87,44 @@ def test_resume(self):
...
+
+        exp = init(connection_mode='debug', source_files=["internal/*"])
+        self.assertEqual(exp["source_code/entrypoint"].get(), "main.py")
```
aniezurawski, line 107: Again according to my understanding entrypoint should be `../main.py`.

aniezurawski (changes requested):

tests/neptune/alpha/test_client.py
szymon-kuklewicz: source_files are ../*, so it generates one absolute_path /home/user/file.txt (in parent directory of main_dir). Common root is then calculated as /home/user/ .

tests/neptune/alpha/test_client.py
szymon-kuklewicz: It is not because there is a requirement that common root cannot be nested inside the working directory. If common root is calculated to be such, it is fixed to match the working directory. Because of that, effective entrypoint is still "main.py"

aniezurawski (approved):

PR #349, opened 2020-12-17
Add support for uploading streams

szymon-kuklewicz:


neptune/alpha/internal/containers/disk_queue.py
```diff
@@ -138,43 +138,54 @@ def wait_for_empty(self, seconds: Optional[float] = None) -> None:
...
+        self._last_ack_file.write(version)
 
     def is_empty(self) -> bool:
         return self.size() == 0
 
     def size(self) -> int:
         return self._last_put_file.read_local() - self._last_ack_file.read_local()
 
+    def _get_blob_file(self, index: int) -> str:
+        return "{}/data-{}.blob".format(self._dir_path, index)
```
aniezurawski, line 157: 1) Let's make a dedicated directory for blobs.
2) In future there can be more then one file for a signle operation (file sets, image series) so we should create a directory per operation.
3) It is not a good idea to keep everything in single directory if we don't know how many entries we can expect to be there. W need some more complex hierarchy of directories.

neptune/alpha/internal/containers/disk_queue.py
```diff
@@ -138,43 +138,54 @@ def wait_for_empty(self, seconds: Optional[float] = None) -> None:
...
-        self._last_ack_file.write(version)
-        log_versions = self._get_all_file_versions()
+        log_versions = self._get_all_log_file_versions()
         for i in range(0, len(log_versions) - 1):
             if log_versions[i + 1] <= version:
                 os.remove(self._get_log_file(log_versions[i]))
             else:
                 break
+        self._remove_old_blob_files(version)
+        self._last_ack_file.write(version)
```
aniezurawski, line 148: Why did you moved it here? It seems to be safer to keep it as first operation.

neptune/alpha/internal/operation.py
```diff
@@ -114,19 +117,27 @@ def from_dict(data: dict) -> 'AssignDatetime':
...
+    stream: Optional[IOBase] = None
 
     def accept(self, visitor: 'OperationVisitor[Ret]') -> Ret:
         return visitor.visit_upload_file(self)
 
-    def to_dict(self) -> dict:
-        ret = super().to_dict()
+    def to_dict(self, blob_file_supplier: Callable[[], str]) -> dict:
+        ret = super().to_dict(blob_file_supplier)
+        if not self.file_path:
```
aniezurawski, line 129: `if self.stream is not None` will be more clear.

aniezurawski (changes requested):

neptune/alpha/internal/utils/__init__.py
```diff
@@ -74,3 +86,35 @@ def get_common_root(absolute_paths: List[str]) -> Optional[str]:
...
+            chunk = stream.read(BYTES_IN_ONE_MB)
+
+
+STREAM_SIZE_LIMIT_MB = 15
+
+
+def get_stream_content(stream: IOBase) -> (Optional[str], str):
+    if stream.seekable():
+        stream.seek(0)
+    content = stream.read()
```
aniezurawski, line 110: Do not read whole stream at once. Read up to `STREAM_SIZE_LIMIT_MB * 1024 * 1024` bytes only and check if there is no data left.

aniezurawski (changes requested):

aniezurawski (approved):

PR #355, opened 2021-01-07
Fix variable's path in error description

szymon-kuklewicz:


neptune/alpha/internal/experiment_structure.py
```diff
@@ -32,10 +32,10 @@ def get_structure(self) -> Dict[str, Any]:
     def get(self, path: List[str]) -> Optional[T]:
         ref = self._structure
 
-        for part in path:
+        for index, part in enumerate(path):
             if not isinstance(ref, dict):
                 raise MetadataInconsistency("Cannot access path '{}': '{}' is already defined as an attribute, "
-                                            "not a namespace".format(path_to_str(path), part))
+                                            "not a namespace".format(path_to_str(path), path[index-1]))
```
aniezurawski, line 38: What if index is 0?

aniezurawski: We should print whole path of an attribute, not just last part.

szymon-kuklewicz: Index cannot be 0 because for 0 ref is always dict (root structure)

aniezurawski (approved):

PR #360, opened 2021-01-12
Adds forward compatibility with PyJWT==2

PiotrJander:


neptune/oauth.py
```diff
@@ -66,7 +69,7 @@ def _refresh_token(self):
     def _refresh_session_token(self):
         self.session.refresh_token(self.session.auto_refresh_url, verify=self.session.verify)
         if self.session.token is not None and self.session.token.get('access_token') is not None:
-            decoded_json_token = jwt.decode(self.session.token.get('access_token'), verify=False)
```
hubertjaworskineptune, line 69: Make sure that this works on ALL environments and why was this given here. It's VERY easy to break authentication

hubertjaworskineptune: algorithm is no longer required in 2.0?

hubertjaworskineptune (approved):

PR #364, opened 2021-01-13
Dev/client integration with alpha

shnela:


neptune/internal/backends/backend_factory.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def backend_factory(*, backend_name, api_token=None, proxies=None) -> Backend:
+    if backend_name == 'offline':
+        return OfflineBackend()
+
+    elif backend_name is None:
+        credentials = Credentials(api_token)
+        # TODO: Improvement. How to determine which backend class should be used?
+        if credentials.token_origin_address.startswith('https://alpha.'):
```
shnela, line 34: Do you know if we can detect alpha tokens better?

neptune/__init__.py
```diff
@@ -136,27 +136,22 @@ def init(project_qualified_name=None, api_token=None, proxies=None, backend=None
 
     if project_qualified_name is None:
```
shnela, line 137: Right now we're not accepting string which is `None`, but empty string is ok.
Does empty string make sense? If we'd force non empty string, the code could be simplified.

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -0,0 +1,149 @@
...
+from neptune.exceptions import STYLES
+from neptune.internal.backends.hosted_neptune_backend import HostedNeptuneBackend
+from neptune.internal.utils.http import extract_response_field
+from neptune.projects import Project
+from neptune.utils import with_api_exceptions_handler
+
+_logger = logging.getLogger(__name__)
+
+
+class AlphaIntegrationBackend(HostedNeptuneBackend):
```
shnela, line 35: Right now we simply inherit from 'HostedNeptuneBackend', but it's likely that we'll create own implementation based on abstract 'Backend' in the future.

PR #367, opened 2021-01-15
One space missing in exception

wuchangsheng951:
image
need one space between "or" and 3


aniezurawski (approved):

pitercl:
Thanks for the catch @wuchangsheng951 :smile:

PR #371, opened 2021-01-20
Initial fake alpha system channels implementation

shnela:


neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -50,6 +67,31 @@ def get_project(self, project_qualified_name):
...
+
+        # Assign source entrypoint
+        init_operations.append(alpha_operation.AssignString(
+            path=['source_code', 'entrypoint'],
+            value=entrypoint,
+        ))
+        # Assign experiment parameters
+        for p_name, p_val in params.items():
+            parameter_type, string_value = self._get_parameter_with_type(p_val)
+            operation_cls = alpha_operation.AssignFloat if parameter_type == 'double' else alpha_operation.AssignString
```
shnela, line 82: In old client everything but numbers was converted to string, but in alpha we have `AssignDatetime` operation and series and sets which may be used as well.

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -122,25 +169,100 @@ def create_experiment(self,
...
+            #     timestampMillis=int(value.ts * 1000.0),
+            #     x=value.x,
+            #     y=Y(numericValue=value.y.get('numeric_value'),
+            #         textValue=value.y.get('text_value'),
+            #         inputImageValue=value.y.get('image_value'))
+            # ) for value in channel_with_values.channel_values]
+            ch_values = [
+                alpha_operation.LogStrings.ValueType(
+                    value=value.y.get('text_value'),
+                    step=None,  # TODO: what is step?
```
shnela, line 187: `step`, do we need to set this value?

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -122,25 +169,100 @@ def create_experiment(self,
...
+        dummy_log_string = alpha_operation.LogStrings(
+            path=alpha_path_utils.parse_path(f'{MONITORING_ATTRIBUTE_SPACE}{name}'),
+            values=[],
+        )
+        # pylint: disable=unused-variable
+        errors = self._alpha_backend.execute_operations(
+            experiment_uuid=uuid.UUID(experiment.internal_id),
+            operations=[dummy_log_string],
+        )
+        system_channels = self.get_system_channels(experiment)
```
shnela, line 239: Additional api call is required to construct, `AlphaChannelWithLastValue`.

Eventually we may assume that if there's no errors, operation completed, and create `AlphaChannelWithLastValue` without asking server for data.

neptune/alpha/internal/backends/hosted_neptune_backend.py
```diff
@@ -228,8 +228,8 @@ def _execute_operations(self,
         try:
             result = self.leaderboard_client.api.executeOperations(**kwargs).response().result
             return [MetadataInconsistency(err.errorDescription) for err in result]
-        except HTTPNotFound:
-            raise ExperimentUUIDNotFound(exp_uuid=experiment_uuid)
+        except HTTPNotFound as e:
```
shnela, line 231: Do we want to update all exception raising code in `except` blocks now?

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -91,21 +152,20 @@ def create_experiment(self,
...
-            }
-
-            kwargs = {
-                'experimentCreationParams': params,
-                'X-Neptune-CliVersion': self.client_lib_version,
-            }
             api_experiment = self.leaderboard_swagger_client.api.createExperiment(**kwargs).response().result
-
-            return self._convert_to_experiment(api_experiment, project)
         except HTTPNotFound:
```
shnela, line 169: Does alpha backend use the same exceptions for quota?

aniezurawski: I'm not sure if it even recognize quota limit. Verify it and if it's not supported and task to jira.

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -122,25 +184,101 @@ def create_experiment(self,
...
+            #     x=value.x,
+            #     y=Y(numericValue=value.y.get('numeric_value'),
+            #         textValue=value.y.get('text_value'),
+            #         inputImageValue=value.y.get('image_value'))
+            # ) for value in channel_with_values.channel_values]
+            ch_values = [
+                alpha_operation.LogStrings.ValueType(
+                    value=value.y.get('text_value'),
+                    step=None,
+                    ts=int(value.ts),
```
aniezurawski, line 211: Make sure we should not multiply `value.ts` by 1000. We probably should or we should not convert it to int. Api expects millis. However I do not remember what is expected in `LogStrings.ValueType`.

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -122,25 +184,101 @@ def create_experiment(self,
...
+        try:
+            experiment = self.leaderboard_swagger_client.api.getExperimentAttributes(**params).response().result
+        except HTTPNotFound:
+            # pylint: disable=protected-access
+            raise ExperimentNotFound(
+                experiment_short_id=experiment.id, project_qualified_name=experiment._project.full_id)
+        return [
+            AlphaChannelWithLastValue(
+                ch_id=attr.stringSeriesProperties.attributeName,
+                ch_name=alpha_path_utils.parse_path(attr.stringSeriesProperties.attributeName)[-1],
```
aniezurawski, line 236: Shouldn't be `[:-1]` insted of `[-1]`?

aniezurawski: Shouldn't `ch_name` be string instead of list?

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -122,25 +184,101 @@ def create_experiment(self,
...
+            'experimentId': experiment.internal_id,
+        }
+        try:
+            experiment = self.leaderboard_swagger_client.api.getExperimentAttributes(**params).response().result
+        except HTTPNotFound:
+            # pylint: disable=protected-access
+            raise ExperimentNotFound(
+                experiment_short_id=experiment.id, project_qualified_name=experiment._project.full_id)
+        return [
+            AlphaChannelWithLastValue(
```
aniezurawski, line 234: Why is this class necessary? Can't we use the old one?

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -15,24 +15,46 @@
...
+from neptune.internal.utils.alpha_integration import MONITORING_ATTRIBUTE_SPACE, PARAMETERS_ATTRIBUTE_SPACE
 from neptune.internal.utils.http import extract_response_field
+from neptune.model import AlphaChannelWithLastValue
 from neptune.projects import Project
 from neptune.utils import with_api_exceptions_handler
 
 _logger = logging.getLogger(__name__)
 
 
 class AlphaIntegrationBackend(HostedNeptuneBackend):
```
aniezurawski, line 53: Why is this inheritance necessary?

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -122,25 +184,101 @@ def create_experiment(self,
...
+        raise ChannelNotFound(channel_id=channel_id)
 
     def upload_experiment_source(self, experiment, data, progress_indicator):
+        # TODO: handle `FileChunkStream` or update `neptune.experiments.Experiment._start`
         pass
 
     @with_api_exceptions_handler
     def get_experiment(self, experiment_id):
-        return NonCallableMagicMock()
+        experiment = super().get_experiment(experiment_id)
```
aniezurawski, line 267: Is this a call to legacy backend? Why? Does it even work on new api?

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -122,25 +184,101 @@ def create_experiment(self,
...
+        fake_experiment = NonCallableMagicMock()
+        # `timeOfCreation` is required by `TimeOffsetGenerator`
+        fake_experiment.timeOfCreation = dateutil.parser.parse(experiment.creationTime)
```
aniezurawski, line 270: Nothing more is required?

neptune/model.py
```diff
@@ -52,6 +95,47 @@ def y(self, y):
         self.channel_with_value_dto.y = y
 
 
+class AlphaChannelWithLastValue(AbstractChannelWithLastValue):
+    def __init__(self, ch_id, ch_name, ch_type):
```
aniezurawski, line 99: Use type hint wherever you can.

aniezurawski (changes requested):

neptune/alpha/internal/backends/hosted_neptune_backend.py
shnela: Yes

neptune/internal/backends/alpha_integration_backend.py
shnela: That's something I've discussed earlier - it probably will be abandoned, but it's useful now:
https://github.com/neptune-ai/neptune-client/pull/364#discussion_r558226382

neptune/internal/backends/alpha_integration_backend.py
shnela: Old one requires `ChannelWithValueDTO` which has been removed from alpha `leaderboard`.
https://github.com/neptune-ai/neptune-client/blob/dev/client-integration-with-alpha/neptune/internal/backends/hosted_neptune_backend.py#L835

neptune/internal/backends/alpha_integration_backend.py
shnela: But it is string. `[-1]` returns last element from list, so example values used in our code could be:
```
ch_id='monitoring/stdout',
ch_name='stdout',
```

neptune/internal/backends/alpha_integration_backend.py
shnela: It actually called 'getExperiment', so it worked. But of course I shouldn't use old client code here.
I've just didn't pay much attention to implement this function yet.

neptune/internal/backends/alpha_integration_backend.py
shnela: Not yet, but in future we'll have to figure out where to get missing information from - new `getExperiment` endpoint doesn't return some of the values, old one returned.
[Usage of get_experiment](https://github.com/neptune-ai/neptune-client/blob/dev/client-integration-with-alpha/neptune/experiments.py#L166).

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -91,21 +154,20 @@ def create_experiment(self,
             "remotes": git_info.remote_urls
         } if git_info else None
 
+        api_params = {
+            "projectIdentifier": str(project.internal_id),
+            "cliVersion": self.client_lib_version,
+            "gitInfo": git_info,
+            "customId": name,
```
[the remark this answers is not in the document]
shnela: Ok, so `name` parameter should be rather used for setting 'sys/name' attribute, right?

neptune/internal/backends/alpha_integration_backend.py
shnela: Good catch! Sorry for that...

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -50,6 +72,47 @@ def get_project(self, project_qualified_name):
...
+        except alpha_exceptions.InternalClientError as e:
+            raise NeptuneException(e) from e
+
+    def _get_init_experiment_operations(self, entrypoint, params, tags) -> List[alpha_operation.Operation]:
+        """Returns operations required to initialize newly created experiment"""
+        init_operations = list()
+
+        # Assign source entrypoint
+        init_operations.append(alpha_operation.AssignString(
+            path=['source_code', 'entrypoint'],
```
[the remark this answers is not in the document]
shnela: Right

neptune/internal/backends/alpha_integration_backend.py
shnela: Ok, I'll check that

shnela: Simplified to extent used in alpha when creating new experiment.

neptune/alpha/internal/backends/hosted_neptune_backend.py
aniezurawski: It would be good to have it but I do not see a reason to do it everywhere in this merge request. Let's keep it clean.

neptune/internal/backends/alpha_integration_backend.py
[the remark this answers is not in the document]
aniezurawski: Yes.

neptune/internal/backends/alpha_integration_backend.py
aniezurawski: Ok.

neptune/internal/backends/alpha_integration_backend.py
aniezurawski: Ok.

neptune/internal/backends/alpha_integration_backend.py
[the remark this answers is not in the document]
shnela: So it's done

neptune/internal/backends/alpha_integration_backend.py
aniezurawski: From `getExperimentAtributes` I guess. But it is not to be considered in this merge request, is it?

shnela: > But it is not to be considered in this merge request

No, it's not.

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -94,7 +97,7 @@ def _get_init_experiment_operations(self, entrypoint, params, tags) -> List[alph
 
         # Assign source entrypoint
         init_operations.append(alpha_operation.AssignString(
-            path=['source_code', 'entrypoint'],
+            path=alpha_path_utils.parse_path(f'{SOURCE_CODE_ATTRIBUTE_SPACE}entrypoint'),
```
aniezurawski, line 103: It would be good to have constant for `entrypoint` as well since it's already used in more then one place in code.

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -178,7 +181,7 @@ def create_experiment(self,
         # Initialize new experiment
         self._execute_alpha_operation(
             experiment=experiment,
-            operations=self._get_init_experiment_operations(entrypoint, params, tags),
+            operations=self._get_init_experiment_operations(name, entrypoint, params, tags),
```
aniezurawski, line 184: There are more arguments that should be supported (e.g. description and hostname). I guess it is a good moment to create an epic in our Jira and add tasks for such things so we will not forget about it.

aniezurawski (changes requested):

neptune/internal/backends/alpha_integration_backend.py
shnela: We do, right.

neptune/internal/backends/alpha_integration_backend.py
shnela: > since it's already used in more then one place in code

Where?
If I've implemented 'entrypoint' constant in `neptune/internal/utils/alpha_integration.py`, the only place where I'd use it is here.

But I have nothing against  adding such constant, but then we should do the same for name.

I suggest
```python
SYSTEM_NAME_ATTRIBUTE_PATH = f'{SYSTEM_ATTRIBUTE_SPACE}name'
SOURCE_CODE_ENTRYPOINT_ATTRIBUTE_PATH = f'{SOURCE_CODE_ATTRIBUTE_SPACE}entrypoint'
```

Sounds good?

aniezurawski (approved):

PR #376, opened 2021-01-26
Jk/fake alpha user channels

shnela:


neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -174,39 +194,64 @@ def get_system_channels(self, experiment):
...
 
+    @with_api_exceptions_handler
+    def get_channels(self, experiment) -> Dict[str, AlphaChannelDTO]:
+        api_channels = [
+            channel for channel in self._get_channels(experiment)
+            # return channels from LOG_ATTRIBUTE_SPACE namespace only
+            if channel.id.startswith(alpha_consts.LOG_ATTRIBUTE_SPACE)
+        ]
+        channels = dict()
+        for ch in api_channels:
```
shnela, line 211: This code is analogous to [this experiment methods](https://github.com/neptune-ai/neptune-client/pull/376/files#diff-f9469dadddb4d894b4ab0167d6248ad4208edceaeb7f04f8e2a4c6791625b150L274).
But we're relying on `ch.lastX` alone. Do you think that we may need some alpha  equivalent of `channels_last_values_by_name`?

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -146,25 +151,40 @@ def upload_experiment_source(self, experiment, data, progress_indicator):
...
+        Since channels are abandoned in alpha api, we're mocking them using empty logging operation."""
+
+        if channel_type == ChannelType.TEXT.value:
+            operation = alpha_operation.LogStrings
+        elif channel_type == ChannelType.NUMERIC.value:
+            operation = alpha_operation.LogFloats
+        elif channel_type == ChannelType.IMAGE.value:
+            operation = alpha_operation.LogFloats
+        else:
+            raise NeptuneException(f"We're not supporting {channel_type} channel type.")
```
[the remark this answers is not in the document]
shnela: Right, I've created helper function responsible for raising error as well.

BTW `ChannelType.IMAGE -> alpha_operation.LogFloats` is fixed, but it's still not enough to make log images work.

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -146,25 +151,40 @@ def upload_experiment_source(self, experiment, data, progress_indicator):
...
     @with_api_exceptions_handler
-    def get_system_channels(self, experiment):
+    def create_channel(self, experiment, name, channel_type) -> ChannelWithLastValue:
+        channel_id = f'{alpha_consts.LOG_ATTRIBUTE_SPACE}{name}'
+        self._create_channel(experiment, channel_id, channel_type)
+        try:
+            channel = self.get_channels(experiment)[name]  # TODO: implement this
+            return self._convert_channel_to_channel_with_last_value(channel)
+        except KeyError:
+            raise ChannelNotFound(channel_id=channel_id)
```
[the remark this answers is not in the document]
shnela: Replaced it with `raise NeptuneException(f"Channel {channel_id} wasn't created.")`

neptune/internal/backends/alpha_integration_backend.py
shnela: > So, if I am correct, old code returned only USER channels, but y value was filled only for NUMERIC or TEXT channels.
> 
> Your code seems to operate on AlphaChannelDTO without value. Isn't it a problem?

No `AlphaChannelDTO` has `lastX` value which returns `last` of stored attribute (which is one of: `StringSeriesAttributeDTO`, `FloatSeriesAttributeDTO` or `ImageSeriesAttributeDTO`).
I've just noticed that `ImageSeriesAttributeDTO` doesn't have mentioned `last` value, so it's matching with old client behaviour.

I think that current implementation of `AlphaChannelDTO.lastX` may be good equivalent of `channels_last_values_by_name` in old client.

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -146,25 +151,40 @@ def upload_experiment_source(self, experiment, data, progress_indicator):
...
+        if channel_type == ChannelType.TEXT.value:
+            operation = alpha_operation.LogStrings
+        elif channel_type == ChannelType.NUMERIC.value:
+            operation = alpha_operation.LogFloats
+        elif channel_type == ChannelType.IMAGE.value:
+            operation = alpha_operation.LogFloats
+        else:
+            raise NeptuneException(f"We're not supporting {channel_type} channel type.")
+
+        log_empty_operation = operation(
```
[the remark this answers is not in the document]
shnela: Good point. Do you want those tests in this branch?

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -146,25 +151,40 @@ def upload_experiment_source(self, experiment, data, progress_indicator):
...
-                return channel
-        raise ChannelNotFound(channel_id=channel_id)
 
     @with_api_exceptions_handler
-    def get_system_channels(self, experiment):
+    def create_channel(self, experiment, name, channel_type) -> ChannelWithLastValue:
+        channel_id = f'{alpha_consts.LOG_ATTRIBUTE_SPACE}{name}'
+        self._create_channel(experiment, channel_id, channel_type)
+        try:
+            channel = self.get_channels(experiment)[name]  # TODO: implement this
```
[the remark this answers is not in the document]
shnela: 1. Good idea.
2. It looks like [this discussion](https://github.com/neptune-ai/neptune-client/pull/371/files#r561777573) gained some arguments for. I'll assume that attribute was created if there's no error in run operations
3. Good catch. Since we're decided to solution 2. the problem no longer exists. But such double wrapping with `_execute_alpha_operation` exists in case of `_execute_alpha_operation`. I've created an issue for that.

neptune/internal/backends/alpha_integration_backend.py
aniezurawski: `last` in api attribute is ment to be value, not a step (x). Unfortunatelly lastStep is missing in api.
I don't think it is important to have last step here. Is it used by our code or does it exists only to be returned to user? If we don't need it let's just add a bug to Jira and set it to None. However, it is important to return value to the user. So `AlphaChannelDTO` should have last value as well.

neptune/internal/backends/alpha_integration_backend.py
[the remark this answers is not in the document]
aniezurawski: Yes.

neptune/internal/utils/alpha_integration.py
```diff
@@ -64,6 +64,9 @@ def channelType(self):
 
     @property
     def lastX(self):
+        if self.channelType == ChannelType.IMAGE.value:
```
aniezurawski, line 67: X is step, not a value.

aniezurawski (changes requested):

neptune/internal/backends/alpha_integration_backend.py
shnela: Right...
I've renamed `lastX` to `lastY` in `AlphaChannelDTO` - it's additional field we may use for initializing `ch.y`. And I've updated `lastX` to always return None.
Right now code works, but I've created issue for fixing problem with missing X.

neptune/internal/utils/alpha_integration.py
shnela: yes... Fixed.

aniezurawski (approved):

PR #387, opened 2021-02-03
Jk/adapt standalone client functions to alpha

shnela:


neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -146,8 +147,63 @@ def get_experiment(self, experiment_id):
         fake_experiment = NonCallableMagicMock()
         # `timeOfCreation` is required by `TimeOffsetGenerator`
         fake_experiment.timeOfCreation = experiment.creationTime
+
+        try:
+            tags = self._alpha_backend.get_string_set_attribute(
```
aniezurawski, line 152: It is already returned in `self._get_attributes(experiment_id)` below so you don't need to do this api call.

neptune/internal/utils/alpha_integration.py
```diff
@@ -13,38 +13,82 @@
...
+        assert self._allowed_atribute_types is not None
+        if not self.is_valid_attribute(attribute):
             raise NeptuneException(f"Invalid channel attribute type: {attribute.type}")
 
         self._attribute = attribute
 
     @classmethod
-    def is_valid_attribute_for_channel(cls, attribute):
+    def is_valid_attribute(cls, attribute):
         """Checks if attribute can be used as channel"""
```
aniezurawski, line 45: This comment is not true anymore.

neptune/internal/utils/alpha_integration.py
```diff
@@ -13,38 +13,82 @@
...
+class AlphaPropertyDTO(AlphaAttributeWrapper):
+    """It's simple wrapper for `AttributeDTO` objects which uses alpha variables attributes to fake properties.
+
+    Alpha leaderboard doesn't have `KeyValueProperty` since it doesn't support properties at all,
+    so we do need fake `KeyValueProperty` class for backward compatibility with old client's code."""
+
+    _allowed_atribute_types = [
+        AlphaAttributeType.FLOAT.value,
+        AlphaAttributeType.STRING.value,
+        AlphaAttributeType.DATETIME.value,
```
aniezurawski, line 63: In contrast to parameters, properties was strings only in old Neptune.

aniezurawski (changes requested):

neptune/internal/utils/alpha_integration.py
shnela: Right, I didn't notice this at first, and then I forgot to fix it in this place...

neptune/internal/utils/alpha_integration.py
shnela: Right,

neptune/internal/backends/alpha_integration_backend.py
shnela: Indeed...

aniezurawski (approved):

PR #390, opened 2021-02-03
Jk/support handling artifacts

shnela:
Quota limits will be checked in separate issue: NPT-9248


neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -315,6 +310,42 @@ def ping_experiment(self, experiment):
...
+            if destination is not None:
+                operations = [alpha_operation.UploadFileContent(
+                    path=dest_path,
+                    file_name=dest_path[-1],
+                    file_content=base64_encode(artifact.read().encode('utf-8')),
+                )]
+            else:
+                raise ValueError("destination is required for file streams")
+        else:
+            raise ValueError("artifact is a local path or an IO object")
```
shnela, line 336: It's copy pasted from old client, but I think that it should be "Artifact must be a local path or an IO object".
Can you confirm?

neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -315,6 +310,42 @@ def ping_experiment(self, experiment):
...
+
+        self._execute_alpha_operation(experiment, operations)
+
+    def delete_artifacts(self, experiment, path):
+        self._remove_attribute(experiment, str_path=f'{alpha_consts.ARTIFACT_ATTRIBUTE_SPACE}{path}')
+
+    def upload_experiment_output(self, experiment, data, progress_indicator):
+        raise NeptuneException(f'upload_experiment_output should not be called in alpha version')
+
+    def extract_experiment_output(self, experiment, data):
```
aniezurawski, line 346: Shouldn't these two become private methods? How about `upload/extract_experiment_source`?

aniezurawski (changes requested):

neptune/internal/backends/alpha_integration_backend.py
shnela: +1

aniezurawski (approved):

PR #408, opened 2021-02-14
Jk/detect backend version

shnela:


neptune/internal/backends/backend_factory.py
```diff
@@ -13,28 +13,47 @@
...
+                                                          boot_http_client)
 
-        return HostedNeptuneBackend(api_token, proxies)
+        response = boot_backend_client.api.getProject(projectIdentifier=project_qualified_name).response()
+        warning = response.metadata.headers.get('X-Server-Warning')
+        if warning:
+            click.echo('{warning}{content}{end}'.format(content=warning, **STYLES))
+        project = response.result
+        if not hasattr(project, 'version'):
+            pass  # what now?
```
aniezurawski, line 51: It is version 1.

neptune/internal/backends/backend_factory.py
```diff
@@ -13,28 +13,47 @@
...
-        if credentials.token_origin_address.startswith('https://alpha.'):
-            return AlphaIntegrationBackend(api_token, proxies)
+        ssl_verify = alpha_check_if_ssl_verify()
+        boot_http_client = alpha_create_http_client(ssl_verify, proxies)
+        config_api_url = credentials.api_url_opt or credentials.token_origin_address
+        boot_backend_client = alpha_create_swagger_client(f'{config_api_url}/api/backend/swagger.json',
+                                                          boot_http_client)
 
-        return HostedNeptuneBackend(api_token, proxies)
+        response = boot_backend_client.api.getProject(projectIdentifier=project_qualified_name).response()
```
aniezurawski, line 45: Can we avoid additional round trips somehow? I guess we will make this calls (get swagger and getProject) twice since it will be called on returned backend in init function again.

aniezurawski (changes requested):

pylintrc
```diff
@@ -42,9 +42,10 @@ file-header=#\n# Copyright \(c\) (2019|2020|2021), Neptune Labs Sp\. z o\.o\.\n#
 # W0401, W0614 Allow wildcard imports.
 # W0511 Allow TODO/FIXME comments.
 # W0703 Allow too broad except clause (Exception).
+# W1202 Allow logging-format-interpolation.
```
shnela, line 45: Allow f-strings in logs formatter.
PyLint requirement for '%s' has to be some kind of joke...

More about it: https://stackoverflow.com/a/59027338/1565454

PR #410, opened 2021-02-18
Jk/init exp attributes

shnela:


alpha_integration_dev/common_client_code.py
```diff
@@ -20,10 +20,16 @@
...
+        'init_text_parameter': 'some text',
+        'init_number parameter': 42,
+        'init_list': [1, 2, 3],
+        'init_datetime': datetime.now()
+    }
+    properties = {
+        'init_text_property': 'some text',
+        'init_number property': 42,
+        # 'init_list': [1, 2, 3],  # TODO: Error 500 in old client
+        # 'init_datetime': datetime.now()  # TODO: Non serializable in old client
```
aniezurawski, line 32: It should work already.

shnela: Yes, on master.
But our alpha branch is way ahead/before master and has lots of conflicting changes.

If you want I can take a look on this, so we could merge master to alpha from time to time instead of duplicating fixes on master and alpha as I've had to do with unhandled limit error in old client.

aniezurawski (approved):

PR #416, opened 2021-02-24
[NPT-9216] Return dict consistent with docs from get_logs()

aniezurawski:
Closes NPT-9216

### DIFF
--- a/neptune/internal/utils/alpha_integration.py
+++ b/neptune/internal/utils/alpha_integration.py
@@ -140,12 +140,11 @@ class AlphaChannelDTO(AlphaAttributeWrapper):
             return ChannelType.IMAGE.value
 
     @property
-    def lastX(self):
-        # TODO: NPT-9216
-        return None
+    def x(self):
+        return self._properties.lastStep
 
     @property
-    def lastY(self):
+    def y(self):
         if self.channelType == ChannelType.IMAGE.value:
             # We do not store last value for image series
             return None
--- a/neptune/experiments.py
+++ b/neptune/experiments.py
@@ -976,7 +976,7 @@ class Experiment(object):
         """
 
         channels_data = {}
-        channels_by_name = self.get_channels()
+        channels_by_name = self._backend.get_channels(self)
         for channel_name in channel_names:
             channel_id = channels_by_name[channel_name].id
             try:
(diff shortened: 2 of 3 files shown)


alpha_integration_dev/old_client.py
```diff
@@ -106,8 +106,11 @@ def log_series(self):
         neptune.log_image('g_img', self.img_path)
 
         # see what we've logged
+        from time import sleep
+        sleep(5)
```
shnela, line 110: Is there any equivalent of `exp.sync()` in old client?

alpha_integration_dev/old_client.py
```diff
@@ -106,8 +106,11 @@ def log_series(self):
...
         logs = neptune.get_experiment().get_logs()
         print(f'Logs: {logs}')
+        assert logs == {'m1': '2.0', 'g_img': None, 'nested/m1': '1.0', 'm2': 'c'}
```
shnela, line 113: * Old API returns strings while new one integer/float values.
* No idea why `g_img` is `None`, but we have the same behavior for old client. `x` and `lastX` values for returned `g_img` channel is equal 1, so images has been uploaded.

neptune/internal/utils/alpha_integration.py
```diff
@@ -140,12 +140,11 @@ def channelType(self):
             return ChannelType.IMAGE.value
 
     @property
-    def lastX(self):
-        # TODO: NPT-9216
-        return None
+    def x(self):
+        return self._properties.lastStep
```
shnela, line 144: Here's no such value: `_properties.lastStep`.
None of `*SeriesAttributeDTO` contains it when returned from API.

shnela (changes requested):
Can you start PR title with `[NPT-9216]`, please?

alpha_integration_dev/old_client.py
aniezurawski: Nope.

neptune/internal/utils/alpha_integration.py
aniezurawski: Fetch and look again. ;)

alpha_integration_dev/old_client.py
aniezurawski: 1. Fixed
2. We do not return images in old client by design. It was always this way.

neptune/internal/utils/alpha_integration.py
shnela: Hmmm, still no `x`  value on `stage` (screen).
image

But I assume it'll work after deploying `[e6f65a16] Add last step to series attributes ` from backend repo.

shnela (approved):
Assuming `[e6f65a16] Add last step to series attributes ` will solve the problem with channel `x` attribute, it's ok.

PR #420, opened 2021-02-26
Do not use alpha backend in old client

shnela:


neptune/internal/backends/alpha_integration_backend.py
```diff
@@ -446,21 +447,108 @@ def _get_client_config_args(api_token):
...
+        if other_operations:
+            errors.extend(
+                self._execute_operations(experiment=experiment,
+                                         operations=other_operations)
+            )
+
+        if errors:
+            raise AlphaOperationErrors(errors)
+
+        return None
```
aniezurawski, line 551: Isn't it an overkill to copy all this logic here? Is there even a case when you need to perform upload operation and simple operation? These logic was necessary in alpha backend because we do not know what operations will be passed. But in this class you exactly know what do you want to perform:
upload_file_set_attribute - when uploading source code
upload_file_attribute - in case of artifacts
_execute_operations - in case of simple operations

aniezurawski (approved):

PR #423, opened 2021-03-01
Implement notebooks in alpha client

HubertJaworski:


neptune/alpha/internal/backends/hosted_neptune_backend.py
```diff
@@ -195,6 +201,18 @@ def create_experiment(self,
...
 
+    @with_api_exceptions_handler
+    def create_checkpoint(self, notebook_id: uuid.UUID, jupyter_path: str) -> Optional[uuid.UUID]:
+        checkpoint = self.leaderboard_client.createEmptyCheckpoint(
+            notebookId=notebook_id,
+            checkpoint={
+                "path": jupyter_path
+            }
+        ).response().result
+        if checkpoint is not None:
```
aniezurawski, line 212: Why do we have to check it for None?

neptune/alpha/internal/backends/neptune_backend_mock.py
```diff
@@ -114,6 +116,9 @@ def create_experiment(self,
             self._experiments[new_experiment_uuid].set(["source_code", "git"], git_ref)
         return ApiExperiment(new_experiment_uuid, short_id, 'workspace', 'sandbox', False)
 
+    def create_checkpoint(self, notebook_id: uuid.UUID, jupyter_path: str) -> Optional[uuid.UUID]:
+        return None
```
aniezurawski, line 120: How about NotImplementedError?

neptune/alpha/internal/init_impl.py
```diff
@@ -143,7 +145,25 @@ def init(project: Optional[str] = None,
...
+
+        notebook_path = None
+        if os.getenv(NEPTUNE_NOTEBOOK_PATH, None) is not None:
+            notebook_path = os.environ[NEPTUNE_NOTEBOOK_PATH]
+
+        checkpoint_id = None
+        if notebook_id is not None and notebook_path is not None:
+            checkpoint_id = create_checkpoint(backend=backend,
+                                              notebook_id=notebook_id,
+                                              notebook_path=notebook_path)
```
aniezurawski, line 164: Let's move this whole block to separate function.

aniezurawski (changes requested):

neptune/alpha/internal/backends/hosted_neptune_backend.py
HubertJaworski: for safety I suppose, It was like this in old implementation

neptune/alpha/internal/backends/neptune_backend_mock.py
HubertJaworski: This is a mock/stub/offline version and notebooks are "optional" so they should not throw on non-hosted backend

neptune/alpha/internal/backends/hosted_neptune_backend.py
aniezurawski: Api response was not checked in old implementation. If I am not mistaked, a return value from helper method `create_checkpoint` in `notebooks` module is what was checked for None. Still don't know why. But this None check have to be moved at least.

aniezurawski (approved):

PR #448, opened 2021-03-11
Support exporting pandas to html

aniezurawski:


tests/neptune/alpha/internal/utils/test_images.py
```diff
@@ -183,6 +184,17 @@ def test_get_html_from_bokeh(self):
         # then
         self.assertTrue(result.startswith('\n\n\n\n<!DOCTYPE html>\n<html lang="en">'))
 
+    def test_get_html_from_pandas(self):
+        # given
+        table = pandas.read_csv('https://raw.githubusercontent.com/mwaskom/seaborn-data/master/iris.csv', nrows=100)
```
HubertJaworski, line 189: can we have an offline copy or any offline csv? making http requests in tests seems like the wrong way

HubertJaworski (approved):

PR #452, opened 2021-03-11
Print pretty warning in case of deleted experiments in sync command

aniezurawski:


neptune/new/sync.py
```diff
@@ -34,7 +34,7 @@
 from neptune.new.exceptions import (
     CannotSynchronizeOfflineExperimentsWithoutProject,
     NeptuneException,
-    ProjectNotFound,
+    ProjectNotFound, ExperimentNotFound,
```
shnela, line 37: Import should be in newline

shnela: You can check if such config will help:
Screenshot from 2021-03-12 09-53-22

shnela (changes requested):

shnela (approved):

PR #454, opened 2021-03-11
Add some more pretty errors

aniezurawski:


neptune/api_exceptions.py
```diff
@@ -33,32 +34,129 @@ def __init__(self):
...
+
+
+class WorkspaceNotFound(NeptuneApiException):
     def __init__(self, namespace_name):
-        super(NamespaceNotFound, self).__init__("Namespace '{}' not found.".format(namespace_name))
+        message = """
+{h1}
+----WorkspaceNotFound-------------------------------------------------------------------------
+{end}
+Workspace {python}{workspace}{end} not found.
```
Herudaio, line 144: I would add a line where to find the correct name or sth like that.

neptune/api_exceptions.py
```diff
@@ -33,32 +34,129 @@ def __init__(self):
...
 
 
 class ProjectNotFound(NeptuneApiException):
     def __init__(self, project_identifier):
-        super(ProjectNotFound, self).__init__("Project '{}' not found.".format(project_identifier))
+        message = """
+{h1}
+----ProjectNotFound-------------------------------------------------------------------------
+{end}
+Project {python}{project}{end} not found.
```
Herudaio, line 156: I would add a line where to find the correct name or sth like that.

Herudaio (changes requested):

Herudaio (approved):

PR #461, opened 2021-03-17
Rename download_zip to download

HubertJaworski:


neptune/new/experiments_table.py
```diff
@@ -92,10 +98,13 @@ def get(self):
         return self._experiment.get_attribute_value(self._path)
 
     def download(self, destination: Optional[str]):
-        return self._experiment.download_file_attribute(self._path, destination)
+        attr_type = self._experiment.get_attribute_type(self._path)
+        if attr_type == AttributeType.FILE:
+            return self._experiment.download_file_attribute(self._path, destination)
+        elif attr_type == AttributeType.FILE_SET:
+            return self._experiment.download_file_set_attribute(self._path, destination)
+        raise InternalClientError("Unsupported attribute type {}".format(attr_type))
```
aniezurawski, line 106: If this is internal method then you do not have to rename it. If it's not, if it can be used by user is it really internal error?

aniezurawski (changes requested):

neptune/new/experiments_table.py
HubertJaworski: this is not an internal method - I think I shoul've used `MetadataInconsistency` error instead

aniezurawski (approved):

PR #463, opened 2021-03-18
Implemnet fetch_values for non-file series

HubertJaworski:


neptune/new/handler.py
```diff
@@ -91,6 +91,18 @@ def upload_files(self, value: Union[str, Iterable[str]], wait: bool = False) ->
...
                 attr.upload_files(value, wait)
 
+    def fetch_values(self, include_timestamp=True):
+        with self._experiment.lock():
+            attr = self._experiment.get_attribute(self._path)
+            if attr:
+                if isinstance(attr, StringSeries):
+                    return attr.fetch_values(include_timestamp=include_timestamp)
+                elif isinstance(attr, FloatSeries):
+                    return attr.fetch_values(include_timestamp=include_timestamp)
```
shnela, line 101: This command and command two lines above looks exactly the same, all this instance check is redundant.

Additionally we could use common [abstract class](https://realpython.com/python-interface/) for both `StringSeries` and `FloatSeries` to keep analogical interface for all Series.

neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -49,7 +49,7 @@
     Project,
     StringAttribute,
     StringSeriesAttribute,
-    StringSetAttribute,
+    StringSetAttribute, StringSeriesValues, StringPointValue, FloatSeriesValues, FloatPointValue,
```
shnela, line 52: Keep every imported model in separate line, please

neptune/new/internal/backends/neptune_backend_mock.py
```diff
@@ -37,7 +37,7 @@
     Project,
     StringAttribute,
     StringSeriesAttribute,
-    StringSetAttribute,
+    StringSetAttribute, StringSeriesValues, FloatSeriesValues,
```
shnela, line 40: Keep every imported model in separate line, please

neptune/new/internal/backends/offline_neptune_backend.py
```diff
@@ -19,7 +19,8 @@
 
 from neptune.new.exceptions import NeptuneOfflineModeFetchException
 from neptune.new.internal.backends.api_model import Attribute, FloatAttribute, StringAttribute, \
```
shnela, line 21: For long import lists it's good to [wrap all imported modules in parentheses](https://www.python.org/dev/peps/pep-0328/#rationale-for-parentheses).

I also suggest to keep every item in separate line sorted alphabetically.

neptune/new/attributes/series/string_series.py
```diff
@@ -53,3 +53,29 @@ def get_last(self, wait=True) -> str:
             self._experiment.wait()
         val = self._backend.get_string_series_attribute(self._experiment_uuid, self._path)
         return val.last
+
+    def fetch_values(self, include_timestamp=True):
+        # pylint: disable=import-outside-toplevel
+        import pandas as pd
```
shnela, line 59: Is there any good reason for implementing this inside funciton?

If so, maybe we should mention it in comment.

neptune/new/attributes/series/float_series.py
```diff
@@ -69,3 +70,29 @@ def get_last(self, wait=True) -> float:
...
+
+    def fetch_values(self, include_timestamp=True):
+        # pylint: disable=import-outside-toplevel
+        import pandas as pd
+        limit = 1000
+        val = self._backend.get_float_series_values(self._experiment_uuid, self._path, 0, limit)
+        data = val.values
+        offset = limit
+
+        def make_row(entry: FloatPointValue) -> Dict[str, Optional[Union[str, float, datetime]]]:
```
shnela, line 82: You've implemented `make_row` twice.

The only difference is optional `entry` type: `FloatPointValue` vs `StringPointValue`, which we can somehow connect using subclassing.

shnela (changes requested):

neptune/new/attributes/series/string_series.py
HubertJaworski: In all of out project we do it like this - this kind of import means we do not initialize-on-import pandas lib (I think that was the reasoning behind it). At some point in time, `import neptune` took a long time because all the libs got initialized on import

shnela: Makes sense

neptune/new/attributes/series/fetchable_series.py
```diff
@@ -0,0 +1,55 @@
+#
+# Copyright (c) 2020, Neptune Labs Sp. z o.o.
```
shnela, line 2: I don't know if it matters, but maybe it should be 2021 :p

shnela (commented):
One small comment, I don't know if it require update.

neptune/new/attributes/series/fetchable_series.py
HubertJaworski: haha, you've caught me copy-pasting :)

shnela (approved):

PR #466, opened 2021-03-19
Add api to download image series

HubertJaworski:


neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -321,8 +323,17 @@ def to_attribute(attr) -> Attribute:
             raise ExperimentUUIDNotFound(exp_uuid=experiment_uuid)
 
     def download_file_series_by_index(self, experiment_uuid: uuid.UUID, path: List[str],
-                                      index: int, destination: Optional[str] = None):
-        pass
+                                      index: int, destination: str):
+        try:
+            download_image_series_element(
```
shnela, line 328: I don't get why we're calling `*image*` function in `*file*` function, what's the correlation?

shnela (approved):

neptune/new/internal/backends/hosted_neptune_backend.py
HubertJaworski: I'm not sure as well - the already existing FileSeries in client's code call image's api

PR #468, opened 2021-03-22
Fix tensorflow-keras integration version

HubertJaworski:


integrations/tensorflow-keras/neptune_tensorflow_keras/__init__.py
```diff
@@ -82,8 +75,9 @@ class NeptuneCallback(Callback):
         You need to have Keras or Tensorflow 2 installed on your computer to use this module.
     """
 
-    def __init__(self, experiment: Run, base_namespace: Optional[str] = None):
```
aniezurawski, line 85: experiment -> run

integrations/tensorflow-keras/neptune_tensorflow_keras/__init__.py
```diff
@@ -82,8 +75,9 @@ class NeptuneCallback(Callback):
...
+    def __init__(self, experiment: 'Run', base_namespace: Optional[str] = None):
         super().__init__()
+        from neptune.new import Run
```
aniezurawski, line 80: Why import here?

aniezurawski (changes requested):

kamil-kaczmarek:
@HubertJaworski I think that similar problem occurs in sklearn integration -> can you fix it in this PR as well?

integrations/tensorflow-keras/neptune_tensorflow_keras/__init__.py
HubertJaworski: We haven't release alpha client yet, so according to requirements, this package does not exist and it failds to build when imported

aniezurawski: It's going to be released today.

aniezurawski: Or tomorrow.

aniezurawski (approved):

PR #469, opened 2021-03-22
Raise pretty StorageLimitReach error

aniezurawski:
Additionally fix handling errors which should kill async thread
instead of just being logged.

### DIFF
--- a/neptune/new/exceptions.py
+++ b/neptune/new/exceptions.py
@@ -535,6 +535,18 @@ You may also want to check the following docs pages:
         super().__init__(message.format(**inputs))
 
 
+class StorageLimitReached(NeptuneException):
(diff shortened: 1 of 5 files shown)


neptune/new/exceptions.py
```diff
@@ -535,6 +535,18 @@ def __init__(self, framework):
...
 
 
+class StorageLimitReached(NeptuneException):
+    def __init__(self):
+        message = """
+{h1}     
+----StorageLimitReached-----------------------------------------
+{end}
+Storage limit reached.
+"""
```
aniezurawski, line 545: Mycek dostarczy później lepszy tekst.

HubertJaworski (approved):

PR #487, opened 2021-03-31
Switch to new api when project migration is finished.

aniezurawski:


neptune/internal/api_clients/hosted_api_clients/migration_switch_leaderboard_api_client_proxy.py
```diff
@@ -0,0 +1,52 @@
...
+
+        self._lock = threading.RLock()
+        self._switched = False
+
+    def __getattr__(self, item):
+        def func(*args, **kwargs):
+            try:
+                return getattr(self._client, item)(*args, **kwargs)
+            except ProjectMigratedToNewStructure:
+                if not self._switched:
```
HubertJaworski, line 45: I like this - this will allow us to change endpoints live witch running clients , but this self._switched bothers me - we can do this only once per client run cycle.

Can you remove it? I am aware that this will cause more GET-s for swagger

aniezurawski: No, we can not get rid of it. It is important for threads synchronization. All threads (std, metrics, channels, main thread and ping) use the same backend object and it is highly probable that they will get this exception at the same time.

HubertJaworski (dismissed):

shnela (dismissed):

neptune/internal/api_clients/hosted_api_clients/migration_switch_leaderboard_api_client_proxy.py
```diff
@@ -0,0 +1,52 @@
...
+                    self._lock.acquire()
+                    if not self._switched:
+                        self._client = self._backend_client.get_new_leaderboard_client()
+                        self._switched = True
+                    self._lock.release()
+                return getattr(self._client, item)
```
shnela, line 51: Don't we need `(*args, **kwargs)` here as well?
If so we can go without `def func(*args, **kwargs):`, because `getattr(self._client, item)` returns function.

aniezurawski: Yup. You're right.

neptune/internal/api_clients/hosted_api_clients/migration_switch_leaderboard_api_client_proxy.py
```diff
@@ -0,0 +1,240 @@
...
+
+if TYPE_CHECKING:
+    from neptune.internal.api_clients import HostedNeptuneBackendApiClient
+
+_logger = logging.getLogger(__name__)
+
+
+# pylint: disable=protected-access
+def with_migration_handling(func):
+    def wrapper(self: 'MigrationSwitchLeaderboardApiClientProxy', *args, **kwargs):
```
shnela, line 37: Are you sure that you want to call first argument `self`?
I suggest something like `migration_switch_api_client`.

shnela, line 37: And add [`wraps`](https://www.geeksforgeeks.org/python-functools-wraps-function/) just in case.

shnela (changes requested):

shnela (approved):

PR #532, opened 2021-04-30
Fix defining integer attribute

shnela:


neptune/new/run.py
```diff
@@ -230,5 +233,7 @@ def _define_attribute(self, _path: List[str], _type: AttributeType):
         if _type == AttributeType.RUN_STATE:
             self._structure.set(_path, RunStateAttr(self, _path))
 
+        raise NeptuneException(f"Unexpected type: {_type}")
```
shnela, line 236: Not 100% sure that we can raise exception here, but for sure we shouldn't let this function go through silently without taking any action.

aniezurawski: Exception is fine here. But we should fix hosted_neptune_backend:get_attributes:320 to ignore unknown types so adding them in backend won't break old client.

aniezurawski (changes requested):

neptune/new/attributes/utils.py
```diff
@@ -0,0 +1,22 @@
...
+    AttributeType.DATETIME: Datetime,
+    AttributeType.FILE: File,
+    AttributeType.FILE_SET: FileSet,
+    AttributeType.FLOAT_SERIES: FloatSeries,
+    AttributeType.STRING_SERIES: StringSeries,
+    AttributeType.IMAGE_SERIES: FileSeries,
+    AttributeType.STRING_SET: StringSet,
+    AttributeType.GIT_REF: GitRef,
+    AttributeType.RUN_STATE: RunState,
+    AttributeType.NOTEBOOK_REF: None,
```
shnela, line 21: What with NOTEBOOK_REF - it doesn't have an `Attribute` implemented.

aniezurawski (approved):

PR #534, opened 2021-05-04
Handle Stop/Abort signals when psutil is not installed

HubertJaworski:


neptune/new/internal/utils/process_killer.py
```diff
@@ -39,6 +42,9 @@ def kill_me():
         _, alive = psutil.wait_procs(children, timeout=KILL_TIMEOUT)
         for process in alive:
             _kill(process)
+    else:
+        click.echo(f"Run {run_id} received {'stop' if exit_code else 'abort'} signal. Exiting", err=True)
```
aniezurawski, line 46: What is this "exit_code"? Just print signal type. And do not do this in here. It is not process_killer's responsibility. It does not know anything about neptune domain.

aniezurawski (changes requested):

aniezurawski (approved):

PR #537, opened 2021-05-04
Raise exception when run from different process is used

aniezurawski:


aniezurawski:
@Raalsky Why have you closed it? Looks like a good change I forgot about.

Raalsky:
It's been stale for over a year. I've prepared a Jira issue if you still would like to reopen it etc.

PR #538, opened 2021-05-05
String set should return set not list

shnela:


neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -502,7 +502,7 @@ def get_string_set_attribute(self, run_uuid: uuid.UUID, path: List[str]) -> Stri
         }
         try:
             result = self.leaderboard_client.api.getStringSetAttribute(**params).response().result
-            return StringSetAttribute(result.values)
+            return StringSetAttribute(set(result.values))
```
shnela, line 505: We do not expect anything but simple `Atom` values here, right?

aniezurawski (approved):

PR #539, opened 2021-05-06
Fetch namespace as dict

shnela:


neptune/new/handler.py
```diff
@@ -44,10 +44,24 @@ def __getitem__(self, path: str) -> 'Handler':
...
+
     def __getattr__(self, attribute_name):
         attr = self._run.get_attribute(self._path)
         if attr:
-            return getattr(attr, attribute_name)
+            if not isinstance(attr, dict):
+                return getattr(attr, attribute_name)
+            else:
+                if attribute_name == 'fetch':
+                    return lambda: self._fetch_namespace_as_dict(attr)
```
shnela, line 61: Nasty hack.

I'd suggest to create `Namespace` class object with `fetch` and `assign` method.

Such object would be our client internal "fake attribute".
It'd be based on `dict` class and be used instead of regular dict in `RunStructure`.

This way we could free `Handler` code from logic regarding namespace fetch/assign management.

shnela (commented):
@Herudaio @kamil-kaczmarek  you probably would like to update `Handler.fetch` documentation.

pkasprzyk:
closing, since we merged https://github.com/neptune-ai/neptune-client/pull/549

PR #540, opened 2021-05-06
readonly mode for client

pkasprzyk:


neptune/new/exceptions.py
```diff
@@ -316,6 +316,11 @@ def __init__(self):
         super().__init__("Cannot synchronize offline runs without a project.")
 
 
+class NeedExistingRunForReadOnlyMode(NeptuneException):
+    def __init__(self):
+        super().__init__("An existing run must be provided when running in read-only mode.")
```
aniezurawski, line 321: Let's write all new exception messages in new pretty format. See `NeptuneRunResumeAndCustomIdCollision` for example.

aniezurawski (changes requested):

neptune/new/exceptions.py
pkasprzyk: updated, thanks

aniezurawski (changes requested):
Please add some simple test in test_client.py.

neptune/new/exceptions.py
```diff
@@ -316,6 +316,23 @@ def __init__(self):
...
+        message = """
+    {h1}
+    ----NeedExistingRunForReadOnlyMode-----------------------------------------
+    {end}
+    Incorrect call of function {python}neptune.init(){end}.
+
+    Parameter {python}run{end} of {python}neptune.init(){end} must be provided and reference
+    an existing run when using {python}mode="read-only"{end}.
+
+    {correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help
```
aniezurawski, line 330: @Herudaio any suggestions?

PR #541, opened 2021-05-07
Add fail_on_exception flag

HubertJaworski:


neptune/new/internal/init_impl.py
```diff
@@ -326,7 +329,8 @@ def init(project: Optional[str] = None,
             run_id=api_run.short_id
         ))
 
-    uncaught_exception_handler.activate()
+    if fail_on_exception:
```
aniezurawski, line 332: Is this `if` necessary? In future we can have many handlers registered.

neptune/new/internal/utils/traceback_job.py
```diff
@@ -31,18 +31,21 @@
...
+    def __init__(self, path: str, fail_on_exception: bool = True):
         self._uuid = uuid.uuid4()
         self._started = False
         self._path = path
+        self._fail_on_exception = fail_on_exception
 
     def start(self, run: 'Run'):
         if not self._started:
             path = self._path
+            fail_on_exception = self._fail_on_exception
```
aniezurawski, line 43: What is it for?

aniezurawski (changes requested):

neptune/new/internal/init_impl.py
HubertJaworski: No, I should remobe this

neptune/new/internal/utils/traceback_job.py
HubertJaworski: because you can't reference self easilsy from a `def` function passed as lambda

aniezurawski (approved):

PR #542, opened 2021-05-07
Integration with Optuna

PiotrJander:
Only merge and release once Optuna integration is published


neptune/new/integrations/optuna/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+try:
+    # pylint: disable=import-error
+    from neptune_optuna.impl import *
```
PiotrJander, line 19: I hope a * import is okay here

shnela: Why not to import one class and two functions explicitly?
It'll be cleaner to user, who would immediately see what's available from `neptune-optuna`.

By default `*` prevents you from importing `_*` names, but you'll be still able to import `from neptune.new.integrations.optuna import Union` - not very nice :p

shnela (changes requested):
https://github.com/neptune-ai/neptune-client/pull/542/files#r633447588

shnela (approved):

PR #548, opened 2021-05-11
PoC: Adds a PoC implementation of calculating and caching SHAs of files

PiotrJander:


pkasprzyk:
made redundant with current artifacts implementation

PR #551, opened 2021-05-12
custom handler for reporting to neptune

pkasprzyk:
ref. https://neptune-labs.atlassian.net/browse/NPT-10086


neptune/new/logging/handler.py
```diff
@@ -0,0 +1,62 @@
...
+import logging
+
+from neptune.new import Run
+from neptune.new.attributes.constants import MONITORING_DEFAULT_LOG_HANDLER_ATTRIBUTE_PATH
+from neptune.new.internal.utils import verify_type
+from neptune.new.logging import Logger
+
+
+class NeptuneLogHandler(logging.Handler):
+    """Custom handler that sends logged messages to Neptune
```
pkasprzyk, line 25: @Herudaio docstring for `NeptuneLogHandler`, comments appreciated :)

Herudaio, line 25: "Custom handler that sends logged messages to Neptune" 
-> 
"Handler that sends the log records created by the logger to Neptune"

neptune/new/logging/handler.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Args:
+        run (Run): An existing run reference (as returned by `neptune.init`)
+            Logger will send messages as a `StringSeries` attribute on this run.
```
Herudaio, line 29: attribute -> field

neptune/new/logging/handler.py
```diff
@@ -0,0 +1,62 @@
...
+        level (int, optional): Log level of the handler. Defaults to `logging.NOTSET`,
+            which logs everything that matches logger's level.
+        path (str, optional): Path to the `StringSeries` attribute used for logging. Default to `None`.
```
Herudaio, line 32: attribute -> field

neptune/new/logging/handler.py
```diff
@@ -0,0 +1,62 @@
...
+    Usage example:
+        >>> import logging
+        >>> import neptune.new as neptune
+        >>> from neptune.new.logging.handler import NeptuneLogHandler
+
+        >>> logger = logging.getLogger("root_experiment")
+        >>> logger.setLevel(logging.DEBUG)
+
+        >>> run = neptune.init(project=‘neptune/sandbox’)
+        >>> npt_handler = NeptuneLoggerHandler(run=run, level=logging.WARNING, path="logging/root_experiment")
```
Herudaio, line 44: In the basic example let's have a basic example and remove level and path parameters :)

neptune/new/logging/handler.py
```diff
@@ -0,0 +1,62 @@
...
+        >>> logger.addHandler(npt_handler)
+
+        >>> logger.debug("this won't be sent")
+        >>> logger.error("but this will!!%d", 1)
```
Herudaio, line 48: Let's provide a bit more practical example here hinting to the user as to why he may want to use this.
E.g. `logger.debug("Starting data preparation")`

neptune/new/logging/handler.py
```diff
@@ -0,0 +1,62 @@
...
+#
+import logging
+
+from neptune.new import Run
+from neptune.new.attributes.constants import MONITORING_DEFAULT_LOG_HANDLER_ATTRIBUTE_PATH
+from neptune.new.internal.utils import verify_type
+from neptune.new.logging import Logger
+
+
+class NeptuneLogHandler(logging.Handler):
```
Herudaio, line 24: The naming is incorrect with regard to the specification.
1. In general the naming from the specification should be a source of truth, and let's keep all the discussions about naming, public API etc. there. They are a form of "design spec" and should be treated similar :)
2. On the other way this prompted me to review the naming there it should be the following:
`from neptune.new.integrations.python_logger import NeptuneHandler`
a) what the thing is about we capture in module name (in this case *integration* with *python logger*)
b) we follow the naming of what we integrate with prefixed with Neptune. In this case all logger naming points to Handler

Herudaio: (I've update the spec accordingly)

neptune/new/logging/handler.py
shnela: @Herudaio 

Something like:
```

    You may also want to check `init docs page`_.

    .. _init docs page:
       https://docs.neptune.ai/api-reference/neptune#init
```
       
With link to new docs page will be required.

neptune/new/logging/handler.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Args:
+        run (Run): An existing run reference (as returned by `neptune.init`)
+            Logger will send messages as a `StringSeries` attribute on this run.
+        level (int, optional): Log level of the handler. Defaults to `logging.NOTSET`,
+            which logs everything that matches logger's level.
+        path (str, optional): Path to the `StringSeries` attribute used for logging. Default to `None`.
+            If `None`, `'monitoring/python_logger'` is used.
+
+    Usage example:
```
shnela: Replace `Usage example:` with `Examples:` - so it'll be rendered in proper way in pyCharm.

Herudaio (changes requested):

neptune/new/logging/handler.py
pkasprzyk: sorry, I'll voice my opinions in design spec next time

shnela (changes requested):

neptune/new/integrations/python_logger.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Examples:
+        >>> import logging
+        >>> import neptune.new as neptune
+        >>> from neptune.new.integrations.python_logger import NeptuneHandler
+
+        >>> logger = logging.getLogger("root_experiment")
+        >>> logger.setLevel(logging.DEBUG)
+
+        >>> run = neptune.init(project=‘neptune/sandbox’)
```
shnela, line 43: \` -> ' or "

I know it doesn't work anyway because of missing public credentials to pass to `neptune.init`, but maybe some time we'll do something with it.

shnela (dismissed):
One small new comment, not so important to update it.

If https://github.com/neptune-ai/neptune-client/pull/551#discussion_r633413626 can wait +1 and let's merge

neptune/new/integrations/python_logger.py
pkasprzyk: great catch, thanks!

Herudaio (approved):

PR #557, opened 2021-05-17
warning on out-of-range image pixel values

pkasprzyk:
ref. https://neptune-labs.atlassian.net/browse/NPT-9970

### DIFF
--- a/tests/neptune/new/internal/utils/test_images.py
+++ b/tests/neptune/new/internal/utils/test_images.py
@@ -15,6 +15,7 @@
 #
 
 # pylint: disable=protected-access
+import contextlib
 import io
 import os
 import sys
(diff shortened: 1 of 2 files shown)


shnela (dismissed):

tests/neptune/new/internal/utils/test_images.py
```diff
@@ -68,19 +69,26 @@ def test_get_image_content_from_2d_grayscale_array(self):
...
 
         # expect
-        self.assertEqual(get_image_content(image_array), self._encode_pil_image(expected_image))
+        stderr = io.StringIO()
+        with contextlib.redirect_stderr(stderr):
+            self.assertEqual(get_image_content(image_array), self._encode_pil_image(expected_image))
+        self.assertEqual(
+            stderr.getvalue(),
+            "Image values should be in [0, 1] range, but smallest array value is -3 and greatest array value is 6."
+            " Colour values may be interpreted incorrectly.\n"
```
pkasprzyk, line 90: @Herudaio is that warning message OK for https://neptune-labs.atlassian.net/browse/NPT-9970 ?

Only relevant part is shown, so if each value is >=0 but some exceeds 1, it's going to be:
> Image values should be in [0, 1] range, but greatest array value is 6. Colour values may be interpreted incorrectly.

Consequently, if each value is <=1, but some is <0, you get: 
> Image values should be in [0, 1] range, but smallest array value is -3. Colour values may be interpreted incorrectly.

Herudaio: Colour -> color as we are using American English and I would re-organize the info a bit. So that what happened is together and expectations are together:

"Image values should be in [0, 1] range, but greatest array value is 6. Colour values may be interpreted incorrectly."
->
"The largest value in the array is 6. To be interpreted as colors correctly values in the array need to be in the [0, 1] range."

pkasprzyk: updated, thanks:

> The largest value in the array is 6. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.

> The smallest value in the array is -3. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.

> The smallest value in the array is -3 and the largest value in the array is 6. To be interpreted as colors correctly values in the array need to be in the [0, 1] range.

shnela (approved):

PR #561, opened 2021-05-19
update limits and limit warning messages

pkasprzyk:
also refactored a little bit to collect all limits in single file

ref. https://neptune-labs.atlassian.net/browse/NPT-10081

### DIFF
--- a/neptune/new/internal/utils/limits.py
+++ b/neptune/new/internal/utils/limits.py
@@ -1,0 +1,56 @@
+#
+# Copyright (c) 2021, Neptune Labs Sp. z o.o.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


neptune/new/internal/utils/limits.py
```diff
@@ -0,0 +1,59 @@
...
+                        content_size / BYTES_IN_MB,
+                        _IMAGE_SIZE_LIMIT_MB)
+        return True
+    return False
+
+
+def file_size_exceeds_limit(content_size):
+    if content_size > _IN_MEMORY_SIZE_LIMIT_MB * BYTES_IN_MB:
+        _logger.warning('You are attempting to create an in-memory file that is %.1fMB large. '
+                        'Neptune supports logging files in-memory objects smaller than %dMB. '
```
aniezurawski, line 44: "files in-memory objects"? Sounds weird.
There is "in-memory file objects" in spec.

aniezurawski (approved):

PR #584, opened 2021-05-25
Fix websocket-client version

aniezurawski:


HubertJaworski (approved):
Test version 1.0.1 before merging

PR #602, opened 2021-06-21
Jk/npt 9386 prevent logging after stop

shnela:


neptune/new/run.py
```diff
@@ -57,7 +69,7 @@ class Run(AbstractContextManager):
 
     You can log many ML metadata types, including:
         * metrics
-        * losses
+        * losses]
```
pkasprzyk, line 72: :-1:

neptune/new/run.py
```diff
@@ -111,6 +123,7 @@ def __init__(
         self._structure: RunStructure[Attribute, NamespaceAttr] = RunStructure(NamespaceBuilder(self))
         self._lock = threading.RLock()
         self._started = False
+        self._stopped = False
```
pkasprzyk, line 126: I would very much prefer not to have separate `_started` and `_stopped`.

Maybe `self._state` with enum value (`CREATED`, `STARTED`, `STOPPING`, `STOPPED`)?

neptune/new/exceptions.py
```diff
@@ -155,6 +155,29 @@ def __init__(self, run_uuid: uuid.UUID):
         super().__init__("Run with UUID {} not found. Could be deleted.".format(run_uuid))
 
 
+class InactiveRunException(NeptuneException):
+    def __init__(self, short_id: str):
+        message = """
+{h1}
+----InactiveRunException----------------------------------------
```
pkasprzyk, line 162: @Herudaio could you confirm you like this help screen? :)

pkasprzyk (changes requested):

neptune/new/run.py
shnela: Can do

neptune/new/exceptions.py
shnela: Yes it's his version

pkasprzyk (approved):

PR #607, opened 2021-06-29
heuristics to help users who mixup legacy/new client API

pkasprzyk:
ref. https://neptune-labs.atlassian.net/browse/NPT-10404

### DIFF
--- a/neptune/__init__.py
+++ b/neptune/__init__.py
@@ -22,6 +22,7 @@ from neptune._version import get_versions
 from neptune.exceptions import (
     InvalidNeptuneBackend,
     NeptuneUninitializedException,
+    NeptuneIncorrectImportException,
 )
 from neptune.internal.api_clients import backend_factory
 from neptune.internal.api_clients.offline_backend import OfflineBackend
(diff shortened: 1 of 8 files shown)


neptune/__init__.py
```diff
@@ -52,7 +53,19 @@
 ANONYMOUS_API_TOKEN = constants.ANONYMOUS_API_TOKEN
 
 
-def init(project_qualified_name=None, api_token=None, proxies=None, backend=None):
+CURRENT_KWARGS = ('project', 'run', 'custom_run_id', 'mode')
```
shnela, line 56: Shouldn't we have more of new-style `init` arguments?
* `custom_run_id`
* `mode`
* `name`
* `description`
* `tags`
* and so on...

neptune/new/run.py
```diff
@@ -69,6 +71,19 @@ def inner_fun(self, *args, **kwargs):
     return inner_fun
 
 
+LEGACY_METHODS = (
+    'create_experiment',
```
shnela, line 75: `create_experiment`  isn't part of `Run` class.

It should be defined in `neptune.new` similar to other old `neptune` methods like:
* `set_project`
* `get_experiment`
* `append_tag`
* etc.

shnela (changes requested):
I'd also add some tests.

neptune/__init__.py
pkasprzyk: I was aiming for a "heuristic" approach: detect args that you can realistically expect to be used each time.

But maybe I'm too cunning for my own good. I'll add more args.

neptune/new/run.py
pkasprzyk: 1. you're right, I'll add old `neptune` methods to `neptune.new`
2. I'd argue that this one actually makes sense, too. I think this is a legit flow for old API:
```
project = neptune.init(...)
project.create_experiment(...)
```

shnela: Ad 2.
Yes

shnela (dismissed):

shnela (approved):

PR #608, opened 2021-06-29
cache pip in CI

pkasprzyk:


pkasprzyk:
doesn't seem to be worth it, for now

PR #609, opened 2021-06-30
return path from requested prefix instead of root when fetching namespace

pkasprzyk:
ref. https://neptune-labs.atlassian.net/browse/NPT-10417

before:
```
import neptune.new as neptunerun = neptune.init()
run['model/params'] = {'a': 3}
run.wait()
params run['model/params'].fetch() == {'model': {'params': {'a': 3}}}
```

after:
```
import neptune.new as neptunerun = neptune.init()
run['model/params'] = {'a': 3}
run.wait()
params run['model/params'].fetch() == {'a': 3}
```

Also refactored `fetch_atom_attribute_values` to make it more unit-testable.

### DIFF
--- a/neptune/new/attributes/namespace.py
+++ b/neptune/new/attributes/namespace.py
@@ -17,9 +17,10 @@ from collections.abc import MutableMapping
 from typing import Any, Dict, TYPE_CHECKING, Iterator, List, Mapping, Union
 
 from neptune.new.attributes.attribute import Attribute
+from neptune.new.internal.run_structure import RunStructure
 from neptune.new.internal.utils.generic_attribute_mapper import atomic_attribute_types_map, NoValue
+from neptune.new.internal.utils.paths import parse_path, path_to_str
 from neptune.new.types.namespace import Namespace as NamespaceVal
-from neptune.new.internal.utils.paths import path_to_str
 
 if TYPE_CHECKING:
     from neptune.new.run import Run
@@ -75,8 +76,6 @@ class Namespace(Attribute, MutableMapping):
 
     def fetch(self) -> dict:
         attributes = self._backend.fetch_atom_attribute_values(self._run_uuid, self._path)
-        from neptune.new.internal.run_structure import RunStructure
-        from neptune.new.internal.utils.paths import parse_path
         run_struct = RunStructure()
         prefix_len = len(self._path)
         for attr_name, attr_type, attr_value in attributes:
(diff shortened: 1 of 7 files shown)


neptune/new/attributes/namespace.py
```diff
@@ -75,8 +74,14 @@ def _collect_atom_values(self, attribute_dict) -> dict:
...
     def fetch(self) -> dict:
-        namespace_values = self._backend.fetch_atom_attribute_values(self._run_uuid, self._path)
-        return self._collect_atom_values(namespace_values)
+        attributes = self._backend.fetch_atom_attribute_values(self._run_uuid, self._path)
+        from neptune.new.internal.run_structure import RunStructure
+        from neptune.new.internal.utils.paths import parse_path
+        run_struct = RunStructure()
+        prefix_len = len(self._path)
+        for attr_name, attr_type, attr_value in attributes:
+            run_struct.set(parse_path(attr_name)[prefix_len:], (attr_type, attr_value))
```
Raalsky, line 83: I think this "deserialization" logic should be placed in the RunStructure class.

neptune/new/attributes/namespace.py
```diff
@@ -75,8 +74,14 @@ def _collect_atom_values(self, attribute_dict) -> dict:
         return result
 
     def fetch(self) -> dict:
-        namespace_values = self._backend.fetch_atom_attribute_values(self._run_uuid, self._path)
-        return self._collect_atom_values(namespace_values)
+        attributes = self._backend.fetch_atom_attribute_values(self._run_uuid, self._path)
+        from neptune.new.internal.run_structure import RunStructure
```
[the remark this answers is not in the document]
pkasprzyk: sorry, I did it while testing and forgot to clean it up :(

neptune/new/attributes/namespace.py
pkasprzyk: do you mean `parse_path(attr_name)[prefix_len:]`?

the thing is I'm abusing `RunStructure` a little bit here (it's normally used for run **structure**, not values) so probably the by-the-book thing to do would to subclass `RunStructure` as `RunWithValuesStructure` and add deserialization there... not sure if it's worth it, though?

Raalsky: So maybe in some utils file/internal method in Namespace?

Raalsky (approved):

PR #611, opened 2021-07-01
exception for new integrations using old exps

pkasprzyk:
ref. https://neptune-labs.atlassian.net/browse/NPT-10408

### DIFF
--- a/neptune/new/exceptions.py
+++ b/neptune/new/exceptions.py
@@ -748,3 +748,23 @@ You may also want to check the following docs pages:
 """
         inputs = dict(list({}.items()) + list(STYLES.items()))
         super().__init__(message.format(**inputs))
+
+
+class NeptuneLegacyIncompatibilityException(NeptuneException):
+    def __init__(self):
+        message = """
+{h1}
+----NeptuneLegacyIncompatibilityException----------------------------------------
+{end}
+It seems you are passing the legacy Experiment object, when a Run object is expected.
+
+What can I do?
+    - Updating your code to the new Python API requires few changes, but to help you with this process we prepared a handy migration guide:
+    https://docs.neptune.ai/migration-guide
+    - You can read more about neptune.new in the release blog post:
+    https://neptune.ai/blog/neptune-new
+
+{correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help
+"""
+        inputs = dict(list({}.items()) + list(STYLES.items()))
+        super().__init__(message.format(**inputs))
--- a/neptune/new/internal/utils/__init__.py
+++ b/neptune/new/internal/utils/__init__.py
@@ -18,9 +18,13 @@ from io import IOBase, BytesIO
 import logging
 import os
 from glob import glob
-from typing import Union, TypeVar, Iterable, List, Set, Optional, Mapping
+from typing import Union, TypeVar, Iterable, List, Set, Optional, Mapping, TYPE_CHECKING
 
 from neptune.new.internal.utils import limits
+
+if TYPE_CHECKING:
+    from neptune.new import Run
+
 
 T = TypeVar('T')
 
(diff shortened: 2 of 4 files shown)


neptune/new/internal/utils/__init__.py
```diff
@@ -170,3 +174,13 @@ def is_ipython() -> bool:
...
         return False
+
+
+def expect_not_an_experiment(run: 'Run'):
+    from neptune.new.exceptions import NeptuneLegacyIncompatibilityException
+    try:
+        from neptune.experiments import Experiment
+        if isinstance(run, Experiment):
+            raise NeptuneLegacyIncompatibilityException()
+    except ImportError:
```
shnela, line 185: Why do expect `ImportError`? It's part of the same repository.

shnela (approved):

neptune/new/internal/utils/__init__.py
pkasprzyk: I'm trying to think about it as a legacy API, and protect against dropping it someday but leaving in this import? Dunno, really...

neptune/new/exceptions.py
```diff
@@ -704,3 +704,23 @@ def __init__(self):
...
+What can I do?
+    - Updating your code to the new Python API requires few changes, but to help you with this process we prepared a handy migration guide:
+    https://docs.neptune.ai/migration-guide
+    - You can read more about neptune.new in the release blog post:
+    https://neptune.ai/blog/neptune-new
+
+{correct}Need help?{end}-> https://docs.neptune.ai/getting-started/getting-help
+"""
+        inputs = dict(list({}.items()) + list(STYLES.items()))
+        super().__init__(message.format(**inputs))
```
Raalsky, line 726: Maybe just `super().__init__(message.format(**STYLES))`?

neptune/new/internal/utils/__init__.py
```diff
@@ -170,3 +174,13 @@ def is_ipython() -> bool:
         return ipython is not None
     except ImportError:
         return False
+
+
+def expect_not_an_experiment(run: 'Run'):
+    from neptune.new.exceptions import NeptuneLegacyIncompatibilityException
```
Raalsky, line 180: Why not at the beginning of the file?

neptune/new/internal/utils/__init__.py
Raalsky: Nah, :+1: to @shnela . In case of dropping the compatibility, we should be still aware of every "legacy" part in the codebase. It will be omitted someday instead of just trash it.

neptune/new/internal/utils/__init__.py
pkasprzyk: not proud of it, but this deals with circular deps (`exceptions.py` uses `utils/__init__.py`)

neptune/new/exceptions.py
pkasprzyk: I think it's some kind of convention we have -- it's the same in all other exceptions, so I wanted to stay consistent

neptune/new/internal/utils/__init__.py
Raalsky: Ok

neptune/new/exceptions.py
Raalsky: I've already break it into other PR :wink:

neptune/new/internal/utils/__init__.py
pkasprzyk: ok, removed try-except

I also had to do local import due to circular dep (see https://github.com/neptune-ai/neptune-client/pull/611#discussion_r664337859). Maybe you'd prefer me to move this function to a new module (eg. `utils/compatibility.py`) ?

Raalsky: Sure

pkasprzyk: fixed, could you also reapprove https://github.com/neptune-ai/neptune-tensorflow-keras/pull/5 then? :)

neptune/new/exceptions.py
pkasprzyk: so un-break it, consistency is good :P

if you think this is a stupid convention (I won't really argue), we can change it in all instances, in a separate PR

Raalsky (approved):

shnela (approved):

PR #637, opened 2021-08-03
upgrade pylint to current

pkasprzyk:


neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -457,8 +457,8 @@ def to_attribute(attr) -> Attribute:
                 set(attr.type for attr in run.attributes) - set(attr.type for attr in accepted_attributes)
             if ignored_attributes:
                 _logger.warning(
-                    f"Ignored following attributes (unknown type): {ignored_attributes}.\n"
-                    f"Try to upgrade `neptune-client."
+                    "Ignored following attributes (unknown type): %s.\n"
```
shnela, line 460: Why?
I know that pylint complains about that, but I've deliberately disabled this kind of check (f-string in warnings).

shnela (approved):

neptune/new/internal/backends/hosted_neptune_backend.py
pkasprzyk: discussed internally, tldr: `%` formatting allows for log-message-template grouping (e.g. when reporting to sentry), so I think it's a good pattern

PR #638, opened 2021-08-03
ArtifactDriversMap and base ArtifactDriver

pkasprzyk:
ref. https://neptune-labs.atlassian.net/browse/NPT-10596


tests/neptune/internal/artifacts/test_types.py
```diff
@@ -0,0 +1,77 @@
...
+        driver_instance = ArtifactDriversMap.match_type("test")
+
+        self.assertEqual(driver_instance, self.test_driver_instance)
+
+    def test_unmatched_type_raises_exception(self):
+        with self.assertRaises(NeptuneUnhandledArtifactTypeException):
+            ArtifactDriversMap.match_type("test2")
+
+
+if __name__ == '__main__':
```
Raalsky, line 76: The main runner is not necessary to be included in unittest.

neptune/internal/artifacts/__init__.py
```diff
@@ -0,0 +1,15 @@
+#
```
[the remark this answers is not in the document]
Raalsky: Or even `neptune/new/artifacts`

neptune/internal/artifacts/types.py
```diff
@@ -0,0 +1,78 @@
+#
```
Raalsky, line 1: Same as preceding. Maybe package under `neptune/new/artifacts/` will be more suitable? or including `internal`

neptune/internal/artifacts/types.py
```diff
@@ -0,0 +1,78 @@
...
+    type: str
+    metadata: typing.Dict[str, str]
+
+
+class ArtifactDriversMap:
+    _implementations: typing.List['ArtifactDriver'] = []
+
+    @classmethod
+    def match_path(cls, path: str) -> 'ArtifactDriver':
+        for artifact_driver in cls._implementations:
```
Raalsky, line 44: Instead of storing implementations list consider the profs/cons of `ArtifactDriver.__subclasses__()`

neptune/internal/artifacts/__init__.py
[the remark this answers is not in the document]
pkasprzyk: ... sorry

tests/neptune/internal/artifacts/test_types.py
Raalsky: Sorry, dup after browser restart. Closing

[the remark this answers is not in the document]
pkasprzyk: c&p from other tests... not sure who thought of this and why, but you're right -- no use doing that

neptune/internal/artifacts/types.py
pkasprzyk: 1. ArtifactDriver.__subclasses__() only lists immediate subclasses (you'd have to do recursion or remember to NEVER do sub-specialization)
2. I like to store/use instances. This way typing is nicer (I don't have to annotate `typing.Type['ArtifactDriver']` everywhere), and it feels more natural to NOT use static-/classmethods everywhere ;)

Raalsky (approved):

PR #645, opened 2021-08-06
Classes instead of instances in implementations list

Raalsky:


neptune/new/internal/artifacts/types.py
```diff
@@ -32,23 +32,23 @@ class ArtifactFileType(enum.Enum):
 class ArtifactFileData:
     file_path: str
     file_hash: str
-    type: str
+    type: ArtifactFileType
```
pkasprzyk, line 35: I'd prefer this to stay `str` -- otherwise user adding custom filetype drivers would either have to monkey-patch `ArtifactFileType` or break type annotation

pkasprzyk (changes requested):
see comment

neptune/new/internal/artifacts/types.py
Raalsky: Sure. I've changed it already. The cons to `str` is how the enum works. It returns `ArtifactFileType.S3` as a type instead of `str` which is accessible via `.value` so in most of the places we should remember to call `ArtifactFileType.S3.value` which might be confusing.

pkasprzyk (approved):

PR #659, opened 2021-08-16
Fix get_tracked_files

shnela:


Raalsky (approved):

neptune/new/internal/artifacts/drivers/local.py
```diff
@@ -51,7 +51,8 @@ def _deserialize_metadata(cls, metadata: typing.Dict[str, str]) -> typing.Dict[s
 
     @classmethod
     def get_tracked_files(cls, path: str, namespace: str = None) -> typing.List[ArtifactFileData]:
-        parsed_path = urlparse(path).path
+        path_parts = urlparse(path)
+        parsed_path = f'{path_parts.netloc}{path_parts.path}'
```
pkasprzyk, line 55: this (and `matches()` implementation, for that matter) will not work correctly with windows path without `file` scheme:
```python
>>> urlparse(r'c:\windows')
ParseResult(scheme='c', netloc='', path='\\windows', params='', query='', fragment='')
```

I think that's acceptable for now, but @shnela can you add a docstring in `LocalArtifactDriver` saying that for Windows you have to use the File URI scheme?

pkasprzyk (approved):

PR #670, opened 2021-08-24
add Artifact.fetch() and make Artifact.assign() public

pkasprzyk:


neptune/new/attributes/atoms/artifact.py
```diff
@@ -24,14 +24,17 @@
...
+        # this function should be used only with ArtifactVal
         if not isinstance(value, ArtifactVal):
-            value = ArtifactVal(value)
+            raise TypeError("Value of unsupported type {}".format(type(value)))
 
         with self._run.lock():
             self._enqueue_operation(AssignArtifact(self._path, value.hash), wait)
 
+    def fetch(self) -> ArtifactVal:
+        return ArtifactVal(self.fetch_hash())
```
Raalsky, line 36: Add `hash=self.fetch_hash()`

pkasprzyk: fun fact: because(?) Artifact extends Atom, its signature is `__init__(self, value: Optional[str] = None)`

I didn't want to expose that, honestly -- and on the other hand I don't want to break the Unified Signature (if there is such a thing)

tests/neptune/new/attributes/atoms/test_artifact_hash.py
```diff
@@ -31,6 +31,17 @@ def test_assign_type_error(self):
...
 
+    def test_fetch(self):
+        exp, path = self._create_run(), self._random_path()
+        var = Artifact(exp, path)
+        var._enqueue_operation(
+            AssignArtifact(
+                var._path,
+                ArtifactVal("e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855").hash
+            ),
+            False)
```
Raalsky, line 42: I didn't check it but if this argument is `wait` maybe it should be set to `True`?

pkasprzyk: it's setup using mock backend, so meh

neptune/new/attributes/atoms/artifact.py
Raalsky: ohh ok

Raalsky (approved):

PR #674, opened 2021-08-26
Fixes after E2E tests

Raalsky:


neptune/new/internal/artifacts/types.py
```diff
@@ -43,9 +56,9 @@ def from_dto(cls, artifact_file_dto):
             file_hash=artifact_file_dto.fileHash,
             type=artifact_file_dto.type,
             size=artifact_file_dto.size,
-            metadata=ArtifactMetadataSerializer.deserialize([
-                (m.key, m.value) for m in artifact_file_dto.metadata
-            ])
+            metadata=ArtifactMetadataSerializer.deserialize(
+                list(map(lambda m: {'key': str(m.key), 'value': str(m.value)}, artifact_file_dto.metadata))
```
shnela, line 60: List comprehension is more readable than `map` ;)

neptune/new/internal/backends/utils.py
```diff
@@ -130,6 +130,9 @@ def create_swagger_client(url: str, http_client: HttpClient) -> SwaggerClient:
 
 
 def verify_client_version(client_config: ClientConfig, version: Version):
+    if str(version) == '0+unknown':
```
shnela, line 133: Add some comment when version is `'0+unknown'` and that it should be the case only for e2e tests.

You've updated e2e tests to test recent master instead of recent version published on PIP, right?

Raalsky: Fixed in https://github.com/neptune-ai/neptune-client/pull/678

shnela (approved):

PR #676, opened 2021-08-26
Added capture_traceback in neptune.init

Raalsky:


neptune/new/internal/init_impl.py
```diff
@@ -110,6 +110,7 @@ def init(project: Optional[str] = None,
          monitoring_namespace: Optional[str] = None,
          flush_period: float = 5,
          proxies: Optional[dict] = None,
+         capture_traceback: bool = True,
```
aniezurawski, line 113: @Herudaio is it ok to add this parameter?

Herudaio: I'm always against by default adding new init params :D However, if it's needed this one makes sense. I would put them right after `capture_stderr` so that they are at least grouped.

aniezurawski: But it would break backward compatibility if someone is using positional arguments instead of named.

Herudaio: Right, I withdraw that comment :)

aniezurawski (approved):

PR #682, opened 2021-09-02
Do not fetch artifact hash synchronously

shnela:


neptune/new/attributes/atoms/artifact.py
```diff
@@ -53,7 +53,7 @@ def download(self, destination: str = None):
             file_destination.parent.mkdir(parents=True, exist_ok=True)
             driver.download_file(file_destination, file_definition)
 
-    def track_files_to_new(
+    def track_files(
```
aniezurawski, line 56: Why the siganture is different than `Handler::track_files`? Should be the same. And you haven't even changed implementation of handler. It tries to call inexisting methods. Do we have any tests for `Handler::track_files`?

neptune/new/internal/backends/operations_preprocessor.py
```diff
@@ -243,39 +242,21 @@ def visit_delete_attribute(self, op: DeleteAttribute) -> None:
                 # simply perform single delete operation.
                 self._delete_ops.append(op)
 
-    def _artifact_log_modifier(self, ops: List[Operation], new_op: Operation) -> List[Operation]:
+    def _artifact_log_modifier(self, ops: List[TrackFilesToArtifact],
```
aniezurawski, line 245: May be static.

aniezurawski (changes requested):

neptune/new/internal/backends/operations_preprocessor.py
shnela: Let it be.

neptune/new/attributes/atoms/artifact.py
shnela: > Why the siganture is different than Handler::track_files? Should be the same.
IDK, it was implemented that way, I'll take closer look on that.

> And you haven't even changed implementation of handler. It tries to call inexisting methods.
Yes, my bad...

> Do we have any tests for Handler::track_files?
I guess not, since test passed. I'll take care of it.

shnela: > Why the siganture is different than Handler::track_files? Should be the same.

It's not since we're getting `project_uuid` from `run` instance.

aniezurawski: Should be. Like it's said in our jira task - methods in Handler should be subset of methods from all types of Attributes.
Can we pass project_uuid to `Artifact.__init__`?

neptune/new/handler.py
```diff
@@ -447,14 +447,14 @@ def track_files(self, path: str, destination: str = None, wait: bool = True) ->
...
-                attr.track_files_to_new(
+                attr.track_files(
                     project_uuid=self._run._project_uuid,  # pylint: disable=protected-access
                     source_location=path,
                     destination=destination,
                     wait=wait
                 )
             else:
-                attr.track_files_to_existing(
+                attr.track_files(
```
aniezurawski, line 457: This is exactly same call as above.

aniezurawski: And it should be called after `self._run.set_attribute`. Please merge current master to this branch and rebase this PR.

neptune/new/attributes/atoms/artifact.py
shnela: > Can we pass project_uuid to Artifact.__init__?

It's not the best idea, but you were right, both `track_files` can have the same signature.

We do not have pass `project_uuid` at all, since `Artifact (attribute)` contains `run` object with that information.

Done.

neptune/new/attributes/atoms/artifact.py
```diff
@@ -55,13 +54,14 @@ def download(self, destination: str = None):
 
     def track_files(
             self,
-            project_uuid: uuid.UUID,
             source_location: str,
```
aniezurawski, line 57: This parameter name is different then the one in Handler.

aniezurawski (approved):

PR #694, opened 2021-09-07
Added better exceptions handling for lack of files and wildcards

Raalsky:


neptune/new/exceptions.py
```diff
@@ -885,3 +885,29 @@ def __init__(self, location: str):
...
         super().__init__("Cannot upload artifact: {}".format(msg))
+
+
+class NeptuneUnsupportedArtifactFunctionalityException(NeptuneException):
+    def __init__(self, functionality_info: str):
+        message = """
+{h1}
+----NeptuneUnsupportedArtifactFunctionality-------------------------------------
+{end}
+It seems you are using Neptune Artifacts functionality "{functionality_info}" that is currently not supported.
```
aniezurawski, line 896: Please move "functionality_info" to new line so in result we have something like this:

```
It seems you are using Neptune Artifacts functionality that is currently not supported.

Wildcard characters (*,?) in location URI are not supported.
```

neptune/new/exceptions.py
```diff
@@ -885,3 +885,29 @@ def __init__(self, location: str):
...
+        super().__init__(message.format(functionality_info=functionality_info, **STYLES))
+
+
+class NeptuneEmptyLocationException(NeptuneException):
+    def __init__(self, location: str):
+        message = """
+{h1}
+----NeptuneEmptyLocationException----------------------------------------------
+{end}
+Neptune could not find files in the requested location ({location}).
```
aniezurawski, line 909: This exception will be raised asynchronously so without any context user can have some problem to guess what is it about. Let's be more verbose and friendly.

aniezurawski (changes requested):

aniezurawski (approved):

PR #697, opened 2021-09-08
Project name resolver

Raalsky:


neptune/management/exceptions.py
```diff
@@ -26,3 +26,13 @@ def details(self):
             "code": self.code,
             "description": self.details,
         }
+
+
+class MissingWorkspaceName(ManagementOperationFailure):
+    code = -2
```
pkasprzyk, line 32: I get the idea, but I think positive error code would be more natural

neptune/management/exceptions.py
```diff
@@ -26,3 +26,13 @@ def details(self):
...
+    description = "Missing workspace name"
```
pkasprzyk, line 33: let's try a little more verbose, please ;) maybe something like:
`Cannot resolve project {name}, you have to provide a workspace name.`

neptune/management/internal/utils.py
```diff
@@ -13,3 +13,21 @@
...
 #
+import re
+from typing import Optional
+
+from neptune.management.exceptions import MissingWorkspaceName, ConflictingWorkspaceName
+from neptune.patterns import PROJECT_QUALIFIED_NAME_PATTERN
+
+
+def normalize_project_name(name: str, workspace: Optional[str] = None):
+    project_spec = re.search(PROJECT_QUALIFIED_NAME_PATTERN, name)
```
pkasprzyk, line 24: this can return `None` if `name` doesn't match `PROJECT_QUALIFIED_NAME_PATTERN` at all (e.g. `"this/project/is/wrong"`)

please handle that graciously (`ManagementOperationFailure`) and add a unit test :)

pkasprzyk (changes requested):

pkasprzyk (approved):

PR #699, opened 2021-09-09
Expand ~ when tracking artifacts

shnela:


tests/neptune/new/internal/artifacts/drivers/test_local.py
```diff
@@ -192,6 +193,19 @@ def test_multiple_retrieval_prefix(self):
         self.assertEqual(46, files[3].size)
         self.assertTrue(files[3].metadata['file_path'].endswith(metadata_path_suffix))
 
+    def test_expand_user(self):
+        with open(pathlib.Path('~/tmp_test_expand_user').expanduser(), 'w') as f:
```
pkasprzyk, line 197: You're writing to a file in home dir, that's kinda worrying.

Could you use monkeypatch.setitem to replace `os.environ['HOME']` to `tempfile.gettempdir()` for this test? :)

pkasprzyk (approved):

PR #709, opened 2021-09-16
gracefully handle artifacts' API unavailability

pkasprzyk:
ref. https://neptune-labs.atlassian.net/browse/NPT-10843

### DIFF
--- a/neptune/new/internal/backends/hosted_neptune_backend.py
+++ b/neptune/new/internal/backends/hosted_neptune_backend.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-import json
 import logging
 import os
 import platform
--- a/neptune/new/internal/backends/hosted_neptune_backend.py
+++ b/neptune/new/internal/backends/hosted_neptune_backend.py
@@ -44,7 +44,8 @@ from neptune.new.exceptions import (
     NeptuneStorageLimitException,
     UnsupportedClientVersion,
     ArtifactNotFoundException,
-    NeptuneFeaturesNotAvailableWarning, NeptuneFeaturesNotAvailableException,
+    NeptuneFeaturesNotAvailableWarning,
+    NeptuneFeaturesNotAvailableException,
 )
 from neptune.new.internal.backends.api_model import (
     ApiRun,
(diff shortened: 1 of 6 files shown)


HubertJaworski (approved):

neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -165,10 +170,14 @@ def __init__(self, credentials: Credentials, proxies: Optional[Dict[str, str]] =
...
-        self.artifacts_client = create_swagger_client(
-            build_operation_url(self._client_config.api_url, self.ARTIFACTS_SWAGGER_PATH),
-            self._http_client
-        )
+        try:
+            self.artifacts_client = create_swagger_client(
+                build_operation_url(self._client_config.api_url, self.ARTIFACTS_SWAGGER_PATH),
+                self._http_client
+            )
+        except Exception:
```
aniezurawski, line 178: A lot of different things can go wrong here. Exception does not necessarily mean artifacts are not supported.
Maybe we should add list of supported features to config returned from backend.
@Herudaio how urgent is it to make new clients work for onprems?

neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -184,11 +193,18 @@ def __init__(self, credentials: Credentials, proxies: Optional[Dict[str, str]] =
             python_version=platform.python_version())
         self._http_client.session.headers.update({'User-Agent': user_agent})
 
+        if self.missing_features:
+            warnings.warn(NeptuneFeaturesNotAvailableWarning(self.missing_features).message)
```
aniezurawski, line 197: I am not a fan of printing this in every single Run.

aniezurawski (changes requested):

neptune/new/internal/backends/hosted_neptune_backend.py
pkasprzyk: it's (now) a python warning, so the use can setup it to print only once
`warnings.simplefilter("once", NeptuneFeaturesNotAvailableWarning)`
or choose to ignore it alltogether
`warnings.simplefilter("ignore", NeptuneFeaturesNotAvailableWarning)`

either way, your fight is with spec, not me ;)

pkasprzyk: spec has changed, so removing

aniezurawski (approved):

PR #716, opened 2021-09-27
Tests and file operations

Raalsky:


neptune/new/exceptions.py
```diff
@@ -681,24 +681,24 @@ def __init__(self, framework):
         ))
 
 
-class NeptuneStorageLimitException(NeptuneException):
```
aniezurawski, line 684: Let's keep old exception ifor a while n case someone try to import it.

neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -540,8 +540,10 @@ def _execute_operations(self,
             return [MetadataInconsistency(err.errorDescription) for err in result]
         except HTTPNotFound as e:
             raise RunUUIDNotFound(run_id=run_id) from e
-        except HTTPUnprocessableEntity:
-            raise NeptuneStorageLimitException()
+        except (HTTPPaymentRequired, HTTPUnprocessableEntity) as e:
+            raise NeptuneLimitExceedException(
+                reason=e.response.json().get("message", "Maximum storage limit reached")
```
aniezurawski, line 545: Have you tested compatibility with old and new api on real Neptune instance? Are you sure there is `message` field in response?

Raalsky: I thought that it was not implemented yet on the backend side currently (passing exact reason). It was tested on the current "stage" for storage limit. It was created based on legacy structure from https://github.com/neptune-ai/neptune-client/blob/2801c717b5c8ad983715421f2ce272be9b5816a9/neptune/internal/channels/channels_values_sender.py#L161. I've updated `"message"` to `"title"` which is based on current error response structure. What do you mean by old vs new API?

aniezurawski: Old - current response from leaderbaord.
New - what is under development on backend side right now.

neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -474,8 +474,11 @@ def _execute_operations(self,
             return [MetadataInconsistency(err.errorDescription) for err in result]
         except HTTPNotFound as e:
             raise RunUUIDNotFound(run_id=run_id) from e
-        except HTTPUnprocessableEntity:
-            raise NeptuneStorageLimitException()
+        except (HTTPPaymentRequired, HTTPUnprocessableEntity) as e:
+            print(e.response.json())
```
aniezurawski, line 478: print?

tests/neptune/new/internal/backends/test_hosted_neptune_backend.py
```diff
@@ -388,3 +393,47 @@ def test_cannot_resolve_host(self, gethostname_mock, _):
...
+    def test_limit_exceed(self, swagger_client_factory):
+        # given
+        swagger_client = self._get_swagger_client_mock(swagger_client_factory)
+        backend = HostedNeptuneBackend(credentials)
+        exp_uuid = str(uuid.uuid4())
+
+        # when:
+        error = MagicMock()
+        error.json.return_value = {
+            "message": "Maximum storage limit reached"
```
aniezurawski, line 406: message or title?

PR #720, opened 2021-09-30
handle projects limit reached in project mgmt api

pkasprzyk:


neptune/management/internal/api.py
```diff
@@ -40,7 +40,7 @@
     UserNotExistsOrWithoutAccess,
     WorkspaceNotFound,
     UserAlreadyHasAccess,
-    BadRequestException,
+    BadRequestException, ProjectsLimitReached,
```
Raalsky, line 43: Newline?

aniezurawski (dismissed):

neptune/management/internal/api.py
```diff
@@ -128,6 +128,9 @@ def create_project(
         if 'ERR_NOT_UNIQUE' in validation_errors:
             raise ProjectAlreadyExists(name=project_identifier) from e
         raise BadRequestException(validation_errors=validation_errors)
+    except HTTPUnprocessableEntity as e:
+        # TODO: better error identification
```
Raalsky: I don't think it's necessary here. Should be replaced by Jira task.

neptune/management/internal/api.py
pkasprzyk: sorry

Raalsky (approved):

PR #747, opened 2021-11-09
Prepare for AttributeContainers other than Run

pkasprzyk:


neptune/new/types/state.py
```diff
@@ -0,0 +1,23 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from enum import Enum
+
+
+class ContainerState(Enum):
```
aniezurawski, line 19: I think it can be internal.

neptune/new/attribute_container.py
```diff
@@ -0,0 +1,461 @@
+#
```
aniezurawski, line 1: Can it be internal?

neptune/new/run.py
```diff
@@ -141,397 +93,18 @@ def __init__(
...
-            >>> # this will print out all Atom attributes stored in run as a dict
-
-        You may also want to check `fetch docs page`_.
-
-        .. _fetch docs page:
-            https://docs.neptune.ai/api-reference/run#fetch
-        """
-        return self._get_root_handler().fetch()
-
     def ping(self):
```
aniezurawski, line 104: It should be moved to Container as well, shouldn't it?

aniezurawski (changes requested):

neptune/new/attribute_container.py
pkasprzyk: This was my initial thought, but it implements a lot of public methods (assign, fetch etc.) -- and I don't really want it hidden in internal class.

neptune/new/run.py
pkasprzyk: We thought (with @Herudaio) that a generic `Container` doesn't need to be pinged -- I was under the impression that `Project` won't have an `active` parameter (and no `running_time`, I think?), so pinging doesn't make sense.

neptune/new/types/state.py
pkasprzyk: Right, I'll move this.

neptune/new/run.py
pkasprzyk: discussed offline -- `Project` **can** have `active` and `running_time` , so moving `ping` to `AttributeContainer`

aniezurawski (approved):

PR #761, opened 2021-11-23
make Project an AttributeContainer

pkasprzyk:


neptune/new/internal/backends/neptune_backend_mock.py
```diff
@@ -105,30 +107,63 @@
...
+        return [Project(str(uuid.uuid4()), self.PROJECT_NAME, self.WORKSPACE_NAME)]
 
     def get_available_workspaces(self) -> List[Workspace]:
-        return [Workspace(str(uuid.uuid4()), "offline")]
+        return [Workspace(str(uuid.uuid4()), self.WORKSPACE_NAME)]
+
+    # pylint: disable=unused-argument
+    # FIXME: stop ignoring container_type on NPT-11117
+    def _create_container(
+        self, container_id: str, container_type: ContainerType, sys_id
```
Raalsky, line 147: Typing for `sys_id`?

neptune/new/project.py
```diff
@@ -30,9 +35,21 @@ class Project:
...
+        op_processor: OperationProcessor,
+        background_job: BackgroundJob,
+        lock: threading.RLock,
+        workspace: str,
+        project_name: str,
+    ):
+        super().__init__(
+            _id, backend, op_processor, background_job, lock, project_id=_id
+        )
+        self._workspace = workspace
```
Raalsky, line 51: Maybe `project_name` and `workspace` should be shared via `AttributesContainer` with Runs? Only type verifycation for non-null seems to be reasonable.

neptune/new/internal/init_run.py
```diff
@@ -20,8 +20,6 @@
 from platform import node as get_hostname
```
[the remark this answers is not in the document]
pkasprzyk: > `internal` doesn't mean to be "private"
it kinda should ;)

added backwards-compa layer, though

neptune/new/project.py
pkasprzyk: done, I think

Raalsky (approved):

PR #764, opened 2021-12-01
Require container_id and container_type in backend

shnela:


neptune/new/exceptions.py
```diff
@@ -275,9 +275,11 @@ def __init__(self, run_id: str) -> None:
         super().__init__("Run {} not found.".format(run_id))
 
 
-class RunUUIDNotFound(NeptuneException):
-    def __init__(self, run_id: str):
-        super().__init__("Run with ID {} not found. Could be deleted.".format(run_id))
+class ContainerUUIDNotFound(NeptuneException):
+    def __init__(self, container_id: str):
+        super().__init__(
+            "Container with ID {} not found. Could be deleted.".format(container_id)
```
pkasprzyk, line 281: I think this should get `container_type` to allow for nicer user-facing errors. Cf. spec in NPT-11119

neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -900,12 +926,14 @@ def fetch_atom_attribute_values(
...
             ]
         except HTTPNotFound:
-            raise RunUUIDNotFound(run_id)
+            raise ContainerUUIDNotFound(container_id)
 
+    # pylint: disable=unused-argument
     @with_api_exceptions_handler
-    def _get_file_set_download_request(self, run_id: str, path: List[str]):
+    def _get_file_set_download_request(self, container_id: str, path: List[str]):
+        # FIXME: NPT-11118 use container_type
```
pkasprzyk, line 934: that's required for this endpoint?

shnela: No, it won't be.

pkasprzyk (approved):

PR #768, opened 2021-12-06
Attribute copy

pkasprzyk:


neptune/new/internal/backends/neptune_backend_mock.py
```diff
@@ -618,6 +625,12 @@ def visit_delete_attribute(self, op: DeleteAttribute) -> Optional[Value]:
                 )
             return None
 
+        def visit_copy_attribute(self, op: CopyAttribute) -> Optional[Value]:
+            getter = op.source_attr_cls.getter
+            create_assignment_operation = op.source_attr_cls.create_assignment_operation
+            value = getter(self._backend, op.container_id, op.source_path)
+            return create_assignment_operation(self._path, value).accept(self)
```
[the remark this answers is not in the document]
pkasprzyk: Correct. In asynchronous processing it fetches the current value and adds new assignment op.

neptune/new/attributes/atoms/integer.py
```diff
@@ -13,22 +13,53 @@
...
 from neptune.new.attributes.atoms.atom import Atom
+from neptune.new.types.value_copy import ValueCopy
+
+if typing.TYPE_CHECKING:
+    from neptune.new.internal.backends.neptune_backend import NeptuneBackend
 
 
 class Integer(Atom):
-    def assign(self, value: Union[IntegerVal, float, int], wait: bool = False):
+    def copy(self, value: ValueCopy, wait: bool = False):
```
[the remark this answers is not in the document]
pkasprzyk: This MR in the current state is meant to be a proof of concept and base for discussion "does it make sense to do copying this way". In the final version both `copy` and `getter` will be implemented for all (non-file) Atoms, and some code will probably be moved around.

neptune/new/types/value_copy.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __init__(self, source_handler: "Handler"):
+        self.source_handler = source_handler
+
+    def accept(self, visitor: "ValueVisitor[Ret]") -> Ret:
+        # pylint: disable=protected-access
+        source_path = self.source_handler._path
+        source_attr = self.source_handler._run.get_attribute(source_path)
+        if source_path is None:
+            raise Exception()
```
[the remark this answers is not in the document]
pkasprzyk: actually, I think I can remove this `if` completely -- it's not possible for a handler to have `None` `source_path`, not sure why I put this here

tests/neptune/new/test_run.py
```diff
@@ -122,6 +122,13 @@ def test_assign_false(self):
 
         self.assertFalse(exp["params/predictor.cheat"].fetch())
 
+    def test_assign_copy(self):
+        exp = init(mode="debug")
+        exp["some/path/num"] = 42
+        exp["copied/path"] = exp["some/path/num"]
+        exp.wait()
+        self.assertEqual(42, exp["copied/path"].fetch())
+
```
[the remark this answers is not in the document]
pkasprzyk: yeah, that was really broken... fixed now, I hope -- thanks!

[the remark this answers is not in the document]
shnela: Have I broken this because String does not support copying yet?

```
        exp["some/path/num"] = "fo9o"
        exp["copied/path"] = 42
        exp["copied/path"] = exp["some/path/num"]
```

```
self = <neptune.new.internal.backends.neptune_backend_mock.NeptuneBackendMock.NewValueOpVisitor object at 0x7f0ef5fe2670>
op = CopyAttribute(path=['copied', 'path'], container_id='bdcc8dc7-a2b7-42a9-ba25-209a43888ed7', source_path=['some', 'path', 'num'], source_attr_cls=<class 'neptune.new.attributes.atoms.string.String'>)

    def visit_copy_attribute(self, op: CopyAttribute) -> Optional[Value]:
>       getter = op.source_attr_cls.getter
E       AttributeError: type object 'String' has no attribute 'getter'
```


Additionally it'd be good to have additional case with copying on existing value (the case recently broken).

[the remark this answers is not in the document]
pkasprzyk: > Have I broken this because String does not support copying yet?
yep
> Additionally it'd be good to have additional case with copying on existing value
will do

shnela (approved):

neptune/new/attributes/atoms/copiable_atom.py
```diff
@@ -0,0 +1,64 @@
...
+from neptune.new.internal.container_type import ContainerType
+from neptune.new.internal.operation import CopyAttribute
+from neptune.new.internal.utils.paths import parse_path
+from neptune.new.types.value_copy import ValueCopy
+
+if typing.TYPE_CHECKING:
+    from neptune.new.internal.backends.neptune_backend import NeptuneBackend
+
+
+class CopiableAtom(Atom):
```
shnela, line 29: Shouldn't be `CopyableAtom`

neptune/new/attributes/atoms/float.py
```diff
@@ -13,24 +13,34 @@
...
+        return AssignFloat(path, value)
+
+    @staticmethod
+    def getter(
+        backend: "NeptuneBackend", container_id: str, container_type: ContainerType, path: typing.List[str]
+    ) -> float:
+        val = backend.get_float_attribute(container_id, container_type, path)
+        return val.value
+
+    def assign(self, value: typing.Union[FloatVal, float, int], wait: bool = False):
```
shnela, line 39: From what I see we could produce generic `CopiableAtom.assign` which will first invoke overriden `verifyValue` method and then enqueue assigment operation since `create_assignment_operation` is overriden.

neptune/new/internal/backends/neptune_backend_mock.py
```diff
@@ -510,8 +511,14 @@ def visit_artifact(self, _: Artifact) -> AttributeType:
         def visit_namespace(self, _: Namespace) -> AttributeType:
             raise NotImplementedError
 
+        def copy_value(
+            self, source_type: Type[Attribute], source_path: List[str]
+        ) -> AttributeType:
+            raise NotImplementedError
```
shnela, line 517: We test it in non mocked backend, right?

neptune/new/internal/backends/operations_preprocessor.py
```diff
@@ -65,7 +66,7 @@ def process(self, operations: List[Operation]):
 
     def get_operations(self) -> List[Operation]:
         result = []
-        for _, acc in sorted(self._accumulators.items()):
+        for _, acc in self._accumulators.items():
```
shnela, line 69: We dropped sort, because we have to make sure that every batch is limited by last `copy` operation, right?

Why did we sorted in the first place?

shnela (changes requested):

neptune/new/internal/backends/neptune_backend_mock.py
pkasprzyk: Sure we do, but that's `AttributeTypeConverterValueVisitor` used to build a "backend" attribute type; we never get attributes of type "Copy" from backend, and we never store it in Mock's map.

neptune/new/internal/backends/operations_preprocessor.py
pkasprzyk: 1. yes, exactly
2. that's a great question. @aniezurawski maybe you know?

neptune/new/attributes/atoms/copiable_atom.py
pkasprzyk: https://www.merriam-webster.com/ which I usually use to solve such dilemmas lists neither.

"Oxford languages by Google" matches copiable: https://www.google.com/search?q=copiable&oq=copiable , Collins -- copyable: https://www.collinsdictionary.com/dictionary/english/copyable . Wiktionary lists both: https://en.wiktionary.org/wiki/copiable

I like the visual look of Copiable better and considering the source material I think it's acceptable ;).

neptune/new/attributes/atoms/float.py
pkasprzyk: That's true, but not sure if it's worth it tbh. 
With `verify_value` a separate def (with one use case) would still have to be implemented for each subclass, and we'd be saving two lines of code per subclass (`with lock / self._enqueue...`).

I've separated out `create_assignment_operation` mainly because it's used both in `assign` and in `CopyAttribute.resolve`.

neptune/new/internal/backends/operations_preprocessor.py
aniezurawski: I do not remember. It is sorted by path so it should not metter. Maybe it was easier to test it this way?

Anyway, are you guys sure it is good idea to change anything in operation_preprocessor? Shouldn't be part-batches preprocessed independently?

aniezurawski: I think you could just split a batch by copy operations and reuse existing code for every part-batch. Much simpler and elegant.

aniezurawski: Moreover we probably should ack after every part-batch, so maybe single call to execute_operations should just return number operations actually performed and do a single part-batch at the time?

pkasprzyk: That would change the idea that in processing a queue batch we do upload (and artifact) ops first -- we would do upload "subbatch", then other "subbatch", and upload again (there's a special comment saying "Upload operations should be done first since they are idempotent").

I'm not if we actually need that -- other than that, I'm perfectly OK with what you suggest.

aniezurawski: > That would change the idea that in processing a queue batch we do upload (and artifact) ops first

No, it would not. It would still apply but to potentially smaller batches. Just split first, and do upload-related logic later for every part-batch. ;)

pkasprzyk: > Moreover we probably should ack after every part-batch, so maybe single call to execute_operations should just return number operations actually performed and do a single part-batch at the time?

Right, that sounds good (although the "much simpler" may be lost on me ;))

pkasprzyk: done

shnela (approved):

PR #770, opened 2021-12-06
Renamed from parentId to experimentId in API calls

Raalsky:


shnela (changes requested):
There are `parentIdentifier` used in `neptune/new/internal/backends/hosted_artifact_operations.py` as well.

I think we should change it.

shnela (changes requested):
For `create_artifact_version` as well ;)

PR #771, opened 2021-12-08
Sync projects

shnela:


neptune/new/sync.py
```diff
@@ -147,7 +148,11 @@ def get_offline_runs_ids(base_path: Path) -> List[str]:
 def partition_runs(base_path: Path) -> Tuple[List[ApiRun], List[ApiRun], int]:
     synced_runs_ids = []
     unsynced_runs_ids = []
-    for run_path in (base_path / ASYNC_DIRECTORY).iterdir():
+    async_path = base_path / ASYNC_DIRECTORY
+    if not async_path.is_dir():
```
Raalsky, line 152: What is the purpose of it?

tests/neptune/new/test_client.py
```diff
@@ -269,14 +270,16 @@ def test_debug_mode(self):
...
-        project["some/variable"] = 13
-        with self.assertRaises(NeptuneOfflineModeFetchException):
-            project["some/variable"].fetch()
-        self.assertIn(str(project._id), os.listdir(".neptune/offline"))
-        self.assertIn(
-            "data-1.log", os.listdir(".neptune/offline/{}".format(project._id))
-        )
+        with self.assertRaises(NeptuneException):
+            init_project(name=self.PROJECT_NAME, mode="offline")
+        # project = init_project(name=self.PROJECT_NAME, mode="offline")
```
Raalsky, line 275: ?

neptune/new/sync.py
shnela: If we make `offline` run only, there will be `./neptune/offline`, but no `./neptune/assync` dir, so command would fail.
I've treated missing `async` dir as empty one.

tests/neptune/new/test_client.py
shnela: We've agreed with product that we're not allowing to invoke `init_project` in `offline` mode, we'll improve it when implementing ModelRegistry.

Raalsky: I know but please don't leave commented out code. I think black/pylint should warn about it.

shnela: But I think that we'll support this soon (several weeks from now), so we may easily uncomment it.
But if you're strongly against I can drop it. Your call ;)

Raalsky: Drop it. Things may change, memory is volatile :wink:

Raalsky (approved):

PR #784, opened 2021-12-14
Limit size of in-memory File content and string atoms

aniezurawski:


neptune/new/attributes/atoms/string.py
```diff
@@ -13,19 +13,45 @@
...
+        ):
+            # the first truncation
+            self._value_truncation_occurred = True
+            click.echo(
+                f"Warning: string '{path_to_str(self._path)}' value was "
+                f"longer than {String.MAX_VALUE_LENGTH} characters and was truncated. "
+                f"This warning is printed only once.",
+                err=True,
+            )
+        value.value = value.value[: String.MAX_VALUE_LENGTH]
```
shnela, line 53: It looks like NOP when previous' if expression is `False` but still I'd prefer truncation to be inside `if` block.

pkasprzyk (approved):

pkasprzyk (approved):

PR #793, opened 2021-12-27
AttributesContainer for both Model and ModelVersion

shnela:


neptune/new/internal/operation.py
```diff
@@ -553,3 +554,30 @@ def resolve(self, backend: "NeptuneBackend") -> Operation:
             backend, self.container_id, self.container_type, self.source_path
         )
         return create_assignment_operation(self.path, value)
+
+
+@dataclass
+class ChangeStage(Operation):
+    container_id: str
+    stage: str
```
Raalsky, line 562: I think the internal state could be `ModelVersionStage` and the `value` of it could be used only in serialization/deserialization + the constructor will be the default for dataclass

neptune/new/internal/operation.py
```diff
@@ -553,3 +554,30 @@ def resolve(self, backend: "NeptuneBackend") -> Operation:
...
+        ret = super().to_dict()
+        ret["container_id"] = self.container_id
+        ret["stage"] = self.stage
+        return ret
+
+    @staticmethod
+    def from_dict(data: dict) -> "ChangeStage":
+        return ChangeStage(
+            container_id=data["container_id"],
+            stage=ModelVersionStage(data["stage"]),
```
Raalsky, line 582: Hmmmm. Inconsistency? But I prefer to do it this way

neptune/new/model.py
```diff
@@ -0,0 +1,64 @@
...
+            _id,
+            backend,
+            op_processor,
+            background_job,
+            lock,
+            project_id,
+            project_name,
+            workspace,
+        )
+        self._sys_id = sys_id
```
Raalsky, line 60: Made `sys_id` a part of `AttributesContainer`

Raalsky: Or introduce additional class which will be common for `Run` etc. and use multiinheritance

shnela: To discuss in other issue.

Raalsky (approved):

Raalsky (approved):

PR #795, opened 2021-12-28
Added GPU Monitor to new client with proper url

Raalsky:


shnela (changes requested):
Add changelog and it's ok.

shnela (approved):

PR #801, opened 2021-12-30
Move e2e tests to main repository

Raalsky:


shnela (changes requested):
I think you should move all requirements to `requirements` directory, also `requirements.txt`.

shnela (dismissed):

shnela (approved):

PR #809, opened 2022-01-10
Add ModelVersion.change_stage implementation

shnela:


neptune/new/model_version.py
```diff
@@ -65,7 +67,19 @@ def __init__(
...
         return self._sys_id
 
-    def change_stage(self, stage: str, wait=False):
-        self._op_processor.enqueue_operation(
-            ChangeStage(container_id=self._id, stage=ModelVersionStage(stage)), wait
+    def change_stage(self, stage: str):
+        stage = ModelVersionStage(stage)
+
+        if self._op_processor.mode == Mode.OFFLINE:
+            raise NeptuneOfflineModeChangeStageException()
```
shnela, line 74: I wasn't 100% sure about adding `OperationProcessor.model` attribute, but I think it's better than checking `op_processor` class type.

neptune/new/exceptions.py
```diff
@@ -775,6 +779,36 @@ def __init__(self):
...
+        message = """
+        TODO: NPT-11349
+{h1}
+----NeptuneOfflineModeChangeStageException---------------------------------------------------
+{end}
+It seems you are trying to change stage, while working in an offline mode.
+You need to work in non-offline connection mode to change stage.
+
+You can set connection mode when creating a new run:
+    {python}run = neptune.init(mode="async"){end}
```
pkasprzyk, line 793: example shouldn't list a run, as it's a model_version feature

neptune/new/handler.py
```diff
@@ -38,32 +40,49 @@
...
-    from neptune.new.run import Run
+    from neptune.new.attribute_container import AttributeContainer
+
+
+def check_protected_paths(fun):
+    @wraps(fun)
+    def inner_fun(self: "Handler", *args, **kwargs):
+        # pylint: disable=protected-access
+        for path in self.PROTECTED_PATHS:
+            if self._path == path:
```
pkasprzyk, line 51: `it self._path in self.PROTECTED_PATHS:`?

neptune/new/handler.py
```diff
@@ -38,32 +40,49 @@
...
+                raise NeptuneProtectedPathException(path)
+        return fun(self, *args, **kwargs)
+
+    return inner_fun
 
 
 class Handler:
-    def __init__(self, run: "Run", path: str):
+    # paths which can't be modified by client directly
+    PROTECTED_PATHS = [
```
pkasprzyk, line 60: name it `_PROTECTED_PATHS`? it's an implementation detail for me

neptune/new/handler.py
```diff
@@ -291,6 +316,7 @@ def pop(self, path: str = None, wait: bool = False) -> None:
         "fetch_files_list",
     ]
 
+    @check_protected_paths
```
pkasprzyk, line 319: This (and a few others) won't work, I'm afraid -- you'll never reach the annotated version due to magic `__getattribute__` + `DOCSTRING_ATTRIBUTES`.

Simplest fix I can think of would be to add `super().__getattribute__(attribute_name)` to `__getattribute__` and check whether the (unused) method is wrapped in `check_protected_paths`. You're welcome to thinking up something nicer ;)

neptune/new/handler.py
```diff
@@ -307,6 +333,7 @@ def remove(self, values: Union[str, Iterable[str]], wait: bool = False) -> None:
         """
         raise NeptuneException("Should be never called.")
 
+    @check_protected_paths
     def clear(self, wait: bool = False):
         """Removes all tags from the `StringSet`.
```
pkasprzyk, line 339: can't comment on unchanged lines, but `delete_files` also needs `@check_protected_paths` IMO

neptune/new/internal/operation_processors/async_operation_processor.py
```diff
@@ -40,6 +41,8 @@ class AsyncOperationProcessor(OperationProcessor):
     STOP_QUEUE_STATUS_UPDATE_FREQ_SECONDS = 30
     STOP_QUEUE_MAX_TIME_NO_CONNECTION_SECONDS = 300
 
+    mode = Mode.SYNC
```
pkasprzyk, line 44: `AsyncOperationProcessor.mode == SYNC`. Tell me it's a typo.

neptune/new/model_version.py
pkasprzyk: I guessss. Thing is `SyncOperationProcessor` is used in `mode == Mode.DEBUG`, which cause a little bit o weird situation for me

tests/neptune/new/client/test_model_version.py
```diff
@@ -101,3 +104,21 @@ def test_wrong_parameters(self):
...
+
+    @patch(
+        "neptune.new.internal.backends.neptune_backend_mock.NeptuneBackendMock.get_attributes",
+        new=lambda _, _uuid, _type: [Attribute("sys/stage", AttributeType.STRING)],
+    )
+    def test_change_stage(self):
+        exp = self.call_init()
+        exp.change_stage(stage="production")
+
+        self.assertEqual("production", exp["sys/stage"].fetch())
```
pkasprzyk, line 116: I'd prefer either `ModelVersionStage.PRODUCTION.value` both here and in `test_protected_paths`, or `"production"` both here in `test_protected_paths` (I probably like the latter more)

neptune/new/handler.py
shnela: Shame on me...

neptune/new/model_version.py
shnela: Right, I didn't notice that. Let's don't use mode then.

neptune/new/exceptions.py
shnela: Let's don't change this for now. I've added `TODO: NPT-11349` since it's going to be udpated in the same story.

neptune/new/handler.py
shnela: Ok

neptune/new/handler.py
shnela: Yes

neptune/new/exceptions.py
pkasprzyk: Yeah, but you're actually adding incorrect docstring in this PR, this is error-prone IMO. Either add a correct one or do as in `NeptuneProtectedPathException` (empty message).

tests/neptune/new/client/test_model_version.py
shnela: Yes, good point.

neptune/new/exceptions.py
shnela: Right, ok.

neptune/new/handler.py
shnela: Yes, ofc it won't work...

pkasprzyk (approved):

PR #811, opened 2022-01-11
Replace AttributeContainer with MetadataContainer with dedicated sub-module

shnela:


neptune/new/__init__.py
```diff
@@ -32,7 +32,7 @@
     NeptuneUninitializedException,
     NeptunePossibleLegacyUsageException,
 )
-from neptune.new.run import Run
+from neptune.new.attributes_containers import Run
```
shnela: Maybe we should call this module `containers` so the imports are shorter?

neptune/new/attributes_containers/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from neptune.new.attributes_containers.attribute_container import AttributeContainer
+from neptune.new.attributes_containers.handler import Handler
```
shnela, line 18: `Handler` is something in the middle between `Attribute` and `AttributeContainer`, I'm not 100% sure about moving it here.

neptune/new/attributes_containers/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+from neptune.new.attributes_containers.model import Model
+from neptune.new.attributes_containers.model_version import ModelVersion
+
+# from neptune.new.attributes_containers.model_versions_table import (
+#     ModelVersionsTable,
+#     ModelVersionsTableEntry,
+# )
+from neptune.new.attributes_containers.project import Project
+from neptune.new.attributes_containers.run import Run
```
shnela, line 27: I think that moving generic `AttributeContainer`, `Model`, `Run` etc as `module-level` import is good idea, to simplify imports in other parts of code.

I'm not 100% sure about placing `*Table` and `*TableEntry` here.

neptune/new/__init__.py
```diff
@@ -32,7 +32,7 @@
     NeptuneUninitializedException,
     NeptunePossibleLegacyUsageException,
 )
-from neptune.new.run import Run
+from neptune.new.attributes_containers import Run, Project, Model, ModelVersion
```
shnela, line 35: Should we make `*Table` and `*TableEntry` objects importable for our users as well?

neptune/new/attributes/atoms/string.py
```diff
@@ -24,7 +24,7 @@
 from neptune.new.types.atoms.string import String as StringVal
 
 if typing.TYPE_CHECKING:
-    from neptune.new.attribute_container import AttributeContainer
+    from neptune.new.attributes_containers import AttributeContainer
```
pkasprzyk, line 27: I have to admit: plural "attributes containers" does sound BAD for me (vs attribute container)

neptune/new/__init__.py
```diff
@@ -32,7 +32,7 @@
     NeptuneUninitializedException,
     NeptunePossibleLegacyUsageException,
 )
-from neptune.new.run import Run
+from neptune.new.metadata_containers import Run, Project, Model, ModelVersion
```
pkasprzyk, line 35: Haven't we come to a conclusion NOT to do this?

neptune/new/runs_table.py
```diff
@@ -17,44 +17,20 @@
...
-            attributes=attributes,
-        )
-
-
-class RunsTable(AttributesContainersTable):
-    table_entry_cls = RunsTableEntry
-
-    def to_runs(self):
-        return self.to_table_entries()
+from neptune.new.metadata_containers.runs_table import RunsTable, RunsTableEntry
```
pkasprzyk, line 36: I know you have a separate issue for deprecation https://neptune-labs.atlassian.net/browse/NPT-11395 , but please either leave an easily greppable comment in files like that or add them to NPT-11395 description

neptune/new/__init__.py
shnela: Yes we did, I'l fix.

neptune/new/runs_table.py
shnela: Isn't

```
# backwards compatibility
# pylint: disable=unused-import,wrong-import-order
```

enough to find such places?

pkasprzyk: right, sorry -- I've seen you've removed from https://github.com/neptune-ai/neptune-client/pull/811/files#diff-51310aa42e9441fc7625038b9dffa9f353864634268799198b624f273f74f4ecL29 and got lost a little bit

pkasprzyk (approved):

PR #815, opened 2022-01-13
Do not forward every attribute from Handler to Attribute

shnela:


neptune/new/handler.py
```diff
@@ -435,23 +414,30 @@ def download_last(self, destination: str = None) -> None:
...
         """
-        raise NeptuneException("Should be never called.")
+        return self._pass_call_to_attr(function_name="fetch_files_list")
+
+    def _pass_call_to_attr(self, function_name, **kwargs):
+        attr = self._run.get_attribute(self._path)
+        if attr:
+            return getattr(attr, function_name)(**kwargs)
+        else:
+            raise AttributeError(f"No such method '{function_name}'.")
```
shnela, line 440: I was thinking that we could add some information, that there's no such method, because field is not "initialized" yet, but:
1. it's quite due to our internal implementation, and probably this should be not user facing info
2. I wanted to keep behaviour compliant with the previous one

tests/neptune/new/client/abstract_experiment_test_mixin.py
```diff
@@ -72,6 +73,16 @@ def test_async_mode(self):
                 os.listdir(".neptune/async/{}/{}".format(exp._id, execution_dir)),
             )
 
+    def test_missing_attribute(self):
+        exp = self.call_init(mode="debug")
+        with self.assertRaises(MissingFieldException):
+            exp["non/existing/path"].fetch()
```
shnela, line 79: Waiting for @Herudaio accept.

pkasprzyk (approved):

pkasprzyk:
LGTM -- please update changelog and you can merge it

PR #817, opened 2022-01-13
Do not use containerType in DiskQueue

shnela:


tests/neptune/new/sync/utils.py
```diff
@@ -0,0 +1,140 @@
...
+        path / "async" / str(sync_project.id) / execution_id / "last_ack_version"
+    ).write(2)
+    SyncOffsetFile(
+        path / "async" / str(sync_project.id) / execution_id / "last_put_version"
+    ).write(2)
+
+    return unsync_project, sync_project, generate_get_run_impl(registered_projects)
+
+
+def prepare_runs(path):
```
pkasprzyk, line 73: please add a run with "old" storage path (`async/{exp._id}`)

neptune/new/constants.py
```diff
@@ -31,3 +31,7 @@
 ASYNC_DIRECTORY = "async"
 
 OFFLINE_NAME_PREFIX = "offline/"
+CONTAINER_DIR_REGEXP = (
```
pkasprzyk, line 34: Really not sure about usefulness of this. Couldn't we just `split("__")` and be done with it?

neptune/new/sync/sync.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def register_offline_runs(
+        self, base_path: Path, project: Project, offline_dirs: Iterable[str]
+    ) -> List[ApiExperiment]:
+        result = []
+        for dir_ in offline_dirs:
+            dir_path = base_path / OFFLINE_DIRECTORY / dir_
+            if dir_path.is_dir():
+                run, registered = self._register_offline_run(
+                    project, container_type=ContainerType.RUN
```
pkasprzyk, line 178: that looks wrong -- `dir_` should contain `type` (at least in some cases)

neptune/new/sync/sync.py
```diff
@@ -0,0 +1,233 @@
...
+                )
+                if run:
+                    self._move_offline_run(
+                        base_path=base_path,
+                        offline_dir=dir_,
+                        server_id=run.id,
+                        server_type=run.type,
+                    )
+                    verb = "registered as" if registered else "recognized as"
+                    click.echo(f"Offline run {dir_} {verb} {get_qualified_name(run)}")
```
pkasprzyk, line 188: current master has `Offline {container_type.value} ...`, why change that since you have `run.type`?

neptune/new/sync/sync.py
```diff
@@ -0,0 +1,233 @@
...
+        self,
+        base_path: Path,
+        project_name: Optional[QualifiedName],
+        offline_run_ids: Sequence[str],
+    ):
+        if offline_run_ids:
+            project = get_project(project_name, backend=self._backend)
+            if not project:
+                raise CannotSynchronizeOfflineRunsWithoutProject
+            offline_run_dirs = [f"run__{run_id}" for run_id in offline_run_ids]
```
pkasprzyk, line 207: why purposefully make this unable to process `model__3e849bc9-6ab3-4c00-87ee-5a56c2a02a3c`?

neptune/new/sync/sync.py
```diff
@@ -0,0 +1,233 @@
...
+            registered_runs = self.register_offline_runs(
```
pkasprzyk, line 208: also, what about existing offline dirs, `3e849bc9-6ab3-4c00-87ee-5a56c2a02a3c`?

neptune/new/sync/utils.py
```diff
@@ -0,0 +1,167 @@
...
+        yield container_type, unique_id, path
+
+
+def get_offline_runs_ids(base_path: Path) -> List[str]:
+    result = []
+    if not (base_path / OFFLINE_DIRECTORY).is_dir():
+        return []
+    for run_path in (base_path / OFFLINE_DIRECTORY).iterdir():
+        dir_name = run_path.name
+        run_id = dir_name[len("run__") :]
```
pkasprzyk, line 159: 1. what about existing offline dirs, 3e849bc9-6ab3-4c00-87ee-5a56c2a02a3c?
2. why not use fancy `CONTAINER_DIR_REGEXP` (or `split("__")``)?

tests/neptune/new/sync/utils.py
```diff
@@ -0,0 +1,140 @@
...
+        / "last_put_version"
+    ).write(2)
+
+    return unsync_exp, sync_exp, generate_get_run_impl(registered_runs)
+
+
+def prepare_offline_run(path):
+    offline_exp_uuid = str(uuid.uuid4())
+    offline_exp_path = path / OFFLINE_DIRECTORY / f"run__{offline_exp_uuid}"
+    offline_exp_path.mkdir(parents=True)
```
pkasprzyk, line 128: please add another run with "old" storage path (`offline/{exp._id}`)

pkasprzyk (changes requested):

neptune/new/sync/sync.py
shnela: I didn't restore that, but still. We support only run for now.

I didn't want to worry about other types yet.

neptune/new/sync/sync.py
shnela: Because it's not part of this issue.

pkasprzyk: What you're saying make 0 sense to me.

In `iterate_experiments` you're using regex to catch all metadata container types. In `sync_selected_registered_runs` you're using `run.type`, in `sync_all_registered_runs` you're also correctly handling all the types.

Here you could do the same with minimal work (you have the container type in path), yet you purposefully decided to:
1. hardcode "run"
2. break syncing offline runs that have no "run__" prefix (i.e. existing runs from previous client versions).

Why? So someone else will have to find all the parts you left broken and fix them in a separate issue?

tests/neptune/new/sync/utils.py
shnela: Tests were refactored and there are separate cases for deprecated RUN format ;)

tests/neptune/new/sync/utils.py
shnela: Tests were refactored and there are separate cases for deprecated RUN format

neptune/new/sync/sync.py
shnela: Right, fixed. I'm trying both option now.

neptune/new/sync/sync.py
shnela: I've changed this code, but kept explicit assumption about `RUN` type.

I didnt' want to support other types here, because it's not a straightforward change.
There's for example case when user manually types UUID of offline run to sync, and now I don't know whether product will agree on changing this to `model_UUID` format, or assume separate input groups for script.
Right now I've simply assumed `UUID` input format for offline runs.
So I don't want to write code matching given UUID with some (maybe) existing patch to determine type, if there's chance that this code will have to be removed.

We've explicitly assumed support of offline mode for runs only for now.

We have separate story for other types, as it requires contact with product and further code refactor.

Additionally supporting other than run types in register_offline_**runs** is odd to me, and at the same time I'd like to merge this branch instead of refactoring this again and again.

neptune/new/constants.py
shnela: But it's used in other places now.
It's a nice way to rely on regexp group to determine run UUID since `<type>__` prefix is optional.

pkasprzyk: Actually, it's not nice. There is no contract that container IDs are/contain UUID AFAIK. The format we use is "<type>__<unique_id>" and that's what the should check for IMO (again, `split("__")`?)

neptune/new/sync/utils.py
```diff
@@ -162,11 +162,25 @@ def get_offline_runs_ids(base_path: Path) -> List[str]:
...
+            raise ValueError(f"Wrong container directory name: {dir_name}")
+        run_id = match.group(3)
         result.append(run_id)
     return result
 
 
+def get_offline_run_dir_path(base_path: Path, run_id: UniqueId):
+    for dir_ in [
+        base_path / OFFLINE_DIRECTORY / run_id,
+        base_path / OFFLINE_DIRECTORY / f"run__{run_id}",
```
pkasprzyk, line 176: OK, I think now you're purposefully writing broken code which will have to be removed around next PR.

You could just list directories in `offline` and try to recognize them as container IDs (exactly as it is done for `sync`??), yet you do a "hit-and-miss" strategy of checking for hypothetical dirnames. Why?

neptune/new/constants.py
shnela: Ok

neptune/new/sync/utils.py
shnela: I've mentioned this problem [here](https://github.com/neptune-ai/neptune-client/pull/817#discussion_r793751495).

But as we've finally agreed on.
Right now I'm listing and accepting parameters which are indentical as directory in which queue is stored.

So user has to pass full directory name: `run__<uuid>` (except situation with depracted run format without type prefix).
I don't especially asume `run_` prefix anywhere, but in the end if  prefix doesn't match to any existing `ContainerType` nor is `ContainerType.run` the program will finally crash.

@pkasprzyk right?

neptune/new/sync/status.py
```diff
@@ -0,0 +1,129 @@
...
+        unsynced_runs = []
+        async_path = base_path / ASYNC_DIRECTORY
+        for container_type, container_id, path in iterate_containers(async_path):
+            metadata_container = get_metadata_container(
+                container_id=container_id,
+                container_type=container_type,
+                backend=self._backend,
+            )
+
+            if is_run_synced(path):
```
pkasprzyk, line 67: I have to say I really don't care for the way you're renamed some functions to e.g. `partition_containers`, yet kept others e.g. `is_run_synced`. But I guess it's going to be cleaned up in a future task.

tests/neptune/new/utils/api_experiments_factory.py
```diff
@@ -13,56 +13,72 @@
...
-        container_type=ContainerType.PROJECT,
+def api_model_version() -> ApiExperiment:
+    return _api_container_type(
+        sys_id=f"{_random_key()}-{_random_key()}-{randint(42, 12342)}",
+        container_type=ContainerType.MODEL_VERSION,
     )
 
 
-def _api_experiment(sys_id: str, container_type: ContainerType):
+def _api_container_type(sys_id: str, container_type: ContainerType) -> ApiExperiment:
```
pkasprzyk, line 72: that's not really a helpful function name...

tests/neptune/new/utils/api_experiments_factory.py
```diff
@@ -13,56 +13,72 @@
...
-        sys_id=sys_id,
+        sys_id=SysId(sys_id),
         workspace="workspace",
         project_name="sandbox",
         trashed=False,
     )
 
 
-def _random_key(n=3):
+def _random_key(n: int = 3) -> str:
```
pkasprzyk, line 83: since you were touching that, `n` is a terrible parameter name -- it doesn't suggest in any way it's key length

tests/neptune/new/utils/api_experiments_factory.py
```diff
@@ -13,56 +13,72 @@
...
     )
 
 
-def api_project():
-    return _api_experiment(
-        sys_id=_random_key(),
-        container_type=ContainerType.PROJECT,
+def api_model_version() -> ApiExperiment:
+    return _api_container_type(
+        sys_id=f"{_random_key()}-{_random_key()}-{randint(42, 12342)}",
```
pkasprzyk, line 67: don't you think that "refactoring" should include separating out this funny copy-pasted `randint(42, 12342)`?

neptune/new/sync/utils.py
```diff
@@ -0,0 +1,178 @@
...
+            textwrap.fill(_project_not_found_message(project_name)), file=sys.stderr
+        )
+        return None
+
+
+def get_qualified_name(run: ApiExperiment) -> str:
+    return "{}/{}/{}".format(run.workspace, run.project_name, run.sys_id)
+
+
+def is_valid_uuid(val: Any) -> bool:
```
pkasprzyk, line 116: I think this one's unused now

neptune/new/sync/utils.py
```diff
@@ -0,0 +1,178 @@
...
+        dir_name = path_.name
+        result.append(UniqueId(dir_name))
+    return result
+
+
+def is_offline_run_name(name: str) -> bool:
+    if not name.startswith(OFFLINE_NAME_PREFIX):
+        return False
+
+    # just make sure that dir has accepted format or raise exception
```
pkasprzyk, line 176: err, what? this function shouldn't raise -- it should return `True/False`

neptune/new/sync/sync.py
pkasprzyk: discussed offline, marking as resolved

pkasprzyk (approved):

neptune/new/internal/backends/neptune_backend_mock.py
```diff
@@ -799,3 +810,6 @@ def _create_type_error(self, op_name, expected):
                     op_name, self._path, expected, type(self._current_value)
                 )
             )
+
+    class ContainerIndex:
```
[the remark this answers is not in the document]
shnela: Sorry, I wanted to improve `NeptuneBackendMock` a bit, but finally I didn't use it, I've used regular mock as fake backend in tests.

neptune/new/sync/abstract_backend_runner.py
```diff
@@ -0,0 +1,28 @@
...
+
+__all__ = ["AbstractBackendRunner"]
+
+import abc
+
+from neptune.new.internal.backends.neptune_backend import NeptuneBackend
+
+
+class AbstractBackendRunner(abc.ABC):
+    _backend: NeptuneBackend
```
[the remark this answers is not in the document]
shnela: What would you prefer?
Passing backend for every function?

neptune/new/sync/status.py
shnela: Yes, it will.

tests/neptune/new/utils/api_experiments_factory.py
shnela: You're right.

tests/neptune/new/utils/api_experiments_factory.py
shnela: Ok

tests/neptune/new/utils/api_experiments_factory.py
shnela: To be honest not.
I don't think that generating random number generates as much code, to be worth moving it to separate function like with `_random_key`.
But if you insist, I can move it.

neptune/new/sync/utils.py
shnela: +1

neptune/new/sync/utils.py
shnela: Ok

shnela: Ok, right.
I've also changed script behaviour.

If user passes wrong offline path, he gets info that such run is not found, instead of info about wrong format.

neptune/new/internal/operation_processors/factory.py
```diff
@@ -45,10 +47,10 @@ def get_operation_processor(
     lock: threading.RLock,
     flush_period: float,
 ) -> OperationProcessor:
-
     if mode == Mode.ASYNC:
-        data_path = "{}/{}/{}".format(
-            NEPTUNE_DATA_DIRECTORY, ASYNC_DIRECTORY, container_id
+        data_path = (
+            f"{NEPTUNE_DATA_DIRECTORY}/{ASYNC_DIRECTORY}"
+            f"/{container_type.value}__{container_id}"
```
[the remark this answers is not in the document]
shnela: +1

neptune/new/internal/operation_processors/factory.py
```diff
@@ -38,6 +38,8 @@
 
 __all__ = ["get_operation_processor"]
 
+from ...sync.utils import create_dir_name
```
pkasprzyk, line 41: please format this as a normal import (without leading ellipsis ;)) + put it with the others (before `__all__`)

neptune/new/sync/abstract_backend_runner.py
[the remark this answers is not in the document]
pkasprzyk: No, declaring attribute `_backend` (and passing it) in both `SyncRunner` and `StatusRunner`. There's apparently no reason for them to have a common superclass (one attribute is not a reason, IMO).

tests/neptune/new/utils/api_experiments_factory.py
pkasprzyk: I'm was thinking about encapsulating magic numbers 42, 12342 rather that generating a random number itself. But nevermind, I guess.

tests/neptune/new/utils/api_experiments_factory.py
```diff
@@ -80,5 +82,7 @@ def _api_container_type(sys_id: str, container_type: ContainerType) -> ApiExperi
     )
 
 
-def _random_key(n: int = 3) -> str:
-    return "".join((random.choice(string.ascii_letters).upper() for _ in range(n)))
+def _random_key(key_length: int = 3) -> str:
+    return "".join(
+        (random.choice(string.ascii_letters).upper() for _ in range(key_length))
```
pkasprzyk, line 87: this parenthesis is redundant

neptune/new/sync/abstract_backend_runner.py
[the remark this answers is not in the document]
shnela: > (one attribute is not a reason, IMO)

I agree.

But I wanted to have common ancestor for both `StatusRunner` `SyncRunner` as they are both responsible for handling neptune click commands.

I like to group `alike` objects by giving them common parent class.

neptune/new/internal/operation_processors/factory.py
shnela: Sorry, I didn't notice this.

And definitely I should move `__all__` before imports.

neptune/new/sync/abstract_backend_runner.py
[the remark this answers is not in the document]
pkasprzyk: Makes sense, I guess.

pkasprzyk (approved):

PR #818, opened 2022-01-14
Fixed when invalid workspace was provided

Raalsky:


neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -207,12 +207,20 @@ def get_project(self, project_id: str) -> Project:
                 raise NeptuneLegacyProjectException(project_id)
             return Project(project.id, project.name, project.organizationName)
         except HTTPNotFound:
+            available_workspaces = self.get_available_workspaces()
+
+            if workspace and not any(
+                filter(lambda aw: aw.name == workspace, available_workspaces)
```
aniezurawski, line 213: `any()` takes an iterable of booleans, doesn't it? So i think it should be either:
```any(map(lambda aw: aw.name == workspace, available_workspaces)```
or
```list(filter(lambda aw: aw.name == workspace, available_workspaces)```

Your code is correct as long as `workspace` evaluates to True (is not empty and is not None). But it's weird to read.

File: neptune/new/internal/backends/hosted_neptune_backend.py [hunk 2]
```diff
@@ -207,12 +207,20 @@ def get_project(self, project_id: str) -> Project:
...
+            ):
+                # Could not found specified workspace, forces listing all projects
+                workspace = None
+            else:
+                available_workspaces = list()
```
aniezurawski, line 218: Why reset this list if workspace is None?
Can we just skip fetching available worksapces?

aniezurawski (changes requested):

neptune/new/internal/backends/hosted_neptune_backend.py
Raalsky: Ok, list was already used in a few lines above :wink:

pkasprzyk: and how would you like `next(filter(lambda aw: aw.name == workspace, available_workspaces), None) is None`? :D

Raalsky: Nah, looks like a Python 2.x code :P

pkasprzyk (approved):

aniezurawski (approved):

PR #825, opened 2022-01-18
Handling exception of invalid header and strip token

Raalsky:


neptune/new/internal/backends/utils.py
```diff
@@ -99,6 +100,10 @@ def wrapper(*args, **kwargs):
 
             try:
                 return func(*args, **kwargs)
+            except requests.exceptions.InvalidHeader as e:
```
pkasprzyk, line 103: that sounds like a perfect scenario for an e2e test, don't you think? :D

pkasprzyk: ok, it turns out it's surprisingly hard to get into that branch after adding `.strip()` -- so I'm resolving that

pkasprzyk (approved):

PR #835, opened 2022-01-27
Experiment related refactor

shnela:


neptune/new/exceptions.py
```diff
@@ -191,7 +238,26 @@ def __init__(
...
+        self.container_id = container_id
+        self.container_type = container_type
+        super().__init__(
+            "{} with ID {} not found. Could be deleted.".format(
+                container_type.value.capitalize(), container_id
+            )
+        )
+
+
+RunUUIDNotFound = ContainerUUIDNotFound
```
pkasprzyk, line 255: 1. We don't raise it anywhere anymore, right?
2. Could you add a comment that it's only for backward compat etc?

neptune/new/exceptions.py
```diff
@@ -191,7 +238,26 @@ def __init__(
...
+
+
+class ProjectNotFoundWithSuggestions(
```
pkasprzyk, line 258: is this the only class extending `ExceptionWithProjectsWorkspacesListing`? if yes, I'd prefer to put its logic here (if nothing else, so we don't have to think/worry about multiple inheritance...).

neptune/new/exceptions.py
shnela: I've added a comment.

neptune/new/exceptions.py
pkasprzyk: ok, no discussion then

pkasprzyk (approved):

PR #836, opened 2022-01-27
Fixed almost all of e2e tests

Raalsky:


neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -281,7 +281,7 @@ def get_metadata_container(
                 .result
             )
 
-            if experiment.type != container_type:
+            if experiment.type != container_type.value:
```
pkasprzyk, line 284: You could actually change `ContainerType` class from `class ContainerType(enum.Enum)` to `class ContainerType(str, enum.Enum)` and solve such problems once and for all. Makes sense?

neptune/new/handler.py
```diff
@@ -454,7 +454,9 @@ def download_last(self, destination: str = None) -> None:
         .. _download_last docs page:
            https://docs.neptune.ai/api-reference/field-types#.download_last
         """
-        return self._pass_call_to_attr(function_name="remove", destination=destination)
+        return self._pass_call_to_attr(
+            function_name="download_last", destination=destination
```
pkasprzyk, line 458: :open_mouth:

e2e_tests/conftest.py
```diff
@@ -44,6 +45,8 @@ def environment():
         api_token=admin_token,
     )
 
+    time.sleep(10)
```
pkasprzyk, line 48: is it really necessary? :(

10s wait in almost(?) every test is a lot...

Raalsky: huh? Not in every. Only once or twice as it's inside of session-wide pytest fixture

pkasprzyk: Right, I've missed that `scope="session"`. Sorry.

pkasprzyk (approved):

PR #840, opened 2022-02-01
Fix ProjectNotFound message

shnela:


Raalsky (approved):

pkasprzyk:
@shnela I thing we also discovered that this exception is misused somewhere. Is this fixed in another PR?

PR #846, opened 2022-02-11
Added exception to handle model key conflict

Raalsky:


neptune/new/exceptions.py
```diff
@@ -1131,3 +1131,26 @@ def __init__(self, missing_feature):
...
+
+class NeptuneExperimentCreationConflict(NeptuneException):
+    pass
+
+
+class NeptuneKeyAlreadyExistsError(NeptuneExperimentCreationConflict):
+    def __init__(self, model_key, models_tab_url):
+        message = """
+{h1}
+----NeptuneKeyAlreadyExistsError---------------------------------------------------
```
pkasprzyk, line 1144: project also has a key (and it also has to be unique) -- so combination of `NeptuneKeyAlreadyExistsError` (without saying that it's about model) and `model_key` parameter isn't good IMO

neptune/new/exceptions.py
```diff
@@ -1131,3 +1131,26 @@ def __init__(self, missing_feature):
 """
         self.message = message.format(missing_feature=missing_feature, **STYLES)
         super().__init__(message)
+
+
+class NeptuneExperimentCreationConflict(NeptuneException):
```
pkasprzyk, line 1136: if uncaught, this exception can reach user -- so I'm not very enthusiastic about using the "Experiment" word

Raalsky: Any suggestion?

neptune/new/exceptions.py
Raalsky: True

neptune/new/exceptions.py
pkasprzyk: I think my preferred course of action would be to just:
1. delete `NeptuneExperimentCreationConflict` altogether, let `HTTPConflict` fly from `HostedNeptuneBackend`
2. make `NeptuneKeyAlreadyExistsError` (after rename, possibly - cf. other comment ;)) simply a subclass of `NeptuneException`

Raalsky: > if uncaught, this exception can reach a user

Same for HTTPConflict :wink:. 

> let `HTTPConflict` fly from `HostedNeptuneBackend`

I'm not a big fan as the main idea was to overwrap the HttpConflict with something to hide the implementation details. HostedNeptuneBackend is only implementation and you would like to use part of that in the "upper layer".  I could imagine that for instance, OfflineNeptuneBackend could implement such logic and what it should raise? HttpConflict?

pkasprzyk: > Same for HTTPConflict
yep, but that's a legit exception, not our weird invention ;)

as to rest -- weeeell, you do have a point
`NeptuneObjectCreationConflict`? :)

pkasprzyk:
@Raalsky please fix linter :)

pkasprzyk (approved):

PR #853, opened 2022-02-25
Pc/ssl-error-message

Blaizzy:
Improved SSL error message

### DIFF
--- a/neptune/api_exceptions.py
+++ b/neptune/api_exceptions.py
@@ -47,7 +47,7 @@ Neptune client was unable to verify your SSL Certificate.
             - Debian >= 8
 
 {bold}What can I do?{end}
-You can manually configure Neptune to skip all SLL checks. To do that
+You can manually configure Neptune to skip all SSL checks. To do that
 set the NEPTUNE_ALLOW_SELF_SIGNED_CERTIFICATE environment variable to 'TRUE'.
 {bold}Note that might mean your connection is less secure{end}.
 
--- a/neptune/api_exceptions.py
+++ b/neptune/api_exceptions.py
@@ -29,7 +29,7 @@ class NeptuneSSLVerificationError(NeptuneException):
 {end}
 
 Neptune client was unable to verify your SSL Certificate.
- 
+
 {bold}What could go wrong?{end}
     - You are behind a proxy that inspects traffic to Neptune servers.
         - Contact your network administrator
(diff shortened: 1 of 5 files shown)


Blaizzy:
I know that some checks will fail because of the new naming. It needs to be updated in all places where SSLError is called.

Let me know if it is something I can do.

neptune/api_exceptions.py
```diff
@@ -21,12 +21,55 @@ class NeptuneApiException(NeptuneException):
...
+            {fail}"curl: (60) server certificate verification failed..."{end}
+    - Your machine software is not up-to-date.
+        - Minimal OS requirements:
+            - Windows >= XP SP3
+            - macOS >= 10.12.1
+            - Ubuntu >= 12.04
+            - Debian >= 8
+
+{bold}What can I do?{end}
+You can manually configure Neptune to skip all SLL checks. To do that
```
pkasprzyk, line 50: `SLL` -> `SSL`

pkasprzyk:
@Blaizzy a pro tip for the future: 
If you have test requirements installed (and you should ;) `pip install requirements/test_requirements.txt`) you can do: `pre-commit install`, which will install a git precommit hook for `black`. 
This way your code gets autoformatted each time you try to commit a change -- no more "trailing whitespace" or fomatting errors!

Blaizzy:
@pkasprzyk thank you very much!

I realized that while working at it. I found out about pre-commit and installed it locally as well as turn on the trim-trailing-whitespace feature on vscode.

I was getting an error every single time. It seems like the pre-commit used in the GH action has problems removing trailing whitespaces automatically.

neptune/api_exceptions.py
Blaizzy: Thanks, it is fixed ✅

pkasprzyk: it's not in the copy-pasted message of the second exception ;)

Blaizzy: Can you elaborate?

Blaizzy: No, it's not!

I did a complete search for SLL on all scripts just to double check.

Blaizzy: 😅 It was, but I fixed it ✅ 

vscode global search couldn't match 'SLL' the second file 😞

pkasprzyk (approved):

PR #866, opened 2022-03-16
Pass custom git information to init_run.

michalsustr:


michalsustr:
Hello, for our use case we'd like to pass git information in a custom way. Tagging @Raalsky for review. Thank you!

Raalsky:
Hey @michalsustr, I really appreciate your efforts :wink: Before jumping into a review I would like to tag @Blaizzy to check the expected form of this.

Blaizzy:
Hi @michalsustr

Thank you for your patience!

@Raalsky I left one comment and after that is fixed, it LGTM!

Herudaio:
Hey @michalsustr, thanks for the PR!

Unfortunately in this form, we don't want to add it to the code base:
1. It's one more parameter in the `init_run`, and we are very aggressively limiting adding new parameters there.
2. The other problem with GitInfo is that sometimes you want to resume a Run and update that information.  
The suggested solution here would be to enable updating `the "sys/git_info"` field, but that, unfortunately, requires some additional work on the server-side and is one of the reasons why we did not enable this option so far :(

Blaizzy:
Closing it as it was not accepted.

But this should not demotivate you to make future contributions because your effort was not in vain as it will go as a feature request that may or may not be implemented in the future. You can track upcoming features here: https://portal.neptune.ai/tabs/6-under-development

Thank you very much for your feedback and contribution!

PR #877, opened 2022-04-04
restart upload when file changes during

pkasprzyk:


neptune/new/internal/backends/hosted_file_operations.py
```diff
@@ -308,64 +310,72 @@ def _multichunk_upload(
...
-            if result.errors:
-                raise MetadataInconsistency(
-                    [err.errorDescription for err in result.errors]
-                )
-        return []
-    finally:
-        file_stream.close()
+    while True:
+        try:
+            file_stream = upload_entry.get_stream()
```
aniezurawski, line 315: Please, move this whole original body of "try" to new method and keep only while, try and single call here.

neptune/new/internal/backends/hosted_file_operations.py
```diff
@@ -308,64 +310,72 @@ def _multichunk_upload(
...
+                        .result
+                    )
+                    if result.errors:
+                        raise MetadataInconsistency(
+                            [err.errorDescription for err in result.errors]
+                        )
+                return []
+            finally:
+                file_stream.close()
+        except UploadedFileChanged as e:
```
aniezurawski, line 377: Let's retry in case of 500 http code as well.

neptune/internal/storage/datastream.py
```diff
@@ -65,6 +65,8 @@ def generate(self) -> Generator[FileChunk, Any, None]:
         while last_offset < self._total_size:
             chunk = self._fobj.read(chunk_size)
             if chunk:
+                if len(chunk) != min(chunk_size, self._total_size - last_offset):
```
aniezurawski, line 68: Do you think we can depend on mtime instead?

aniezurawski (changes requested):

neptune/new/internal/backends/hosted_file_operations.py
pkasprzyk: That doesn't sound good, tbh.

neptune/internal/storage/datastream.py
pkasprzyk: I think we could -- when I think about it, it probably better illustrates "changed during upload", even if it sometimes restarts unnecessarily.
Will do that.

neptune/new/internal/backends/hosted_file_operations.py
aniezurawski: 1. Why not? It can save Runs from crashing in case of random server errors. How about retrying specific http call instead of whole upload?
2. Can we abort old pending multipart upload in case of error (retrying whole upload) to save space?

pkasprzyk: Ad 2. sure we could, but server has currently not API for that. I'd prefer to classify this as "a future improvement" rather than "part of this fix".
Ad 1. I cringe when I hear "random server errors". There should be no such thing, and it certainly shouldn't be routinely handled in client-side code.

aniezurawski: - We already have had this discussion few months ago and we decided to retry 500 (this place in code was just missed back then). And after the time I can say it was probably good decision as we saved probably thousends of Runs from crash in case of temporary errors not handled properly in servers. It is very important thing for user experience. We should do what we can to save Runs from crashing.
- We should totally retry at least 502 and 503. Do we do now?
- What are actual cons of retrying 500 besides of you don't liking it? I can't see any if they are not too often.
- Do you actually believe it is possible to have bug-free system of this size?

pkasprzyk: You're right, we missed those API calls in our generic error handling, so this is not the place for this discussion. Fixed, but nevertheless:

> Do you actually believe it is possible to have bug-free system of this size?

I do not. But I strongly believe in failing early instead of going forward no matter what, which can cause invalid state, corrupted data etc. Bugs should be fixed, not hidden.

> What are actual cons of retrying 500 besides of you don't liking it?

500 can and often does mean that server encountered a situation or data batch so unexpected that it crashed. Retrying such request doesn't help, but creates an instant DOS script -- it just retries the request continuously, potentially crashing server instances faster that they recover.

aniezurawski: > I do not. But I strongly believe in failing early instead of going forward no matter what, which can cause invalid state, corrupted data etc. Bugs should be fixed, not hidden.
They are not hidden. They all are reported.

> I strongly believe in failing early instead of going forward no matter what, which can cause invalid state
We do on our side. But it is not necessary to bother our users.

> 500 can and often does mean that server encountered a situation or data batch so unexpected that it crashed. Retrying such request doesn't help, but creates an instant DOS script -- it just retries the request continuously, potentially crashing server instances faster that they recover.
It is not true if retries are rare.

neptune/new/internal/backends/hosted_file_operations.py
```diff
@@ -289,8 +289,12 @@ def _build_multipart_urlset(
...
     urlnameset = MULTIPART_URLS[target]
     return MultipartUrlSet(
-        start_chunked=getattr(swagger_client.api, urlnameset.start_chunked),
-        finish_chunked=getattr(swagger_client.api, urlnameset.finish_chunked),
+        start_chunked=with_api_exceptions_handler(
+            getattr(swagger_client.api, urlnameset.start_chunked)
+        ),
+        finish_chunked=with_api_exceptions_handler(
+            getattr(swagger_client.api, urlnameset.finish_chunked)
+        ),
```
aniezurawski, line 297: Is chunk upload call retried?

pkasprzyk: yes, it uses internally `upload_raw_data` function, which already had a `@with_api_exceptions_handler` annotation

aniezurawski (approved):

PR #878, opened 2022-04-04
pin kedro to <0.18 in e2e

pkasprzyk:


pkasprzyk:
aborted, fix for kedro integration incoming

PR #880, opened 2022-04-07
split execute_ops after delete preceeding file op

pkasprzyk:


neptune/new/internal/backends/operations_preprocessor.py
```diff
@@ -52,17 +52,26 @@
...
-                path_str, _OperationsAccumulator(op.path)
-            ).visit(op)
-        result = []
-        for acc in self._accumulators.values():
-            result.extend(acc.get_operations())
-        return result
+            acc: "_OperationsAccumulator" = self._process_op(op)
+            self.processed_ops_count += 1
+            # pylint: disable=protected-access
+            if acc.has_file_ops and len(acc._delete_ops) > 0:
```
aniezurawski, line 62: Since this class started to make some assumptions about in what order returned operations will be performed (returned order is not honored and this class started to be aware of it) following segragating code should be moved here from `execute_oeprations`:
```
upload_operations, artifact_operations, other_operations = [], [], []
for op in operations_preprocessor.get_operations():
    if isinstance(op, (UploadFile, UploadFileContent, UploadFileSet)):
        upload_operations.append(op)
    elif isinstance(op, TrackFilesToArtifact):
        artifact_operations.append(op)
    else:
        other_operations.append(op)
```
and it should have three methods instead of one `get_operations`.

neptune/new/internal/backends/operations_preprocessor.py
```diff
@@ -252,6 +267,7 @@ def visit_delete_attribute(self, op: DeleteAttribute) -> None:
                 self._modify_ops = []
                 self._config_ops = []
                 self._type = None
+                # does not change has_file_ops
```
aniezurawski, line 270: Why not?

neptune/new/internal/backends/operations_preprocessor.py
```diff
@@ -52,17 +52,26 @@
...
-            ).visit(op)
-        result = []
-        for acc in self._accumulators.values():
-            result.extend(acc.get_operations())
-        return result
+            acc: "_OperationsAccumulator" = self._process_op(op)
+            self.processed_ops_count += 1
+            # pylint: disable=protected-access
+            if acc.has_file_ops and len(acc._delete_ops) > 0:
+                self.processed_ops_count -= len(acc._modify_ops) + len(acc._config_ops)
```
aniezurawski, line 63: > `len(acc._modify_ops) + len(acc._config_ops)`

Let's have a method in accumulator for this.

neptune/new/internal/backends/operations_preprocessor.py
```diff
@@ -52,17 +52,26 @@
...
         for op in operations:
-            path_str = path_to_str(op.path)
-            self._accumulators.setdefault(
-                path_str, _OperationsAccumulator(op.path)
-            ).visit(op)
-        result = []
-        for acc in self._accumulators.values():
-            result.extend(acc.get_operations())
-        return result
+            acc: "_OperationsAccumulator" = self._process_op(op)
```
aniezurawski, line 59: Instead of maintaining `has_file_ops` it would be easier to just stop processing operations when file operation occur and there already exists any delete operation, wouldn't it?
Let's just throw exception in such case and handle it here instead of doing some reversing of already processed operations. What do you think?

aniezurawski (changes requested):

neptune/new/internal/backends/operations_preprocessor.py
pkasprzyk: I guess my initial idea was to encapsulate the decision process to the `_OperationsAccumulator` -- but it's leaking either way, so what you're proposing would probably result in a cleaner code. I'll try and restructure.

neptune/new/internal/backends/operations_preprocessor.py
pkasprzyk: sounds great, though I think I'd prefer one method return a dataclass/namedtuple of 3 groups -- retrieving preprocessed operations should still be a single method IMO

otherwise it becomes (at least theoretically) possible to e.g. retrieve upload ops and then pre-process some more operations, which would have weird semantics...

neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -503,7 +503,7 @@ def execute_operations(
             )
         )
 
-        return len(operations_batch) + dropped_operations, errors
+        return operations_preprocessor.processed_ops_count + dropped_operations, errors
```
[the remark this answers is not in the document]
pkasprzyk: @aniezurawski done in separate PR to keep it clean(er): https://github.com/neptune-ai/neptune-client/pull/891

neptune/new/internal/backends/operations_preprocessor.py
```diff
@@ -128,8 +176,10 @@ def _process_modify_op(
                 )
             )
         else:
+            self._check_prerequisites(op)
             self._type = expected_type
             self._modify_ops = modifier(self._modify_ops, op)
+            self.has_file_ops = self.has_file_ops or self._type.is_file_op()
```
aniezurawski, line 182: It is not used.

aniezurawski (changes requested):

neptune/new/internal/backends/operations_preprocessor.py
pkasprzyk: good catch, thanks

aniezurawski (approved):

PR #897, opened 2022-05-12
fix computing of a multipart upload chunk size

pkasprzyk:


neptune/internal/storage/datastream.py
```diff
@@ -53,12 +54,12 @@ def _get_chunk_size(self) -> int:
                 f"File {self._filename} is too big to upload:"
```
aniezurawski, line 54: It is not internal client error. It is explicitly unsupported.

aniezurawski (approved):

PR #899, opened 2022-05-16
Matching all listed tags instead of any when calling

### LINKED ISSUE #898: BUG: Inconsistency in `fetch_run_table` tag selector
Hi,

according to `fetch_run_table`'s documentation:

> Only experiments that have all specified tags will match this criterion.

Thus, the query aggregator for tags should be `AND`.

https://github.com/neptune-ai/neptune-client/blob/4b164bc470278edd160b7b9b5059a32c4822b376/neptune/new/metadata_containers/project.py#L235

Raalsky:
Closes #898

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## [UNRELEASED] neptune-client 0.16.3
 
 ## Fixes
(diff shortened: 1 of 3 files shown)


aniezurawski (approved):

aniezurawski (approved):

aniezurawski:
@wjaskowski No. Run matches if it contains all tags passed as a filter instead of any of them. No need to pass all tags that Run contains.

PR #900, opened 2022-05-18
Handle custom Neptune HTTP error types

aniezurawski:


neptune/new/internal/backends/hosted_client.py
```diff
@@ -22,6 +22,7 @@
 
 from neptune.new.exceptions import UnsupportedClientVersion
 from neptune.new.internal.backends.api_model import ClientConfig
+from neptune.new.internal.backends.swagger_client_wrapper import SwaggerClientWrapper
```
pkasprzyk, line 25: why don't wrap the one in `_get_token_client` (for consistency if nothing else)?

pkasprzyk (approved):

PR #903, opened 2022-05-19
replace click.echo with logger

pkasprzyk:


neptune/new/sync/status.py
```diff
@@ -82,43 +81,43 @@ def list_containers(
...
 
         if offline_dirs:
-            click.echo("Unsynchronized offline objects:")
+            logger.info("Unsynchronized offline objects:")
             for run_id in offline_dirs:
-                click.echo("- {}{}".format(OFFLINE_NAME_PREFIX, run_id))
-            click.echo()
-            click.echo(textwrap.fill(offline_run_explainer, width=90))
+                logger.info("- %s", f"{OFFLINE_NAME_PREFIX}{run_id}")
+            logger.info("")
```
Raalsky, line 101: It looks at least weird. Probably this should be just a newline.

neptune/new/sync/sync.py
```diff
@@ -41,6 +38,7 @@
 from neptune.new.internal.disk_queue import DiskQueue
 from neptune.new.internal.id_formats import QualifiedName, UniqueId
 from neptune.new.internal.operation import Operation
+from neptune.new.internal.utils.debug_log import logger
```
Raalsky, line 41: I think we should made a shortcut in `neptune.new.internal.utils.__init__` to be able to import it with `from neptune.new.internal.utils import logger` + `debug_log` looks weird as it was already used for run URL echo.

PR #909, opened 2022-05-24
fix handling of dead async processor

pkasprzyk:


neptune/new/internal/operation_processors/async_operation_processor.py
```diff
@@ -84,6 +84,8 @@ def enqueue_operation(self, op: Operation, wait: bool) -> None:
             self.wait()
 
     def wait(self):
+        if not self._consumer.is_running():
+            raise Exception("Synchronization job is not running, unable to synchronize")
```
aniezurawski: @normandy7 Some nice error message could be useful here.

The case is:
1. User starts a Run/Model/Version in ASYNC mode.
2. Asynchronous syncing thread dies from any reason (limits reached/connection errors/bugs/whatever).
3. Some error message with details about what caused async thread to die is presented printed.
4. User's script continue processing in "offline" mode with all the data being stored on local disk.
5. User calls `wait()` or `sync()`, but synchronization cannot be finished since syncing thread has been already dead for some time.

normandy7: Are these on the right track?

* If it happens in the middle of script execution, i.e. `wait()`/`sync()`: `The waiting/syncing was interrupted because the data synchronization process had already stopped.`
* If it happens on stop: `The synchronization process had already stopped before Neptune could finish the remaining operations.`

Are more solution suggestions or context needed?

aniezurawski: Yeah. They are on right track. I don't know if we use term "offline mode" anywhere in this context but it's kinda similar to being switched from async to offline mode. In real offline mode we continue without any warning. @Herudaio should we throw an exception in this case or just print a warning?

Herudaio: Disclaimer: I'm not sure I have the full context here.

If it behaves similar to offline mode, I would follow behavior from offline mode, so if I understand correctly warning would be more fitting.

aniezurawski: Full context is in first comment and it would be better if you feel you have it. ;)
In our engineering side of view it's similar to offline mode but I have no idea about user's side of view and how you usually communicate this situation to user. I don't want to introduce any inconsistencies in communication and I don't have full context here as well. :)

Herudaio: Re-read this with more coffee. We do describe it in docs as "switching to offline mode", so a warning would be good, but "as in offline mode" would be a guiding principle.

aniezurawski: 1. Something like "cannot finish synchronization data with server due to synchronization thread being dead" at the beginning would be nice.
2. Any suggestions for name of the exception? How about `NeptuneSynchronizationThreadDead`?

aniezurawski: @normandy7 @Herudaio What do you think?

normandy7: I'm not sure the user cares about the inner workings of the thread, specifically :) What's relevant to them is that the synchronization stopped at some point, but the logging continued offline. So they need to understand that they should look into two things:

1. They need to sync the offline data manually
1. Is there some reason (within their control) that the synching was interrupted?

If the above assumptions are correct, and one error is enough, how about something like this:

```
The synchronization process had stopped before Neptune could finish uploading the logged metadata.
Your data is stored locally, but you'll need to finish the synchronization manually.
To synchronize with the Neptune servers, enter the following on your command line:

    neptune sync

For details, see https://docs.neptune.ai/you-should-know/connection-modes#uploading-offline-data

If the synchronization fails, you may want to check your connection and ensure that you're
within limits by going to your Neptune project settings -> Usage.
If the issue persists, our support is happy to help.
```

normandy7: As for the exception name, I'd maybe rephrase the part that involves threads dying. Maybe `NeptuneSynchronizationAlreadyStopped`?

neptune/new/internal/operation_processors/async_operation_processor.py
```diff
@@ -159,6 +168,11 @@ def _wait_for_queue_empty(self, initial_queue_size: int, seconds: Optional[float
                 )
                 return
 
+            if not self._consumer.is_running():
+                exception = NeptuneSynchronizationAlreadyStoppedException()
+                logger.warning(str(exception))
+                return
```
shnela, line 174: Do we need to additionally print `NeptuneSynchronizationAlreadyStoppedException()` here?
We're running `flush` only from `wait`, where we already check consumer for running and raise exception if necessary.

neptune/new/internal/operation_processors/async_operation_processor.py
```diff
@@ -189,6 +203,14 @@ def __init__(
             self._batch_size = batch_size
             self._last_flush = 0
 
+        def run(self):
+            try:
+                super().run()
+            except Exception:
+                with self._processor._waiting_cond:
```
shnela, line 210: I guess it's the part responsible for faster dying when sync process is already dead, but I don't know how.
Some commet would be appreciated ;)

shnela (changes requested):

neptune/new/internal/operation_processors/async_operation_processor.py
aniezurawski: What does it have to do with flush? It's in `_wait_for_queue_empty` called on stop().

neptune/new/internal/operation_processors/async_operation_processor.py
aniezurawski: It just wakes waiting threads when synchronization process is dying. If not for this they would be sleeping forever waiting for synchronization. Nothing magic here.

neptune/new/internal/operation_processors/async_operation_processor.py
shnela: > What does it have to do with flush?

Eh, sorry, I've analyzed the code in github...

shnela (approved):

PR #917, opened 2022-05-30
Modify the messages users see for required and recommended neptune-client versions

Raalsky:


neptune/new/internal/backends/utils.py
```diff
@@ -183,24 +183,24 @@ def verify_client_version(client_config: ClientConfig, version: Version):
...
         )
     if (
         client_config.version_info.min_recommended
         and client_config.version_info.min_recommended > version
     ):
         click.echo(
-            "WARNING: We recommend an upgrade to a new version of neptune-client - {} (installed - {}).".format(
-                client_config.version_info.min_recommended, version
-            ),
+            f"WARNING: Your version of Neptune client library ({version}) is deprecated,"
```
aniezurawski, line 201: Missing header
```
{h1}
----NeptuneClientUpgradeRecommended-------------------------------------------------------------
{end}
````

**Review by aniezurawski — CHANGES_REQUESTED** [review]

File: neptune/new/internal/backends/utils.py (hunk 1 above)
Comment (normandy7) [inline neptune/new/internal/backends/utils.py:201]:
```suggestion
            f"WARNING: Your version of the Neptune client library ({version}) is deprecated,"
```

tests/neptune/new/internal/backends/test_hosted_neptune_backend.py
```diff
@@ -490,10 +490,10 @@ def test_min_compatible_version_fail(self, swagger_client_factory):
         self._get_swagger_client_mock(swagger_client_factory, min_compatible="0.5.14")
 
         # expect
-        with self.assertRaises(UnsupportedClientVersion) as ex:
+        with self.assertRaises(NeptuneClientUpgradeRequiredError) as ex:
             HostedNeptuneBackend(credentials)
 
-        self.assertTrue("Please install neptune-client>=0.5.14" in str(ex.exception))
+        self.assertTrue("minimal required version is >=0.5.14" in str(ex.exception))
```
normandy7, line 496:
```suggestion
        self.assertTrue("minimum required version is >=0.5.14" in str(ex.exception))
```

tests/neptune/new/internal/backends/test_hosted_neptune_backend.py
```diff
@@ -517,10 +517,10 @@ def test_max_compatible_version_fail(self, swagger_client_factory):
         self._get_swagger_client_mock(swagger_client_factory, max_compatible="0.4.999")
 
         # expect
-        with self.assertRaises(UnsupportedClientVersion) as ex:
+        with self.assertRaises(NeptuneClientUpgradeRequiredError) as ex:
             HostedNeptuneBackend(credentials)
 
-        self.assertTrue("Please install neptune-client==0.4.0" in str(ex.exception))
+        self.assertTrue("minimal required version is ==0.4.0" in str(ex.exception))
```
normandy7, line 523:
```suggestion
        self.assertTrue("minimum required version is ==0.4.0" in str(ex.exception))
```

neptune/new/exceptions.py
```diff
@@ -608,11 +608,15 @@ def __init__(
         )
         message = """
 {h1}
-----UnsupportedClientVersion-------------------------------------------------------------
+----NeptuneClientUpgradeRequiredError-------------------------------------------------------------
 {end}
-Your version of neptune-client ({current_version}) library is not supported by the Neptune server.
+Your version of Neptune client library ({current_version}) is no longer supported by the Neptune
+ server, minimal required version is {required_version}.
```
normandy7, line 614:
```suggestion
 server. The minimum required version is {required_version}.
```

neptune/new/exceptions.py
```diff
@@ -608,11 +608,15 @@ def __init__(
...
 
-Please install neptune-client{required_version}
+In order to update the Neptune client library in your terminal run:
```
normandy7, line 616:
```suggestion
In order to update the Neptune client library, run the following command in your terminal:
```

neptune/new/exceptions.py
```diff
@@ -608,11 +608,15 @@ def __init__(
...
+    {bash}pip install -U neptune-client{end}
+Or if you are using Conda run in your terminal instead:
```
normandy7, line 618:
```suggestion
Or if you are using Conda, run the following instead:
```

neptune/new/internal/backends/utils.py
normandy7: Requesting @Herudaio to comment further on this. Personally, I'm not yet sure where we should or should not have H1s.

normandy7: There seems to be no strong preference either way, so I think we should leave out the H1 for now.

PR #927, opened 2022-07-04
Added support of Service Accounts management

Raalsky:


neptune/management/internal/api.py
```diff
@@ -433,3 +437,140 @@ def get_workspace_member_list(name: str, api_token: Optional[str] = None) -> Dic
         }
     except HTTPNotFound as e:
         raise WorkspaceNotFound(workspace=name) from e
+
+
+@with_api_exceptions_handler
+def _get_raw_workspace_service_account_list(
+    name: str, api_token: Optional[str] = None
```
shnela, line 444: Can we keep workspace_name and project_name variables in this file?

neptune/management/internal/api.py
```diff
@@ -433,3 +437,140 @@ def get_workspace_member_list(name: str, api_token: Optional[str] = None) -> Dic
...
+    verify_type("api_token", api_token, (str, type(None)))
+
+    backend_client = _get_backend_client(api_token=api_token)
+    project_identifier = normalize_project_name(name=name, workspace=workspace)
+
+    params = {"projectIdentifier": project_identifier, **DEFAULT_REQUEST_KWARGS}
+
+    try:
+        result = backend_client.api.listProjectServiceAccounts(**params).response().result
+        return {
```
shnela, line 487: You return dict from `get_project_service_account_list` function. Change one of that :p

shnela (changes requested):

neptune/management/internal/api.py
Raalsky: Except for this function, the argument names are a part of product specification :wink:. Leave a comment there if you would like to discuss anything.

neptune/management/internal/api.py
Raalsky: Huh? This is `get_project_service_account_list`

shnela (approved):

shnela (approved):

PR #928, opened 2022-07-07
Added NEPTUNE_MODE env variables and support in all metadata containers

Raalsky:


neptune/new/envs.py
```diff
@@ -15,6 +15,8 @@
 #
 
 
+CONNECTION_MODE = "NEPTUNE_MODE"
```
Raalsky, line 18: @Herudaio What would you like to see? `NEPTUNE_MODE` or `NEPTUNE_CONNECTION_MODE` / something different?

Herudaio: I think `NEPTUNE_MODE` is a bit more in line with other env variables we have, but I'm not an expert in naming for those. @aniezurawski can you recommend anything here?

aniezurawski: No conventions here. Both are fine. Choose what you think will be more intuitive for users.

shnela (approved):

PR #930, opened 2022-07-07
E2E tests of SA admin API

shnela:


neptune/management/__init__.py
```diff
@@ -42,4 +42,8 @@
     "get_workspace_service_account_list",
     "MemberRole",
     "ProjectVisibility",
+    "get_workspace_service_account_list",
```
Raalsky, line 45: Those functions were already listed 2 lines above :wink:

neptune/management/internal/api.py
```diff
@@ -558,6 +560,8 @@ def remove_project_service_account(
 
     backend_client = _get_backend_client(api_token=api_token)
     project_identifier = normalize_project_name(name=name, workspace=workspace)
+    if workspace is None:
```
Raalsky, line 563: You can use `extract_project_and_workspace`

e2e_tests/management/test_management.py
```diff
@@ -155,3 +159,51 @@ def test_visibility_workspace(self, environment: "Environment"):
         delete_project(name=created_project_identifier, api_token=environment.admin_token)
 
         assert project_identifier not in get_project_list(api_token=environment.admin_token)
+
+    @staticmethod
+    def _create_project(environment: "Environment", *, name_suffix):
```
Raalsky, line 164: Why u need this instead of re-use of the project provided by environment?

neptune/management/internal/api.py
shnela: Actually fixed it when rebasing.

neptune/management/__init__.py
shnela: Right...

e2e_tests/management/test_management.py
shnela: Good enough

Raalsky (approved):

PR #931, opened 2022-07-08
Added support of lack of workspace in service accounts management api

Raalsky:


neptune/management/internal/utils.py
```diff
@@ -24,13 +24,13 @@
 from neptune.patterns import PROJECT_QUALIFIED_NAME_PATTERN
 
 
-def normalize_project_name(name: str, workspace: Optional[str] = None):
+def extract_project_and_workspace(name: str, workspace: Optional[str] = None):
```
shnela, line 27: Please use this function as well in `management.api` instead of `project_identifier.split("/")[0]` in two places.

shnela (approved):

PR #935, opened 2022-07-13
NPT-12012: add project key default generation

PK-Neptune:


neptune/management/internal/utils.py
```diff
@@ -52,3 +54,60 @@ def normalize_project_name(name: str, workspace: Optional[str] = None):
...
+        if unique_key is not None:
+            return unique_key
+
+        #  if above code do not find a proper id then add incrementation
+        unique_key_with_incrementation = self._find_unique_project_key_with_incrementation()
+        if unique_key_with_incrementation is not None:
+            return unique_key_with_incrementation
+
+        #  if above code do not find a proper id then add random string as suffix
+        while 1:
```
shnela, line 85: In python the way to go is `while True`, just python idiomatic thing ;)

neptune/management/internal/utils.py
```diff
@@ -52,3 +54,60 @@ def normalize_project_name(name: str, workspace: Optional[str] = None):
...
+    def _generate_random_suffix(self):
+        alphabet_copy = self._PROJECT_KEY_RANDOM_SUFFIX_ALPHABET.copy()
+        random.shuffle(alphabet_copy)
+        return "".join(alphabet_copy[: self._PROJECT_KEY_RANDOM_SUFFIX_SIZE])
+
+    def _find_unique_project_key_with_incrementation(self) -> Optional[str]:
+        for name_incrementation in range(
+            self._PROJECT_KEY_MIN_NUMBER_ITERATOR, self._PROJECT_KEY_MAX_NUMBER_ITERATOR + 1
+        ):
+            unique_key = self._find_unique_project_key(str(name_incrementation))
```
shnela, line 101: In fact we're sure that we should not check all possible prefixes in `_find_unique_project_key` since we did it before in `get_default_project_key`, but it's really small number of additional iteration on data in memory, so we don't care...

requirements.txt
```diff
@@ -17,3 +17,4 @@ dataclasses>=0.6; python_version<"3.7"
 swagger-spec-validator>=2.7.4
 psutil
 jsonschema<4.0.0
+retry<1.0.0
```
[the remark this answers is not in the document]
shnela: If there is any important reason for compability, it's worth mentioning here in comment, so we could check if problem exists in the future when we'd like to upgrade requirements.

neptune/management/internal/utils.py
```diff
@@ -52,3 +54,60 @@ def normalize_project_name(name: str, workspace: Optional[str] = None):
...
+            project_name_substring = (
+                self.project_name[: self.max_substring_length] + self._generate_random_suffix()
+            )
+            if project_name_substring not in self.existing_project_keys:
+                return project_name_substring
+
+    def _generate_random_suffix(self):
+        alphabet_copy = self._PROJECT_KEY_RANDOM_SUFFIX_ALPHABET.copy()
+        random.shuffle(alphabet_copy)
+        return "".join(alphabet_copy[: self._PROJECT_KEY_RANDOM_SUFFIX_SIZE])
```
shnela, line 95: We could simply use [random.choices](https://docs.python.org/3.8/library/random.html#random.choices) instead of playing directly with `_PROJECT_KEY_RANDOM_SUFFIX_ALPHABET`.
It'll be faster (I know that it's stupid argument in Python, especially in such case), but additionally the code would be more succint.

shnela: ofc you can leave `_generate_random_suffix` function for readability, but its content will be one line.

neptune/management/internal/utils.py
```diff
@@ -52,3 +54,60 @@ def normalize_project_name(name: str, workspace: Optional[str] = None):
     )
 
     return f"{extracted_workspace_name}/{extracted_project_name}"
+
+
+class ProjectKeyGenerator:
```
shnela, line 59: I think we could add  link to [docs](https://docs.neptune.ai/api-reference/management#.create_project) with information about key format.

tests/neptune/management/internal/test_utils.py
```diff
@@ -42,3 +47,35 @@ def test_normalize_project_name(self):
 
         with self.assertRaises(ConflictingWorkspaceName):
             normalize_project_name(name="jackie/sandbox", workspace="john")
+
+
+@pytest.mark.parametrize(
```
shnela, line 52: According to docs, the key should be uppercase, so lowercase input is perfect test data, but I'd expect upercase result.

shnela (changes requested):

requirements.txt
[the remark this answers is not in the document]
PK-Neptune: current version of retry is 0.9.2 - https://pypi.org/project/retry/   
I want to avoid situation when they release version 1+ with broken backward compatibility. It is the reason why such requirement was added.

neptune/management/internal/utils.py
PK-Neptune: clarified on call

neptune/management/internal/utils.py
PK-Neptune: corrected

neptune/management/internal/utils.py
PK-Neptune: corrected

neptune/management/internal/utils.py
PK-Neptune: I added some comments, check it

tests/neptune/management/internal/test_utils.py
PK-Neptune: Good point, fixed

requirements.txt
```diff
@@ -17,3 +17,4 @@ dataclasses>=0.6; python_version<"3.7"
 swagger-spec-validator>=2.7.4
 psutil
 jsonschema<4.0.0
+retry>=0.9.2, == 0.9.*
```
shnela, line 20: Why not just `retry~=0.9.2`?

shnela (approved):

PR #941, opened 2022-07-20
More methods called via user or sa token in e2e tests

Raalsky:


e2e_tests/management/test_management.py
```diff
@@ -66,41 +76,64 @@ def test_standard_scenario(self, environment: Environment):
...
-                name=created_project_identifier, api_token=environment.admin_token
-            ).get(environment.user)
-            == "contributor"
+        assert environment.user in project_members
+        assert project_members.get(environment.user) == "contributor"
+
+        project_members = get_project_member_list(
+            name=created_project_identifier, api_token=environment.user_token
         )
+        assert environment.user in project_members
```
shnela, line 110: We could also check:
```python
assert environment.service_account not in project_members
```

shnela (dismissed):

alpha_integration_dev/tmp.py
```diff
@@ -0,0 +1,11 @@
+from neptune.management import add_project_member
```
Raalsky, line 1: ?

shnela: Obviously added by accident...

shnela (approved):

PR #945, opened 2022-07-21
Initial HuggingFace E2E tests - proper run lifecycle

Raalsky:


requirements/e2e_requirements.txt
```diff
@@ -7,3 +7,4 @@ pytest-tap
 pytorch-lightning
 torchvision
 sklearn
+git+https://github.com/neptune-ai/transformers.git@rj/proper-run-lifecycle#egg=transformers
```
shnela, line 10: Could you add comment that we should replace it once we finish developing HF integration?

e2e_tests/integrations/test_huggingface.py
```diff
@@ -0,0 +1,163 @@
...
+            "train_dataset": train_dataset,
+            "eval_dataset": validation_dataset,
+        }
+
+    def test_every_train_should_create_new_run(self, environment):
+        common_tag = fake.nic_handle()
+        project = get_project(name=environment.project, api_token=environment.user_token)
+
+        trainer = Trainer(
+            **self._trainer_default_attributes,
```
shnela, line 101: I don't get idea of adding `**` in prefix of function, have you forgotten `()` to call it?

e2e_tests/integrations/test_huggingface.py
```diff
@@ -0,0 +1,163 @@
...
+            trainer = Trainer(
+                **self._trainer_default_attributes,
+                callbacks=[
+                    NeptuneCallback(api_token=environment.user_token, project=environment.project)
+                ]
+            )
+            trainer.train()
+            del trainer
+
+        assert with_neptune_callback() / standard() <= EXPECTED_OVERWHELMING_FACTOR
```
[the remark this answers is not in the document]
shnela: And I think that `MAX_OVERWHELMING_FACTOR` would be better name.

shnela (changes requested):

e2e_tests/integrations/test_huggingface.py
Raalsky: `_trainer_default_attributes` is property so `self._trainer_default_attributes` is proper call (output is dictionary of Trainer attributes)

e2e_tests/integrations/test_huggingface.py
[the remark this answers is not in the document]
Raalsky: I'll test it, after all, as every feature will impact it :wink:

e2e_tests/integrations/test_huggingface.py
shnela: Right...

shnela (approved):

PR #946, opened 2022-07-21
Pk/npt 12012

PK-Neptune:


PK-Neptune:
All E2E tests are passed. https://github.com/neptune-ai/neptune-client/actions/runs/2717721931

neptune/management/internal/api.py
```diff
@@ -165,33 +181,51 @@ def create_project(
...
     try:
-        response = backend_client.api.createProject(**params).response()
-        return normalize_project_name(
-            name=response.result.name, workspace=response.result.organizationName
+        workspaces = (
+            backend_client.api.listOrganizations(**DEFAULT_REQUEST_KWARGS).response().result
         )
+        workspace_name_to_id = {f"{f.name}": f.id for f in workspaces}
+    except HTTPNotFound:
+        raise WorkspaceNotFound(workspace=workspace)
```
aniezurawski, line 221: Looks like this `workspace` argument was not even used. And I don't see any possibilities of 404 being thrown in this endpoint.

tests/neptune/management/internal/test_utils.py
```diff
@@ -42,3 +47,51 @@ def test_normalize_project_name(self):
...
+        (("ccc", {"aaa", "bbb", "ccc", "ccc2", "ccc3"}), "CCC"),
+        (("aaaa", {"aaa", "aaa2", "aaa3", "aaaa", "aaaa2", "aaaa3", "bbb", "ccc"}), "AAAA"),
+        (("A----", {"AAA", "BBB", "CCC", "CCC2", "CCC3"}), "A"),
+        (("----", {"AAA", "BBB", "CCC", "CCC2", "CCC3"}), ""),
+        (("a-a", {"AAA", "BBB", "CCC", "CCC2", "CCC3"}), "AA"),
+        (("A-A", {"AAA", "BBB", "CCC", "CCC2", "CCC3"}), "AA"),
+        (("-0--1-----", {"AAA", "BBB", "CCC"}), "01"),
+        (("-000--1--", {"AAA", "BBB", "CCC"}), "0001"),
+        (("-000--111", {"AAA", "BBB", "CCC"}), "00011"),
+        (("-0--!@#$%^&*()_+=-[]{};':,./<>?|--1-", {"AAA", "BBB", "CCC"}), "01"),  # not the case now
```
aniezurawski, line 89: Key using only digits are invalid. So probably something like `01XYZ` (where XYZ are random) should be generated.

aniezurawski (changes requested):

tests/neptune/management/internal/test_utils.py
PK-Neptune: Yes, it is done like this. The string on the right side is the prefix of the result.

neptune/management/internal/api.py
PK-Neptune: I just moved this code to separate method, I can do more cleanup if you want

aniezurawski: I know. But it's good to do some small cleanups when possible.

PK-Neptune: Cleaned

neptune/management/internal/api.py
```diff
@@ -110,11 +110,13 @@ def get_project_list(api_token: Optional[str] = None) -> List[str]:
     ]
 
 
-def _get_projects_keys_in_organization(workspace: str, api_token: Optional[str] = None) -> Set[str]:
+def _get_projects_keys_in_organization(
+    workspace_id: str, api_token: Optional[str] = None
```
aniezurawski, line 114: I think we can use name instead of id in this API call and avoid listing organizations.
But in this case it will be actually necessary to catch `HTTPNotFound`.

PK-Neptune: retrieval of id was already done in the workspace name validation, so I used it. Then we don't have to secure our code from HTTPNotFound. Ok it is possible, but what is probability of this situation? Moreover workspace_id is used to create new project

PK-Neptune:
all e2e passed after changes: https://github.com/neptune-ai/neptune-client/actions/runs/2732962025

aniezurawski (approved):

PR #947, opened 2022-07-21
Added tests for logging metadata

Raalsky:


e2e_tests/integrations/test_huggingface.py
```diff
@@ -161,3 +167,51 @@ def test_run_access_without_callback_configured(self):
 
         with pytest.raises(Exception):
             NeptuneCallback.get_run(trainer)
+
+    def test_log_parameters_indicator(self, environment):
+        with init_run(project=environment.project, api_token=environment.user_token) as run:
+            base_namespace = "custom/base/path"
```
shnela, line 173: What is `base_namespace` for, we have separete test for it in `test_log_with_custom_base_namespace`.

shnela (changes requested):
I don't see check for `eval` namespace, nor `args` and `model.config` log tests.

e2e_tests/integrations/test_huggingface.py
```diff
@@ -161,3 +167,51 @@ def test_run_access_without_callback_configured(self):
...
+            callback = NeptuneCallback(run=run, base_namespace=base_namespace)
+            trainer = Trainer(**self._trainer_default_attributes, callbacks=[callback])
+            trainer.train()
+            run.sync()
+
+            assert run.exists(f"{base_namespace}/trainer_parameters")
+            assert run.exists(f"{base_namespace}/model_parameters")
```
shnela, line 180: Don't we want to check if logged values are valid?

e2e_tests/integrations/test_huggingface.py
Raalsky: In `test_log_with_custom_base_namespace` we're testing that `log(...)` (a user-exposed) method applies the `base_namespace` as a prefix. In this case we're testing the common parameters (an internal only). It's something really different as the correctness of the preceding doesn't impact the following :wink:

e2e_tests/integrations/test_huggingface.py
Raalsky: Ok. Will try

Raalsky: I've added a few simple one that we know they exists + In order to use a structure of `config` I've added proper model config that was introduced here: https://github.com/neptune-ai/neptune-client/pull/954

shnela (approved):

PR #948, opened 2022-07-22
Added support of importing HuggingFace

Raalsky:


neptune/new/integrations/transformers/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+try:
+    # pylint: disable=import-error
+    from transformers.integrations import NeptuneCallback
```
shnela, line 19: I think we've decided to use wildcard when importing integration from external library in order to be more error-proof after big changes in ingration libraries.

shnela (approved):
I've checked only two last commits as rest looks like non rebased changes.

neptune/new/integrations/transformers/__init__.py
Raalsky: But this module has all the integrations (W&B, Tensorboard etc.) :wink:. We could use a wildcard in case of specialized modules like Pytorch Lightning etc.

PR #949, opened 2022-07-22
SA docstrings and errors

normandy7:


neptune/management/exceptions.py
```diff
@@ -85,14 +85,15 @@ class UserNotExistsOrWithoutAccess(ManagementOperationFailure):
     code = 9
     description = (
         'User "{user}" does not exist or has no access to project "{project}". '
-        "If the project visibility is set to workspace, a user cannot be added or removed."
+        "If the project visibility is set to 'workspace', a user cannot be added or removed."
```
normandy7, line 88: Is the wording "added or removed" correct? As far as I can tell, this exception is raised only when you try to remove a user from a project. Could you confirm whether the removal will fail in the following scenarios:

1. The user was not found among the members of a private project
2. The project's visibility is "workspace" or "public"

normandy7, line 88: Is the wording "added or removed" correct? As far as I can tell, this exception is raised only when you try to remove a user from a project.

Could you confirm whether the removal will fail in the following scenarios:

1. The user was not found among the members of a private project
1. The project's visibility is "workspace" or "public"

neptune/management/internal/api.py
```diff
@@ -141,37 +143,43 @@ def create_project(
...
-        name(str): The name of the project in Neptune in the format 'WORKSPACE/PROJECT'.
-            If workspace argument was set, it should only contain 'PROJECT' instead of 'WORKSPACE/PROJECT'.
-        key(str): Project identifier. It has to be contain 1-10 upper case letters or numbers.
-            For example, 'GOOD5'
-        workspace(str, optional): Name of your Neptune workspace.
-            If you specify it, change the format of the name argument to 'PROJECT' instead of 'WORKSPACE/PROJECT'.
-            If 'None' it will be parsed from the `name` argument.
-        visibility(str, optional): level of visibility you want your project to have.
+        name: The name of the project in Neptune in the form 'workspace-name/project-name'.
+            If workspace argument was set, it should only contain 'project' instead of 'workspace-name/project-name'.
```
aniezurawski, line 150: Should be 'project-name' instead of 'project', right?

neptune/management/internal/api.py
```diff
@@ -81,21 +81,23 @@ def _get_backend_client(api_token: Optional[str] = None) -> SwaggerClientWrapper
...
+
+    Example:
         >>> from neptune import management
         >>> management.get_project_list()
-    You may also want to check `management API reference`_.
-    .. _management API reference:
-       https://docs.neptune.ai/api-reference/management
+
+    You may also want to check the management API reference:
+    https://docs.neptune.ai/api-reference/management
```
PK-Neptune, line 100: maybe better link would be 'https://docs.neptune.ai/api-reference/management#.get_project_list' ??  
What do you think?

normandy7: You're right that many of these links could be more specific. The reason I slapped the same link everywhere is that the URLs will change with the release of the new docs, so I allowed myself to cut this corner...

neptune/management/internal/api.py
normandy7: Yep, think it's a leftover from some find/replace change. Thanks for spotting

Raalsky (approved):

neptune/management/exceptions.py
Raalsky: User removal in all visibility scenario will raise `UserNotExistsOrWithoutAccess`
In case of adding a user to the project, it will pass in with all visibilities (in case of "workspace" visibility it will not be listed after all but will be stored internally and will be visible after switching to private visibility).

Raalsky: Currently adding users to workspace visibility is some kind of unexpected behavior that we've introduced due to easier implementation. Could be treated as a bug.

normandy7: If I understand your reply correctly, this exception would not be raised when you add a user to a project with `visibility="workspace"`, correct? Just trying to understand if the highlighted line is accurate or not :)

Raalsky: Yes, It won't be raised (currently, probably this will be updated in the future)

normandy7: Okay. There are some other potential clarifications that could be made to project visibility documentation, so I'm going to resolve this in the context of this PR.

PR #952, opened 2022-07-25
Added tests of hyperparameter search

Raalsky:


e2e_tests/integrations/test_huggingface.py
```diff
@@ -383,3 +383,37 @@ def test_non_monitoring_runs_creation_with_initial_run(self, environment):
         )
         assert len(runs) == 2
         assert runs[1].get_attribute_value("finetuning/train/metric3") == 345
+
+    def test_hyperparameter_optimization(self, environment):
+        # given
+        n_trials = 5
+        common_tag = fake.nic_handle()
+        project = get_project(name=environment.project, api_token=environment.user_token)
```
shnela, line 391: ` common_tag` and `project` initialization became so popular in all tests, that we could prepare fixtures for them and receive them as method parameters.

shnela (approved):

e2e_tests/integrations/test_huggingface.py
Raalsky: Ok. I'll try after all of the MRs :D

PR #953, opened 2022-07-26
Tests of different usages

Raalsky:


e2e_tests/integrations/test_huggingface.py
```diff
@@ -417,3 +417,44 @@ def model_init():
         for run_id, run in enumerate(runs):
             assert run.get_attribute_value("finetuning/trial") == f"trial_{run_id}"
             assert run.get_attribute_value("monitoring/cpu") is not None
+
+    def test_usages(self, environment):
+        # given
+        trainer_args = self._trainer_default_attributes.copy()
```
shnela, line 423: I think that copy is note required here, but we can leave it ;)

shnela (approved):

PR #955, opened 2022-07-27
Full name of service account instead of short name

Raalsky:


neptune/management/internal/api.py
```diff
@@ -592,9 +598,12 @@ def remove_project_service_account(
...
 
-    service_account = _get_raw_workspace_service_account_list(
-        workspace_name=workspace, api_token=api_token
-    ).get(service_account_name)
+    try:
+        service_account = _get_raw_workspace_service_account_list(
+            workspace_name=workspace, api_token=api_token
+        )[service_account_name]
+    except KeyError as e:
+        raise ServiceAccountNotFound(name=service_account_name) from e
```
shnela, line 606: I'd add here information about workspace since it'll be much easier to identify why SA was not found.

Raalsky: Done

shnela (approved):

PR #960, opened 2022-08-03
Runtime plotly matplotlib requirements

Raalsky:


requirements/test_requirements.txt
```diff
@@ -8,7 +8,7 @@ pytest-cov
 pytest-xdist
 tensorflow
 pytest-timeout
-matplotlib
+matplotlib<3.5.0
```
aniezurawski, line 11: How about `matplotlib!=3.5.*` for now?

normandy7 (dismissed):
Adjusted the messages a bit. Let me know if you think more context or help would be beneficial to the user.

requirements/test_requirements.txt
Raalsky: `matplotlib==3.6.x` will have even more deprecated methods :D. It's about `plotly` that needs an old version so I don't think it will be proper. + This requirements are only for testing purposes.

neptune/new/internal/utils/images.py
```diff
@@ -170,7 +171,7 @@ def _matplotlib_to_plotly(chart):
                 raise PlotlyIncompatibilityException(
                     matplotlib_version,
                     plotly_version,
-                    "Due to plotly using some deprecated matplotlib methods, we recommend downgrading matplotlib"
+                    "Due to Plotly using some deprecated Matplotlib methods, we recommend downgrading Matplotlib"
```
aniezurawski, line 174: @normandy7 I think it's more common to use lower letters only in names of python packages.

aniezurawski (approved):

neptune/new/internal/utils/images.py
normandy7: Right, I guess these refer literally to the Python packages. Will revert

PR #973, opened 2022-08-09
Added tests for model checkpoints including volatile

Raalsky:


e2e_tests/integrations/test_huggingface.py
```diff
@@ -461,3 +463,173 @@ def test_usages(self):
         assert "NeptuneCallback" in [
             type(callback).__name__ for callback in trainer.callback_handler.callbacks
         ]
+
+    def test_model_checkpoints_same(self, environment):
+        with init_run(project=environment.project, api_token=environment.user_token) as run:
+            run_id = run["sys/id"].fetch()
+            callback = NeptuneCallback(run=run, log_checkpoints="same")
+            training_args = self._trainer_default_attributes.copy()
```
shnela, line 471: It's inconsistent with fixed tests version where we don't use `copy()` in `self._trainer_default_attributes.copy()`.

Here and in several other places.

e2e_tests/integrations/test_huggingface.py
```diff
@@ -461,3 +463,173 @@ def test_usages(self):
...
+            trainer = Trainer(**training_args, callbacks=[callback])
+            trainer.train(resume_from_checkpoint="model/checkpoint-2000")
+
+        run = init_run(
+            run=run_id,
+            project=environment.project,
+            api_token=environment.user_token,
+            mode="read-only",
+        )
+        assert run["finetuning/train/epoch"].fetch_last() == 1000
```
shnela, line 513: I think that we should use `_test_with_run_initialization` twice in this test.

shnela: The same is true for other tests.

shnela (changes requested):

e2e_tests/integrations/test_huggingface.py
Raalsky: Done

e2e_tests/integrations/test_huggingface.py
Raalsky: Done

shnela (approved):

shnela (approved):

PR #977, opened 2022-08-12
Added dev package install to e2e requirements

Raalsky:


.github/workflows/full.yml
```diff
@@ -60,6 +60,10 @@ jobs:
         run: |
           pip install -r requirements.txt -r requirements/e2e_requirements.txt
 
+      - name: List dependencies
```
shnela, line 63: Isn't it some debug leftover?

Raalsky: Yes but I would like to add it for future :wink:

shnela (approved):

PR #978, opened 2022-08-12
Include prophet integration

twolodzko:
I am not changing the changelog, as we are not making it official yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## [UNRELEASED] neptune-client 0.16.6
 
 ## Features
-- Added support for Prophet integration... ([#978](https://github.com/neptune-ai/neptune-client/pull/978))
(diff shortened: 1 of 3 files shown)


Raalsky:
About CHANGELOG, I think it's okay to update it :wink:. It looks like Prophet integration is in almost good condition so the release will be in one of the upcoming weeks + we do not have anything much bigger than this to release.

Raalsky (approved):

twolodzko:
> About CHANGELOG, I think it's okay to update it wink. It looks like Prophet integration is in almost good condition so the release will be in one of the upcoming weeks + we do not have anything much bigger than this to release.

@Raalsky  Do we want to show in changelog an integration with failing (due to misconfiguration) tests?

twolodzko:
@Raalsky  done.

Raalsky (approved):

PR #984, opened 2022-08-23
Log argparse.Namespace objects as dicts

shnela:


neptune/new/metadata_containers/metadata_container.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
```
Raalsky, line 14: Remember to update CHANGELOG :wink:

PR #985, opened 2022-08-24
Add new `with_id` paramter to init functions

shnela:


CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
-## [UNRELEASED] neptune-client 0.16.6
+## [UNRELEASED] neptune-client 0.17.0
```
aniezurawski, line 1: Does it break compatibility? Btw. I would like to release PK's MR before increasing minor to 17.

neptune/new/internal/init/model.py
```diff
@@ -43,9 +44,12 @@
 from neptune.new.metadata_containers import Model
 from neptune.new.types.mode import Mode
 
+_logger = logging.getLogger(__name__)
```
aniezurawski, line 47: We should use logger from `neptune.new.internal.utils.logger` created by Kasprzyk, shouldn't we?

neptune/new/internal/init/model.py
```diff
@@ -55,7 +59,17 @@ def init_model(
...
     proxies: Optional[dict] = None,
 ) -> Model:
-    verify_type("model", model, (str, type(None)))
+    if model:
+        if with_id:
+            raise NeptuneInitParametersCollision("with_id", "model", method_name="init_model")
+
+        _logger.warning(
+            "parameter `model` is deprecated, use `with_id` instead."
+            " We'll end support of it in `neptune-client==1.0.0`."
```
aniezurawski, line 68: Do we need to do it like this? Is't there some cool standard decorator?

aniezurawski (changes requested):

CHANGELOG.md
shnela: No it doesn't, but according to semver and [mycek](https://neptune-labs.slack.com/archives/C01H1A8DA3X/p1661849643485799) it should be minor.

neptune/new/internal/init/model.py
shnela: I think logger created by Kasprzyk is just responsible logging captured stdout/stderr to neptune.
Patter used by my is used in other parts of the code as well.

neptune/new/internal/init/model.py
shnela: There are some tools to mark some "things" as deprecated, but we don't use them. There's still issue for that in Jira.
Here I've used just ad-hoc solution, choosing some library will require substantial research on technlogy which'll suit best our deprecation use-cases (mostly imports).

CHANGELOG.md
aniezurawski: Let's increase PATCH in this case. These rules does not apply to version 0.

neptune/new/internal/init/model.py
aniezurawski: > I think logger created by Kasprzyk is just responsible logging captured stdout/stderr to neptune.
I don't understand what you mean. We should capture all logs. What is the difference?

> Patter used by my is used in other parts of the code as well.
Yeah, because Piotrek replaced only usages of `click` for now. Still our goal should be to be consistent in whole code base.

neptune/new/internal/init/model.py
aniezurawski: > but we don't use them

Maybe it's good moment to start. Now there is a lot of awful code duplication. If you prefer, you can create some very simple decorator by yourself. But I belive there is some simple lib for such things.

CHANGELOG.md
shnela: ok

neptune/new/internal/init/model.py
shnela: Ok, let's create simple decorator for now.

neptune/new/exceptions.py
```diff
@@ -574,21 +574,26 @@ class NeptuneWrongInitParametersException(NeptuneException):
...
         self.parameter1 = parameter1
         self.parameter2 = parameter2
+        self.method_name = method_name
         message = """
 {h1}
-----NeptuneRunResumeAndCustomIdCollision-----------------------------------------
+----NeptuneInitParametersCollision-----------------------------------------
 {end}
-The {python}{parameter1}{end} and {python}{parameter2}{end} parameters of the {python}init(){end} method are mutually exclusive.
+The {python}{parameter1}{end} and {python}{parameter2}{end} parameters of the {python}{method_name}(){end} method are mutually exclusive.
```
aniezurawski, line 585: Exception still has `Init` in its name. Maybe we should have `NeptuneInitParametersCollision` inheritating from `NeptuneParametersCollision` to not break backward compatibility and still have explanatory name.

CHANGELOG.md
```diff
@@ -41,6 +41,8 @@
 - Fix computing of a multipart upload chunk size ([#897](https://github.com/neptune-ai/neptune-client/pull/897))
 - Matching all listed tags instead of any when calling `fetch_runs_table` ([#899](https://github.com/neptune-ai/neptune-client/pull/899))
 - Fix invalid processing of delete followed by file upload in a single batch ([#880](https://github.com/neptune-ai/neptune-client/pull/880))
+- Stop hanging indefinitely on wait when async data synchronization process is dead ([#909](https://github.com/neptune-ai/neptune-client/pull/909))
+- Finish stop() faster when async data synchronization process dies ([#909](https://github.com/neptune-ai/neptune-client/pull/909))
```
aniezurawski, line 45: Something went wrong during mergeing PK's MR. Could you move it to 0.16.7 since it has not been released yet?

neptune/new/internal/utils/deprecation.py
```diff
@@ -0,0 +1,46 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from functools import wraps
+
+from neptune.new.exceptions import NeptuneInitParametersCollision
+from neptune.new.internal.utils.logger import logger
+
+
+def deprecated_id_parameter(*, deprecated_kwarg_name):
```
aniezurawski, line 22: How about having generic method and passing argument name?

neptune/new/internal/utils/deprecation.py
```diff
@@ -0,0 +1,46 @@
...
+            deprecated_param = kwargs.get(deprecated_kwarg_name)
+            with_id = kwargs.get("with_id")
+            if deprecated_param is not None:
+                if with_id is not None:
+                    raise NeptuneInitParametersCollision(
+                        "with_id", deprecated_kwarg_name, method_name=f.__name__
+                    )
+
+                logger.warning(
+                    f"parameter `{deprecated_kwarg_name}` is deprecated, use `with_id` instead."
```
aniezurawski: It's called argument, not parameter, isn't it?

neptune/new/internal/init/model.py
```diff
@@ -59,16 +56,6 @@ def init_model(
...
 ) -> Model:
-    if model:
-        if with_id:
-            raise NeptuneInitParametersCollision("with_id", "model", method_name="init_model")
-
-        _logger.warning(
-            "parameter `model` is deprecated, use `with_id` instead."
-            " We'll end support of it in `neptune-client==1.0.0`."
-        )
-        with_id = model
```
aniezurawski: I think something like `with_id = with_id or model` is still necessary.

aniezurawski (changes requested):

neptune/new/exceptions.py
shnela: I just wanted to have more generic method name, but only for `init_run`, `init_model` and so on.
I don't think that `Init` in exception name could be a problem.
I don't even expect other places where we could have problems with such backward compatybility.

CHANGELOG.md
shnela: Ok, I'll fix that.

shnela: Even 0.16.7 after next rebase

neptune/new/internal/utils/deprecation.py
shnela: Ok

neptune/new/internal/utils/deprecation.py
shnela: Why do you think parameter is not ok in this scenario?

I think we're marking parameter of the function as deprecated, so user trying to pass it as an argument can't do that.

For me two options are ok, but parameter is better.

neptune/new/internal/init/model.py
shnela: Decorator is responsible for passing deprecated kwarg as `with_id` to function.

neptune/new/internal/utils/deprecation.py
aniezurawski: >Why do you think parameter is not ok in this scenario?

I don't know what scenario has to do with this.
Isn't it officially commonly called **arguments** in all programming languages including python?

aniezurawski: Sorry, my mistake.
declaration => parameter
value pass to function call => argument

aniezurawski (approved):

aniezurawski (approved):

PR #986, opened 2022-08-30
Introduce filtering columns when fetching tables

shnela:
…ions tables


aniezurawski (dismissed):

aniezurawski:
Let's mark it as Draft so it's not merged by mistake before release of backend component.

aniezurawski (approved):

PR #987, opened 2022-08-30
E2E tests of filtering columns

shnela:


e2e_tests/standard/test_base.py
```diff
@@ -248,3 +248,44 @@ def test_fetch_model_versions_table(self, container: Model, environment):
...
+        assert columns_none[0].get_attribute_value(key1) == value
+        assert columns_none[0].get_attribute_value(key2) == value
+
+        columns_empty = project.fetch_runs_table(tag=tag, columns=[]).to_rows()
+        assert len(columns_empty) == 1
+        assert columns_empty[0].get_attribute_value("sys/id") == sys_id
+        assert columns_empty[0].get_attribute_value(key1) is None
+        assert columns_empty[0].get_attribute_value(key2) is None
+
+        columns_with_one_key = project.fetch_runs_table(tag=tag, columns=[key1]).to_rows()
```
aniezurawski, line 287: Let's test models and versions as well.

shnela: Ok, tomorrow.

e2e_tests/standard/test_base.py
```diff
@@ -175,19 +175,15 @@ def test_add_and_remove_tags(self, container: MetadataContainer):
...
 class TestFetchTable(BaseE2ETest):
-    def test_fetch_runs_table(self, environment):
+    def test_fetch_runs_by_tag(self, environment):
         tag1, tag2 = str(uuid.uuid4()), str(uuid.uuid4())
 
         with neptune.init_run(project=environment.project) as run:
             run["sys/tags"].add(tag1)
             run["sys/tags"].add(tag2)
-            run["value"] = 12
-            run.sync()
```
aniezurawski, line 185: Why have you removed it?

e2e_tests/standard/test_base.py
```diff
@@ -199,36 +195,9 @@ def test_fetch_runs_table(self, environment):
             key=lambda r: r.get_attribute_value("sys/id"),
         )
         assert len(runs_table) == 1
```
aniezurawski, line 197: Would be nice to check at least id.

e2e_tests/standard/test_base.py
shnela: Because `stop()` is called automatically when we leave context manager block.

e2e_tests/standard/test_base.py
shnela: Why? `_id` is some internal detail, isn't correct `sys/id` enough?

aniezurawski: This is what I meant. But right now only number of runs is asserted.

e2e_tests/standard/test_base.py
aniezurawski: What about testing value od an attribute?

shnela: This is `test_fetch_runs_by_tag` I wanted to check fetching by id only.
Testing of attribute values for runs is done in `test_fetch_runs_table`.

I think it's good to minimalize complexity of tests and make them focused on one aspect only.

aniezurawski (approved):

PR #1001, opened 2022-09-15
Fix passing None as deprecated parameter to deprecated_parameter decorator

shnela:


neptune/new/internal/utils/deprecation.py
```diff
@@ -23,10 +23,8 @@ def deprecated_parameter(*, deprecated_kwarg_name, required_kwarg_name):
     def deco(f):
```
Raalsky, line 23: Changelog and LGTM 🚀

Raalsky (dismissed):

Raalsky (approved):

PR #1006, opened 2022-09-19
Automatically convert image and html like assignments to uploads

shnela:


src/neptune/new/types/type_casting.py
```diff
@@ -0,0 +1,78 @@
...
+)
+from neptune.new.types.atoms.datetime import Datetime
+from neptune.new.types.atoms.float import Float
+from neptune.new.types.atoms.string import String
+from neptune.new.types.namespace import Namespace
+from neptune.new.types.value import Value
+from neptune.new.types.value_copy import ValueCopy
+
+
+def cast_value(value: Any) -> Value:
```
Raalsky, line 44: As we extracted this func (which is so cool!) what do you think about using a single-dispatch? https://docs.python.org/3.7/library/functools.html#functools.singledispatch You should be able to split this function into multiple one that will be triggered base on arguments type.

src/neptune/new/metadata_containers/metadata_container.py
```diff
@@ -14,14 +14,12 @@
 # limitations under the License.
```
Raalsky, line 14: CHANGELOG?

src/neptune/new/types/type_casting.py
shnela: I think it's a good idea.
Let's try it.

shnela: Or maybe not.
We do casting not only based on type, but also on predicate which looks like it's not supported by singledispatch.

Raalsky (approved):

PR #1007, opened 2022-09-19
Changed source files and entry point path resolving behaviour

Raalsky:


neptune/new/internal/utils/git.py
```diff
@@ -75,17 +76,5 @@ def get_git_repo_path(initial_path: str) -> Optional[str]:
...
-    if hasattr(__main__, "__file__"):
-        git_path = get_git_repo_path(
-            initial_path=os.path.dirname(os.path.abspath(__main__.__file__))
-        )
-
-    if not git_path:
-        git_path = get_git_repo_path(initial_path=os.getcwd())
-
-    return git_path
+    potential_initial_path = os.path.dirname(os.path.abspath(get_fullpath_from_main_file()))
```
Raalsky, line 79: This still includes current working directory as whether `get_fullpath_from_main_file` will not find anything applicable then it will return "empty path" = "pathlib.Path()" which is current working directory.

shnela (dismissed):

shnela (approved):

PR #1014, opened 2022-09-22
Fix problems with timeouts in integration tests

shnela:
* Some tests were decorated to timeout after 10s, those tests caused problems, they will be timeouted by default as all other tests (we keep 120s).
* introduced bot for publishing summary of tests
* junitxml files from all tests are available in artifacts

### DIFF
--- a/.github/actions/publish-tests-result/action.yml
+++ b/.github/actions/publish-tests-result/action.yml
@@ -1,0 +1,21 @@
+---
+name: Publish tests
+description: Check python code
+runs:
+    using: "composite"
+    steps:
+      - name: Get artifacts
+        uses: actions/download-artifact@v2
+        with:
+          name: test-artifacts
+          path: ./test-results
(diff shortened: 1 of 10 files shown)


.github/workflows/scheduled_e2e.yml
```diff
@@ -40,10 +40,24 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Publish tests
+        uses: ./.github/actions/publish-tests-result
+
+  notify:
+    needs: [test-e2e]
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
shnela, line 60: `github.ref == 'refs/heads/master'` github.ref, because we sometimes use this workflow in branch to verify e2e

PK-Neptune (approved):

PR #1025, opened 2022-10-03
Deprecation of a few methods

Raalsky:

PR #1028, opened 2022-10-04
Deprecate implicit casting of objects to strings

Raalsky:


neptune/new/internal/utils/deprecation.py
```diff
@@ -19,7 +19,17 @@
...
+
+
+def simple_warning(message: str, stack_level: int):
+    warnings.simplefilter("once", DeprecationWarning)
+    warnings.warn(
+        message,
+        category=DeprecationWarning,
+        stacklevel=stack_level,
+    )
+    warnings.simplefilter("default", DeprecationWarning)
```
aniezurawski, line 32: Stan w tej libce jest per wątek?

neptune/new/internal/utils/deprecation.py
```diff
@@ -19,7 +19,17 @@
 
 from neptune.new.exceptions import NeptuneParametersCollision
 
-__all__ = ["deprecated", "deprecated_parameter"]
+__all__ = ["deprecated", "deprecated_parameter", "simple_warning"]
+
+
+def simple_warning(message: str, stack_level: int):
```
aniezurawski, line 25: Może coś z  `once` w nazwie?

neptune/new/handler.py
```diff
@@ -268,6 +269,12 @@ def log(
                 elif is_float_like(first_value):
                     attr = FloatSeries(self._container, parse_path(self._path))
                 elif is_string_like(first_value):
+                    simple_warning(
+                        "You're logging an object with an implicit cast to a string."
+                        " Use `.log(str(object))` instead."
+                        " We'll end support of this behavior in `neptune-client==1.0.0`.",
+                        stack_level=4,
```
aniezurawski, line 276: 4? Czemu nie 2?

neptune/new/internal/utils/deprecation.py
```diff
@@ -30,14 +40,11 @@ def inner(*args, **kwargs):
...
             )
 
-            warnings.simplefilter("once", DeprecationWarning)
-            warnings.warn(
+            simple_warning(
                 f"`{func.__name__}` is deprecated{additional_info}."
                 f" We'll end support of it in `neptune-client==1.0.0`.",
-                category=DeprecationWarning,
-                stacklevel=2,
+                stack_level=3,
```
aniezurawski, line 46: Można by w środku dodawać +1 i caller już mógłby nie uwzględniać wywołania samej metody simple_warning.

neptune/new/metadata_containers/metadata_container.py
```diff
@@ -274,6 +276,12 @@ def define(
         elif is_dict_like(value):
             value = Namespace(value)
         elif is_string_like(value):
+            simple_warning(
+                "You're assigning an object with an implicit cast to a string."
+                " Use `... = str(object)` instead."
+                " We'll end support of this behavior in `neptune-client==1.0.0`.",
+                stack_level=7,
```
aniezurawski, line 283: `define` jest teoretycznie metodą dostępną dla usera więc 7 brzmi dziwnie.

neptune/new/metadata_containers/metadata_container.py
```diff
@@ -274,6 +276,12 @@ def define(
         elif is_dict_like(value):
             value = Namespace(value)
         elif is_string_like(value):
+            simple_warning(
+                "You're assigning an object with an implicit cast to a string."
+                " Use `... = str(object)` instead."
+                " We'll end support of this behavior in `neptune-client==1.0.0`.",
```
aniezurawski, line 282: @normandy7 Anything you would like to change here?

neptune/new/metadata_containers/metadata_container.py
```diff
@@ -274,6 +276,12 @@ def define(
...
+                stack_level=7,
+            )
```
aniezurawski, line 284: See comments in Jira. You missed String class. ;)

aniezurawski (changes requested):

neptune/new/metadata_containers/metadata_container.py
normandy7: Ah yeah, let me give this another pass. Wasn't sure exactly what to look at, so overlooked this the first time I went over the changed files.

normandy7: To make sure I understand correctly, what kinds of object would this warning show up for? The line `" Use `... = str(object)` instead."` suggests that logging the object as a string is the intention, but could it be that it's not?

aniezurawski: For now **any type** that is not explicitly supported is implicitly converted to `str`. This behaviour will be deprecated.

normandy7: Right. Would it be more clear if we do:
```
"The object you're logging will be implicitly cast to a string."
" We'll end support of this behavior in `neptune-client==1.0.0`.",
" To log the object as a string, use `... = str(object)` instead."
```

neptune/new/internal/utils/deprecation.py
Raalsky: Fixed in source Pull Request -> https://github.com/neptune-ai/neptune-client/pull/1025

neptune/new/internal/utils/deprecation.py
Raalsky: https://github.com/neptune-ai/neptune-client/pull/1025

neptune/new/metadata_containers/metadata_container.py
Raalsky: Done

neptune/new/metadata_containers/metadata_container.py
Raalsky: It was based on my examples, especially this one:
```python
run['a1'] = A()
```

neptune/new/handler.py
Raalsky: It was based on my example (standard user-level logging with `.log`). I've added some +1 logic and will try to think it tomorrow

neptune/new/internal/utils/deprecation.py
Raalsky: There was a problems with `simplefilter` and `GitPython` (probably due to running `git` in a subprocesses). I've rewritten it to simple dictionary storage on the function level (internally it's almost the same as `warnings` library works). Operations on dict in Python should be probably atomic and thread-safe but there is logic like `if not exists then add` so probably several warnings could be called in some race conditions but <=1 per-thread (I think so). It still almost the same as `warnings.simplefilter("once", ...)`.

aniezurawski (approved):

PR #1031, opened 2022-10-05
Max line length limit set to 120 chars

Raalsky:


.pre-commit-config.yaml
```diff
@@ -9,13 +9,12 @@ repos:
     rev: 5.10.1
     hooks:
       - id: isort
-        args: ["--profile", "black", "--filter-files", "--skip", "tests/neptune/test_imports.py"]
+        args: ["--profile", "black", "--filter-files", "--skip", "tests/neptune/test_imports.py", "--line-length=120"]
```
shnela, line 12: I personally prefer multi-line imports when there's more than one module to import.

What do you think about using `--force-grid-wrap` flag instead on line-length?

shnela (approved):

shnela (approved):

PR #1040, opened 2022-10-14
Docs URL update

normandy7:
- Point docs URLs to new docs
- Update people section in README


Raalsky (approved):

Raalsky:
LGTM, FYI I checked only a fraction of the URLs to ensure they're correct.

normandy7:
> LGTM, FYI I checked only a fraction of the URLs to ensure they're correct.

In most cases it was a straightforward remapping and in a couple of cases I replaced the link with a new, more relevant page. Overall it was a rather peripheral change so should be fine.

PR #1045, opened 2022-10-17
adjust e2e tests for lightning CI

AleksanderWWW:
- add a new mark to designate relevant tests
- use conditional environment fixture creation to avoid creating new project every time

### DIFF
--- a/e2e_tests/integrations/test_lightning.py
+++ b/e2e_tests/integrations/test_lightning.py
@@ -36,8 +36,12 @@ from e2e_tests.base import BaseE2ETest
 
 LIGHTNING_ECOSYSTEM_ENV_PROJECT = "NEPTUNE_LIGHTNING_ECOSYSTEM_CI_PROJECT"
 
-skip_if_on_regular_env = pytest.mark.skipif(LIGHTNING_ECOSYSTEM_ENV_PROJECT not in os.environ)
-skip_if_on_lightning_ecosystem = pytest.mark.skipif(LIGHTNING_ECOSYSTEM_ENV_PROJECT in os.environ)
+skip_if_on_regular_env = pytest.mark.skipif(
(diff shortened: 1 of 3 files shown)


e2e_tests/integrations/test_lightning.py
```diff
@@ -33,6 +34,15 @@
 import neptune.new as neptune
 from e2e_tests.base import BaseE2ETest
 
+LIGHTNING_ECOSYSTEM_ENV_PROJECT = "NEPTUNE_LIGHTNING_ECOSYSTEM_CI_PROJECT"
+
+skip_if_on_regular_env = pytest.mark.skipif(
+    LIGHTNING_ECOSYSTEM_ENV_PROJECT not in os.environ, reason="Tests weren't invoked in Lightning Ecosystem CI"
+)
+skip_if_on_lightning_ecosystem = pytest.mark.skipif(
+    LIGHTNING_ECOSYSTEM_ENV_PROJECT in os.environ, reason="Tests invoked in Lightning Ecosystem CI"
```
shnela, line 43: This will be true even if we set `NEPTUNE_LIGHTNING_ECOSYSTEM_CI_PROJECT=0` it's better to rely on  boolean value with default fallback instead of existence of the key.
@Raalsky we had similar problem recently with non detecting `False` value of envs and have some helper methods for that.

e2e_tests/pytest.ini
```diff
@@ -3,3 +3,4 @@ markers =
     s3: a test is using AWS S3
     management: a test is using Management API
     integrations: a test is testing an integration
+    lightning: a test is supposed to run with Lightning EcoSystem CI
```
shnela, line 6: But it can be run for non lightning ecosystem, right (`skip_if_on_lightning_ecosystem`)?

shnela (changes requested):

shnela (approved):

PR #1047, opened 2022-10-18
Legacy client depreaction

Raalsky:


src/neptune/common/deprecation.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def warn_once(message: str, stack_level: int = 1):
+    if message not in warned_once:
+        warnings.warn(
+            message=message,
+            category=NeptuneDeprecationWarning,
+            stacklevel=stack_level + 1,
+        )
+        warned_once.add(message)
```
twolodzko, line 37: Won't `warnings.simplefilter("once", category=NeptuneDeprecationWarning)` suffice?

https://docs.python.org/3/library/warnings.html#warning-filter

Raalsky: We have already been testing it but seems like there are some problems around subprocess (calling `git` by `GitPython` internally) that end up with cleaned up internal counters, so every time we run `neptune.init_run` it shows warnings.

src/neptune/legacy/__init__.py
```diff
@@ -76,6 +77,7 @@ def _check_for_extra_kwargs(caller_name, kwargs: dict):
         raise TypeError(f"{caller_name}() got an unexpected keyword argument '{first_key}'")
 
 
+@legacy_client_deprecation
```
twolodzko, line 80: If we're decorating all the functions in the module, why not just throw a warning on the top of the module so it will be displayed at import from it?

Raalsky: Most of them are imported at root module `__init__.py` (https://github.com/neptune-ai/neptune-client/blob/9a0e977b8345cdba93d3c4d34547591e7649e03e/src/neptune/__init__.py#L17) so it may impact either `neptune.new` & `neptune.management`.

src/neptune/legacy/internal/utils/deprecation.py
```diff
@@ -0,0 +1,31 @@
...
+
+from neptune.common.deprecation import warn_once
+
+
+def legacy_client_deprecation(func):
+    @wraps(func)
+    def inner(*args, **kwargs):
+        warn_once(
+            message="You're using a legacy version of Neptune client."
+            " It will be moved to `neptune.legacy` since `neptune-client==1.0.0`.",
```
shnela, line 26: It will be mobed to `neptune.legacy` or removed completely?

Raalsky: @twolodzko Could you? :D I probably don't know current state.

twolodzko: It will be moved for now.

shnela (approved):

PR #1048, opened 2022-10-19
Version checking with importlib and versioneer config update

Raalsky:


src/neptune/version.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+else:
+    from importlib_metadata import (
+        PackageNotFoundError,
+        version,
+    )
+
+try:
+    __version__ = version("neptune-client")
+except PackageNotFoundError:
```
PatrykGala, line 31: Is possible catch `PackageNotFoundError` when python has the `src/neptune/version.py` in path?

PatrykGala (approved):

src/neptune/version.py
Raalsky: `importlib.metdata.version("something")` works similarly to `pip list | grep something` instead of checking the current namespace.
For instance `typing` is a proper builtin package but:
```python
>>> import typing
>>> importlib.metadata.version("typing")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/neptune/.conda/envs/neptune-client/lib/python3.10/importlib/metadata/__init__.py", line 984, in version
    return distribution(distribution_name).version
  File "/Users/neptune/.conda/envs/neptune-client/lib/python3.10/importlib/metadata/__init__.py", line 957, in distribution
    return Distribution.from_name(distribution_name)
  File "/Users/neptune/.conda/envs/neptune-client/lib/python3.10/importlib/metadata/__init__.py", line 548, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: No package metadata was found for typing
```
so I think that nothing will impact it except `pip install neptune-client` or `pip install -e .` or `pip install .` (including direct call to `python setup.py install`)

PR #1052, opened 2022-10-21
Upload in-memory files using copy stored on disk

shnela:


src/neptune/new/internal/operation_processors/operation_storage.py
```diff
@@ -29,6 +30,10 @@ class OperationStorage:
     def __init__(self, data_path: str):
         self._data_path = Path(data_path)
 
+        # initialize directories
+        os.makedirs(self.data_path)
```
Raalsky, line 34: Maybe with `exist_ok=True`

src/neptune/new/internal/operation.py
```diff
@@ -188,11 +224,19 @@ def to_dict(self) -> dict:
...
     @staticmethod
     def from_dict(data: dict) -> "UploadFile":
-        return UploadFile(data["path"], data["ext"], data["file_path"])
+        return UploadFile(data["path"], data["ext"], data["file_path"], data.get("clean_after_upload", False))
+
+    @staticmethod
+    def get_upload_path(attribute_path: List[str], extension: str, upload_path: Path):
+        now = datetime.now()
+        tmp_file_name = f"{'_'.join(attribute_path)}-{now.timestamp()}-{now}.{extension}"
+        tmp_file_name = tmp_file_name.replace(" ", "_").replace(":", ".")
```
Raalsky, line 238: What about just proper `strftime()` instead of these bad looking replace calls?

src/neptune/new/internal/operation_processors/operation_storage.py
shnela: Definitelly, good catch!

src/neptune/new/internal/operation.py
shnela: I've used to copy-paste this code over and over...
Yes, it's better idea ;)

Raalsky (approved):

PR #1074, opened 2022-11-01
Docstrings for append() and extend()

normandy7:


src/neptune/new/handler.py
```diff
@@ -254,42 +258,172 @@ def log(
...
+            * `FloatSeries` - series of float values
+            * `StringSeries` - series of strings
+            * `FileSeries` - series of files
+
+        When you log the first value, the type of the value determines what type of field is created.
+        For more, see the field types documentation: https://docs.neptune.ai/api/field_types
+
+        Args:
+            values: Values to be added to the series field.
+            steps: Optional indeces of the entries being appended. Must be strictly increasing.
```
PK-Neptune, line 369: Maybe it is worth to mention that steps must be a collection? The same with timestamps.

src/neptune/new/handler.py
```diff
@@ -254,42 +258,172 @@ def log(
...
+        Available for following series field types:
+            * `FloatSeries` - series of float values
+            * `StringSeries` - series of strings
+            * `FileSeries` - series of files
+
+        When you log the first value, the type of the value determines what type of field is created.
+        For more, see the field types documentation: https://docs.neptune.ai/api/field_types
+
+        Args:
+            values: Values to be added to the series field.
```
PK-Neptune, line 368: Maybe it is worth to mention that values have to be a dict or a collection?

src/neptune/new/handler.py
```diff
@@ -307,8 +307,34 @@ def append(
...
-        todo: NPT-12530
+        """Logs a series of values, such as a metric, by appending the provided value to the end of the series.
+
+        Available for following series field types:
+
+            * `FloatSeries` - series of float values
+            * `StringSeries` - series of strings
+            * `FileSeries` - series of files
+
+        When you log the first value, the type of the value determines what type of field is created.
```
shnela, line 318: > When you log the first value, the type of the value determines what type of field is created.


This description is not accurate since `append` is build for single values only.

shnela (approved):

src/neptune/new/handler.py
normandy7: What this means to say is: If there is no field "acc" from before, and you call

```
run["acc"].append(acc)
```

then if `acc` is a float, the field type will be FloatSeries.

Is this not accurate or is it just my phrasing that's confusing? Maybe this statement/explanation doesn't add much value?

shnela: Ok, I get it. sorry.

shnela (approved):

PR #1078, opened 2022-11-04
move sync command to new package

PatrykGala:


src/neptune/new/cli/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    "get_project",
+    "get_qualified_name",
+    "is_container_synced",
+    "get_offline_dirs",
+    "iterate_containers",
+    "create_dir_name",
+    "split_dir_name",
+]
+
+from .abstract_backend_runner import AbstractBackendRunner
```
Raalsky, line 33: Please use non-relative paths so let's use `neptune.new.cli.abstract_backend_runner` etc.

src/neptune/new/cli/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+__all__ = [
+    "AbstractBackendRunner",
```
Raalsky, line 18: Does all these things should be exposes outside of `cli` submodule? I'm not sure but I expect to only expose `status` and `sync` methods.

Raalsky: If this was just copy-pasted let's think of getting some minimal subset that should be user-exposed :wink:

src/neptune/new/internal/operation_processors/async_operation_processor.py
```diff
@@ -29,6 +29,7 @@
     Optional,
 )
 
+from neptune.new.cli.utils import create_dir_name
```
Raalsky, line 32: Ohh, I know that it was already but let's move somewhere else this `create_dir_name` (some `utils` inside of `internal`). all there OpProcessors are pretty low-level stuff.

src/neptune/new/cli/__init__.py
PatrykGala: ok

Raalsky:
CHANGELOG :wink:

Raalsky (approved):

PR #1083, opened 2022-11-09
Remove Junk metadata when script runs

PatrykGala:


src/neptune/new/internal/disk_queue.py
```diff
@@ -152,6 +153,22 @@ def close(self):
         self._last_ack_file.close()
         self._last_put_file.close()
 
+        if self.is_empty():
+            self._remove_data()
+
+    def _remove_data(self):
+        path = self._dir_path
+        shutil.rmtree(path)
```
Raalsky, line 161: Not sure but think of exception that it may raise or even think about `ignore_errors=False`

src/neptune/management/exceptions.py
```diff
@@ -163,3 +163,8 @@ class BadRequestException(ManagementOperationFailure):
 class IncorrectIdentifierException(ManagementOperationFailure):
     code = 21
     description = "Can not parse '{identifier}' as identifier."
+
+
+class ExperimentNotFound(ManagementOperationFailure):
```
Raalsky, line 168: I'm not sure about naming as we shouldn't be using `experiment` anywhere that is user-exposed. It's mostly legacy name. The most recent is probably MetadataContainer etc.

Raalsky (approved):

PR #1089, opened 2022-11-14
Add limit do disk queue

pankin397:
Now there is a limited batch size that can be pulled out of disk_queue.get_batch(), it's determined by the MAX_BATCH_SIZE env variable.


aniezurawski (changes requested):

src/neptune/new/internal/disk_queue.py
```diff
@@ -55,6 +55,11 @@ def __init__(
         self._to_dict = to_dict
         self._from_dict = from_dict
         self._max_file_size = max_file_size
+        self._max_batch_size = os.environ.get("MAX_BATCH_SIZE")
+        if self._max_batch_size is not None:
+            self._max_batch_size = int(self._max_batch_size)
+        else:
+            self._max_batch_size = 100 * 1024**2
```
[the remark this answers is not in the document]
pankin397: Done

src/neptune/new/internal/disk_queue.py
```diff
@@ -55,6 +55,11 @@ def __init__(
         self._to_dict = to_dict
         self._from_dict = from_dict
         self._max_file_size = max_file_size
+        self._max_batch_size = os.environ.get("MAX_BATCH_SIZE")
```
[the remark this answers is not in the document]
pankin397: Set to NEPTUNE_MAX_BATCH_SIZE_BYTES

src/neptune/new/internal/disk_queue.py
```diff
@@ -50,15 +50,15 @@ def __init__(
...
         self._dir_path = dir_path.resolve()
         self._to_dict = to_dict
         self._from_dict = from_dict
         self._max_file_size = max_file_size
-        self._max_batch_size = max_batch_size
-        if max_batch_size is None:
-            self._max_batch_size = int(os.environ.get("MAX_BATCH_SIZE") or str(100 * 1024**2))
+        self._max_batch_size_bytes = max_batch_size_bytes
+        if max_batch_size_bytes is None:
+            self._max_batch_size_bytes = int(os.environ.get("NEPTUNE_MAX_BATCH_SIZE_BYTES") or str(100 * 1024**2))
```
aniezurawski, line 61: Even shorter:

```
self._max_batch_size_bytes = max_batch_size_bytes or
    int(os.environ.get("NEPTUNE_MAX_BATCH_SIZE_BYTES") or str(100 * 1024**2))
```

aniezurawski: And let's declare some constant inside this class instead of using maginc number 

`DEFAULT_MAX_BATCH_SIZE_BYTES = 100 * 1024**2`.

src/neptune/new/internal/disk_queue.py
```diff
@@ -139,12 +139,13 @@ def _get_serialized(self) -> Optional[dict]:
         return _json
 
     def get_batch(self, size: int) -> Tuple[List[T], int]:
-        first, ver = self.get()
-        if not first:
-            return [], ver
+        serialized_first = self._get_serialized()
```
aniezurawski, line 142: `self.get()` was doing something more. There is logic for skipping first X operations if processing was resumed and some ops was already acked. Now this logic is missing.

aniezurawski (changes requested):

src/neptune/new/internal/disk_queue.py
```diff
@@ -111,32 +122,40 @@ def _skip_and_get(self) -> Tuple[Optional[T], int]:
...
         ret = [first]
         for _ in range(0, size - 1):
-            obj, next_ver = self._get()
-            if not obj:
+            if cur_batch_size >= self._max_batch_size_bytes:
+                break
+            serialized_obj = self._get_serialized()
+            if not serialized_obj:
                 break
+            cur_batch_size += len(serialized_obj)
```
aniezurawski, line 156: `_get_serialized()` returns `dict`. So unfortunately `len(serialized_obj)` is number of keys in `dict`, not size of json in bytes, isn't it? Maybe we should add returning size to `JsonFileSplitter.get()`, what do you think?

aniezurawski (changes requested):

src/neptune/new/internal/disk_queue.py
pankin397: I think it's a good idea. I've added a get_with_size function to the JsonFileSplitter, as well as a unit test for it. I had to refactor the DiskQueue as well.

src/neptune/new/internal/disk_queue.py
```diff
@@ -39,9 +41,16 @@
 _logger = logging.getLogger(__name__)
 
 
-class DiskQueue(Generic[T]):
+@dataclass
+class QueueElement(Generic[T]):
```
aniezurawski, line 45: Could be declared as innser class in `DiskQueue` with an underscore prefix (it private implementation detail).
@Raalsky @shnela what do you think? Do you have some reccommendation here? Some good practices about inner classes in python?

Alternatively we can keep it exposed and return it from `get` and `get_batch` as well. I think such refactor would make good to our codebase (even if size is not used outside).

src/neptune/new/internal/disk_queue.py
```diff
@@ -170,6 +177,22 @@ def close(self):
...
+        shutil.rmtree(path, ignore_errors=True)
+
+        parent = path.parent
+
+        files = os.listdir(parent)
+        if len(files) == 0:
+            try:
+                os.rmdir(parent)
+            except OSError:
+                _logger.info(f"Cannot remove directory: {parent}")
```
aniezurawski, line 194: Looks like something from another MR/task. Probably something went wrong during a rebase?

aniezurawski (changes requested):

src/neptune/new/internal/disk_queue.py
pankin397: Fixed, for some reason those changes weren't visible to me locally

src/neptune/new/internal/disk_queue.py
Raalsky: To my knowledge inner classes are not a good pattern in general. If it's only about exposing vs non-exposing something let's indicate that with `__all__` :wink:

pankin397: After searching for _dataclass_ in out repo, I haven't found examples of an inner class, so I decided to keep QueueElement exposed. However I'm not sure myself if we should make the get() function return a QueueElement or keep it unchanged.

aniezurawski: I think it is always better to return dataclass over Tuple.

pankin397: If so do you think it would make sense to create a dataclass for the JsonFileSplitter as well? Currently I refactored it so it returns a tuple with a dict and its size in serialised form.

pankin397: Done

aniezurawski (approved):

PR #1092, opened 2022-11-15
Run clean runner on sync operation

PatrykGala:


src/neptune/new/internal/disk_queue.py
```diff
@@ -233,4 +233,5 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
```
shnela, line 235: We're not using queue as context manager anymore, but if you want we can leave it.

shnela (approved):

PR #1098, opened 2022-11-18
Fix e2e tests: test_sync_container.

PatrykGala:


PatrykGala:
E2e https://github.com/neptune-ai/neptune-client/actions/runs/3496394317

shnela (dismissed):

e2e_tests/standard/test_cli.py
```diff
@@ -91,6 +96,7 @@ def test_sync_container(self, container_type, environment):
                 last_put_version_f.write(str(last_put_version + 2))
 
             with reinitialize_container(container_sys_id, container_type, project=environment.project) as container:
+                self.stop_synchronization_process(container)
```
aniezurawski, line 99: Why is it necessary here? We could even initializie it read-only mode since it's only for verification.

aniezurawski (changes requested):

e2e_tests/standard/test_cli.py
```diff
@@ -55,6 +56,10 @@ def test_sync_container(self, container_type, environment):
                 container_id = container._id
                 container_sys_id = container._sys_id
 
+                self.stop_synchronization_process(container)
```
aniezurawski, line 59: Should we call `container.sync()` before?

PatrykGala: `container.wait()` is called in 55 line

e2e_tests/standard/test_cli.py
PatrykGala: added read-only

PR #1118, opened 2022-11-25
Pre-commit extracted into separate workflow

Raalsky:


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,22 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
```
shnela, line 11: It should be 20.04 for now.

shnela (approved):

PR #1122, opened 2022-11-25
Separated package and python installation into action

Raalsky:


.github/actions/package/action.yml
```diff
@@ -0,0 +1,36 @@
...
+
+runs:
+    using: "composite"
+    steps:
+      - name: Get current date
+        id: date
+        run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - uses: actions/cache@v3
```
shnela, line 17: What is scope of this cache?
If more than one action:
* when the cache invalidates?
* we would like to install dependencies on e2e daily workflow.

If only one action but common for all workflows, then ok.

shnela (approved):

.github/actions/package/action.yml
Raalsky: To my knowledge it's global and I've made the current date a part of the cache name so it invalidates on daily basis. Check the list of currently cached deps here: https://github.com/neptune-ai/neptune-client/actions/caches

PR #1132, opened 2022-11-29
Async logging in StdStreamCaptureLogger

pankin397:
StdStreamCaptureLogger now logs asynchronously to avoid deadlock.

### DIFF
--- a/src/neptune/new/internal/streams/std_stream_capture_logger.py
+++ b/src/neptune/new/internal/streams/std_stream_capture_logger.py
@@ -41,7 +41,7 @@ class StdStreamCaptureLogger:
     def __getattr__(self, attr):
         return getattr(self.stream, attr)
 
-    def close(self, wait_for_all_logs=False):
+    def close(self, wait_for_all_logs=True):
         if not wait_for_all_logs:
             self.enabled = False
         self._log_data_queue.put_nowait(None)
@@ -64,7 +64,7 @@ class StdoutCaptureLogger(StdStreamCaptureLogger):
         super().__init__(container, attribute_name, sys.stdout)
         sys.stdout = self
 
-    def close(self, wait_for_all_logs=False):
+    def close(self, wait_for_all_logs=True):
         sys.stdout = self.stream
         super().close(wait_for_all_logs)
 
@@ -74,6 +74,6 @@ class StderrCaptureLogger(StdStreamCaptureLogger):
         super().__init__(container, attribute_name, sys.stderr)
         sys.stderr = self
 
-    def close(self, wait_for_all_logs=False):
+    def close(self, wait_for_all_logs=True):
         sys.stderr = self.stream
         super().close(wait_for_all_logs)
--- a/src/neptune/new/internal/streams/std_stream_capture_logger.py
+++ b/src/neptune/new/internal/streams/std_stream_capture_logger.py
@@ -30,7 +30,6 @@ class StdStreamCaptureLogger:
         self._thread_local = threading.local()
         self.enabled = True
         self._log_data_queue = Queue()
-        self._finished_logging = threading.Event()
         self._logging_thread = threading.Thread(target=self.__proces_logs, daemon=True)
         self._logging_thread.start()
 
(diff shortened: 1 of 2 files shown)


src/neptune/new/internal/streams/std_stream_capture_logger.py
```diff
@@ -28,41 +29,51 @@ def __init__(self, container: MetadataContainer, attribute_name: str, stream: Te
...
-            self._thread_local.inside_write = False
-
         self.stream.write(data)
-        if self.enabled and not self._thread_local.inside_write:
-            try:
-                self._thread_local.inside_write = True
-                self._logger.log(data)
-            finally:
-                self._thread_local.inside_write = False
+        self._log_data_queue.put_nowait(data)
```
aniezurawski, line 39: Probably we should somehow handle `Full` exception (https://docs.python.org/3/library/queue.html#queue.Queue.put_nowait).
Maybe write some warning to stdout without logging it to Neptune?

aniezurawski (changes requested):

src/neptune/new/internal/streams/std_stream_capture_logger.py
pankin397: Should I add a size limit as well? Because currently the queue has an infinite size and can't be full.

src/neptune/new/internal/streams/std_stream_capture_logger.py
```diff
@@ -28,41 +29,51 @@ def __init__(self, container: MetadataContainer, attribute_name: str, stream: Te
...
 
     def __getattr__(self, attr):
         return getattr(self.stream, attr)
 
-    def close(self):
+    def close(self, wait_for_all_logs=False):
```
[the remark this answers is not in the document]
pankin397: Fixed

src/neptune/new/internal/streams/std_stream_capture_logger.py
aniezurawski: Oh, ok. Then let's keep it as it is.

src/neptune/new/internal/streams/std_stream_capture_logger.py
```diff
@@ -28,41 +29,51 @@ def __init__(self, container: MetadataContainer, attribute_name: str, stream: Te
...
 class StdoutCaptureLogger(StdStreamCaptureLogger):
     def __init__(self, container: MetadataContainer, attribute_name: str):
         super().__init__(container, attribute_name, sys.stdout)
         sys.stdout = self
 
-    def close(self):
+    def close(self, wait_for_all_logs=True):
         sys.stdout = self.stream
-        super().close()
+        super().close(wait_for_all_logs)
```
aniezurawski, line 69: Let's use named arg.

src/neptune/new/internal/streams/std_stream_capture_logger.py
```diff
@@ -28,41 +29,51 @@ def __init__(self, container: MetadataContainer, attribute_name: str, stream: Te
...
 class StderrCaptureLogger(StdStreamCaptureLogger):
     def __init__(self, container: MetadataContainer, attribute_name: str):
         super().__init__(container, attribute_name, sys.stderr)
         sys.stderr = self
 
-    def close(self):
+    def close(self, wait_for_all_logs=True):
         sys.stderr = self.stream
-        super().close()
+        super().close(wait_for_all_logs)
```
aniezurawski, line 79: Let's use named arg.

src/neptune/new/internal/streams/std_stream_capture_logger.py
```diff
@@ -28,41 +29,51 @@ def __init__(self, container: MetadataContainer, attribute_name: str, stream: Te
...
 
     def __getattr__(self, attr):
         return getattr(self.stream, attr)
 
-    def close(self):
+    def close(self, wait_for_all_logs=True):
+        if not wait_for_all_logs:
+            self.enabled = False
+        self._log_data_queue.put_nowait(None)
+        self._finished_logging.wait()
```
aniezurawski, line 48: We should probably `.join()` background thread anyway, so we can wait for join instead of an event.

aniezurawski (approved):

aniezurawski (approved):

PR #1147, opened 2022-12-06
Tracking code coverage

Raalsky:


.github/actions/test-unit/action.yml
```diff
@@ -21,29 +21,26 @@ runs:
         python-version: ${{ inputs.python-version }}
         os: ${{ inputs.os }}
 
-    - name: Unit - New
+    - name: Test
```
shnela, line 24: Did we have to merge `new, management and legacy` tests together?

Raalsky: I split running these tests a few days ago with Github tweaks. I thought it would simplify the code coverage reports and we'll track it at the module level. I've tried several times to configure it but failed. In Codecov.io you'll be able to look through the coverage report on every module level so it will be almost the same :wink:

shnela (approved):

PR #1152, opened 2022-12-08
Fix test notifications on failure

shnela:


.github/workflows/e2e.yml
```diff
@@ -129,7 +129,7 @@ jobs:
   e2e-tests-notify:
     needs: [ e2e, e2e_management, e2e_s3_gcs, e2e_s3 ]
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/master'
+    if: (success() || failure()) && github.ref == 'refs/heads/master'
```
Raalsky, line 132: Isin't it the same as `always()`?

PatrykGala (approved):

.github/workflows/e2e.yml
shnela: `always()` additionally ignores manual cancellation of workflow.

PR #1157, opened 2022-12-09
Fixing code coverage

Raalsky:


pyproject.toml
```diff
@@ -192,9 +192,15 @@ force_grid_wrap = 2
...
+source_pkgs = ['neptune']
+branch = false
+relative_files = false
+
+[tool.coverage.paths]
+source = [
+    'src/neptune',
+    '/opt/hostedtoolcache/Python/*/x64/lib/python*/site-packages/neptune',
+    '/Users/runner/hostedtoolcache/Python/*/x64/lib/python*/site-packages/neptune',
+    'C:\hostedtoolcache\windows\Python\*\x64\Lib\site-packages\neptune'
```
PatrykGala, line 205: Can you read these paths from envs?

Raalsky: I don't think so as there is a wildcard in the middle to handle multiple python versions.

PatrykGala (approved):

PR #1167, opened 2022-12-22
Added __all__ to all modules in internal subpackage

Raalsky:


src/neptune/new/integrations/utils.py
```diff
@@ -13,8 +13,15 @@
...
 # limitations under the License.
 #
+from neptune.common.experiments import Experiment
+from neptune.new import Run
+from neptune.new.exceptions import NeptuneLegacyIncompatibilityException
 from neptune.new.internal.utils import verify_type
-from neptune.new.internal.utils.compatibility import expect_not_an_experiment
 from neptune.new.internal.utils.paths import join_paths
 
 __all__ = ["expect_not_an_experiment", "join_paths", "verify_type"]
```
shnela, line 22: All should be on top of the file.

src/neptune/new/internal/init/project.py
```diff
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+__all__ = ["init_project", "get_project"]
```
shnela, line 16: Do we want to add deprecated funcitons like `get_project`?

Raalsky: Yes at this point

src/neptune/new/integrations/utils.py
Raalsky: Fixed

shnela (approved):

PR #1168, opened 2022-12-22
Added __all__ to all modules in neptune.new subpackage

Raalsky:


src/neptune/new/__init__.py
```diff
@@ -24,6 +23,39 @@
...
    https://docs.neptune.ai/api/neptune
 """
+__all__ = [
+    "types",
+    "ANONYMOUS",
+    "ANONYMOUS_API_TOKEN",
+    "NeptunePossibleLegacyUsageException",
+    "NeptuneUninitializedException",
+    "get_project",
+    "init",
```
shnela, line 33: Are you sure that we want to keep such deprecated modules here as well (`init`)?

Raalsky: Yes at this point :wink:. This could be limited as well after 1.0.0 release

shnela (approved):

PR #1170, opened 2022-12-23
Separate action for notifications about job result

Raalsky:


.github/actions/job-notify/action.yml
```diff
@@ -0,0 +1,32 @@
+---
+name: Notify about job status
+description: Send Slack channel notification about job finish status
+inputs:
+  slack-webhook:
+    description: "A Slack Webhook to post a notification"
+    required: true
+  job-status:
```
shnela, line 8: How does this `job-status:` solve problem of common error status for all messages?

Raalsky: Before it was checking the whole workflow status (all integrations) in every notification job. Now it checks the status of a particular job with tests that should match the `needs` section for every notification jobs.

shnela (approved):

PR #1177, opened 2023-01-13
Draft: Deprecate implicit casting to strings fixes and stringify_unsupported

Raalsky:


src/neptune/new/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    List,
+    Mapping,
+    Union,
+)
+
+from neptune.new.internal.utils.stringify_value import StringifyValue
+
+
+def stringify_unsupported(value) -> Union[StringifyValue, Mapping, List]:
+    if isinstance(value, dict):
```
shnela, line 28: Maybe we should use `is_dict_like` or even take `argaprse.Namespace` and `neptune.Namespace`.

I don't know, for me it's a problematic topi I've addressed earlier.

src/neptune/new/utils.py
```diff
@@ -0,0 +1,34 @@
...
+        return {k: stringify_unsupported(v) for k, v in value.items()}
+
+    if isinstance(value, list):
```
shnela, line 31: Maybe we should use `Collection` or `Iterable`?

src/neptune/common/deprecation.py
```diff
@@ -25,13 +29,22 @@ class NeptuneDeprecationWarning(DeprecationWarning):
 warnings.simplefilter("always", category=NeptuneDeprecationWarning)
 
 warned_once = set()
+path_to_root_module = os.path.dirname(os.path.realpath(neptune.__file__))
+
+
+def get_user_code_stack_level():
```
shnela, line 35: Is it finally working with `run[path] =` and `run[path].function(args)`?

If so, congrats!

src/neptune/new/types/type_casting.py
```diff
@@ -111,12 +121,17 @@ def cast_value_for_extend(values: Union[Namespace, Series, Collection[Any]]) ->
...
-        warn_once(
-            message="The object you're logging will be implicitly cast to a string."
-            " We'll end support of this behavior in `neptune-client==1.0.0`."
-            " To log the object as a string, use `str(object)` instead.",
-            stack_level=4,
-        )
+        import sys
+        import traceback
+
+        traceback.print_stack(file=sys.stdout)
```
shnela, line 127: Why do we print stack for extend, but not for regular cast?

shnela (changes requested):

src/neptune/new/types/type_casting.py
Raalsky: Ouch! That's debug. Good catch

Raalsky: Fixed

src/neptune/common/deprecation.py
Raalsky: Everything ❤️ 
Just look at these messages:
Screenshot 2023-01-17 at 14 26 48

src/neptune/new/utils.py
Raalsky: I know, I know. I just wonder about covering the cases mentioned in the spec. Not sure what I have to cover. The `argparse.Namespace` works correctly:
```python
run['a/with_warning'] = stringify_unsupported(argparse.Namespace(test="test", test2=5.3))
run['a/with_warning'] = argparse.Namespace(test="test", test2=5.3)
```
and doesn't generates any warnings in both usages (as it's a supported type).

src/neptune/new/utils.py
Raalsky: I know but I'm not sure: what do I have to return in such a case? A map generator or something?

src/neptune/new/utils.py
shnela: What about `argparse.Namespace(test=SomeClass(), test2=5.3)`?

shnela (approved):
Good enough.

src/neptune/legacy/internal/utils/deprecation.py
```diff
@@ -23,8 +23,7 @@ def legacy_client_deprecation(func):
     def inner(*args, **kwargs):
         warn_once(
             message="You're using a legacy version of Neptune client."
-            " It will be moved to `neptune.legacy` since `neptune-client==1.0.0`.",
-            stack_level=2,
+            " It will be moved to `neptune.legacy` since `neptune-client==1.0.0`."
```
normandy7, line 26: Change "since" to "as of"

src/neptune/new/handler.py
```diff
@@ -311,12 +316,13 @@ def log(
...
-                        " We'll end support of this behavior in `neptune-client==1.0.0`."
-                        " To log the object as a string, use `.log(str(object))` instead.",
-                        stack_level=3,
-                    )
+                    if not from_stringify_value:
+                        warn_once(
+                            message="The object you're logging will be implicitly cast to a string."
+                            " We'll end support of this behavior in `neptune-client==1.0.0`."
+                            " To log the object as a string, use `.log(str(object))` or"
+                            " `.log(stringify_unsupported(collection))` for collections and dictionaries instead."
```
normandy7, line 324: Remove "instead" from here, it doesn't flow well. Add link to https://docs.neptune.ai/setup/neptune-client_1-0_release_changes

src/neptune/new/handler.py
```diff
@@ -311,12 +316,13 @@ def log(
                 elif is_float_like(first_value):
                     attr = FloatSeries(self._container, parse_path(self._path))
                 elif is_string_like(first_value):
-                    warn_once(
-                        message="The object you're logging will be implicitly cast to a string."
-                        " We'll end support of this behavior in `neptune-client==1.0.0`."
-                        " To log the object as a string, use `.log(str(object))` instead.",
-                        stack_level=3,
-                    )
+                    if not from_stringify_value:
```
normandy7, line 319: This is for `log()`; what about `append()` and `extend()`?

src/neptune/new/internal/utils/deprecation.py
```diff
@@ -22,16 +22,15 @@
...
     def deco(func):
         @wraps(func)
         def inner(*args, **kwargs):
             additional_info = f", use `{alternative}` instead" if alternative else " and will be removed"
 
             warn_once(
                 message=f"`{func.__name__}` is deprecated{additional_info}."
-                f" We'll end support of it in `neptune-client==1.0.0`.",
-                stack_level=stack_level + 1,
+                f" We'll end support of it in `neptune-client==1.0.0`."
```
normandy7, line 33: Can we add to these a link to https://docs.neptune.ai/setup/neptune-client_1-0_release_changes ?

src/neptune/new/types/type_casting.py
```diff
@@ -76,16 +81,16 @@ def cast_value(value: Any) -> Value:
...
-            message="The object you're logging will be implicitly cast to a string."
-            " We'll end support of this behavior in `neptune-client==1.0.0`."
-            " To log the object as a string, use `str(object)` instead.",
-            stack_level=6,
-        )
+        if not from_stringify_value:
+            warn_once(
+                message="The object you're logging will be implicitly cast to a string."
+                " We'll end support of this behavior in `neptune-client==1.0.0`."
+                " To log the object as a string, use `str(object)` or `repr(object)` instead."
```
normandy7, line 92: Add link to https://docs.neptune.ai/setup/neptune-client_1-0_release_changes

src/neptune/new/types/type_casting.py
```diff
@@ -111,12 +121,13 @@ def cast_value_for_extend(values: Union[Namespace, Series, Collection[Any]]) ->
...
-            " We'll end support of this behavior in `neptune-client==1.0.0`."
-            " To log the object as a string, use `str(object)` instead.",
-            stack_level=4,
-        )
+        if not from_stringify_value:
+            warn_once(
+                message="The object you're logging will be implicitly cast to a string."
+                " We'll end support of this behavior in `neptune-client==1.0.0`."
+                " To log the object as a string, use `str(object)` or"
+                " `stringify_unsupported(collection)` for collections and dictionaries instead."
```
normandy7, line 129: Remove "instead". Add link to https://docs.neptune.ai/setup/neptune-client_1-0_release_changes

src/neptune/new/utils.py
```diff
@@ -0,0 +1,34 @@
...
+from typing import (
+    List,
+    Mapping,
+    Union,
+)
+
+from neptune.new.internal.utils.stringify_value import StringifyValue
+
+
+def stringify_unsupported(value) -> Union[StringifyValue, Mapping, List]:
```
normandy7, line 27: Docstring proposal:

```
"""Helper function that converts unsupported values in a collection or dictionary to strings.

Example:
    run = neptune.init_run()
    complex_dict = {"tuple": ("hi", 1), "metric": 0.87}
    run["complex_dict"] = complex_dict
    # Error: The tuple is not a supported type
    from neptune.utils import stringify_unsupported
    run["complex_dict"] = stringify_unsupported(complex_dict)
"""
```

Raalsky: I wonder about `Error: ....` as currently this will work but only shows a warning message. It will raise an error after 1.0.0 release or even a little further. 

What about args and `value` parameter?

src/neptune/new/handler.py
Raalsky: The same message is for both `append` and `extend` and it's located in `type_casting.py`

src/neptune/new/utils.py
normandy7: I'm sorry but I don't understand. What and where is the error?

And I suppose to be complete we can list `value` as an arg, but IMO the description and example make it pretty clear, no?

src/neptune/new/handler.py
normandy7: I cannot find it. I was just wondering since the text mentions `log()` but we generally want to use `append()` instead, since we no longer use `log()` in docs or examples.

Raalsky: `cast_value_for_extend` in t`type_casting.py:100`. As it might be called with `append` and `expand` it wasn't mentioned (or probably should take into account the "or" operator).

About `log` -> I know, I know, but still, it wasn't deprecated ("yet") and `extend` might not be as common for users as we may think and the message that uses `extend` when `log` was originally called might be misleading.

src/neptune/new/utils.py
Raalsky: These line:
```
    # Error: The tuple is not a supported type
```

normandy7: I see what you mean now. Got lost thinking it referred to some actual code. I'll add something

src/neptune/new/handler.py
normandy7: Got it, all good

PR #1178, opened 2023-01-16
Better support for Handler level objects

AleksanderWWW:
- Resolving NPT-12803
- Fixing bug (missing decorator for append and extend)

### DIFF
--- a/src/neptune/new/handler.py
+++ b/src/neptune/new/handler.py
@@ -318,6 +318,7 @@ class Handler:
                 self._container.set_attribute(self._path, attr)
             attr.log(value, step=step, timestamp=timestamp, wait=wait, **kwargs)
 
+    @check_protected_paths
     def append(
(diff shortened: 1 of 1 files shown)


src/neptune/new/handler.py
```diff
@@ -113,6 +120,14 @@ def __getitem__(self, path: str) -> "Handler":
     def __setitem__(self, key: str, value) -> None:
         self[key].assign(value)
 
+    def __getattr__(self, item: str):
+        run_level_methods = {"exists", "get_structure", "get_run_url", "print_structure", "stop", "sync", "wait"}
+
+        if item in run_level_methods:
+            raise AttributeError("Run-level method used on Handler object.")
```
AleksanderWWW, line 127: Shortly will introduce a better error message

shnela (approved):

PR #1191, opened 2023-01-26
Add RunType type alias

twolodzko:
In integrations, in many places, we would allow for using `Handler` or `Run`. I added a type annotation to make annotating more consistent for such a case.

Since this is for internal use, I assumed that there is no need to update CHANGELOG for that.

### DIFF
--- a/src/neptune/new/__init__.py
+++ b/src/neptune/new/__init__.py
@@ -36,6 +36,7 @@ __all__ = [
     "init_project",
     "init_run",
     "Run",
+    "RunType",
     "__version__",
     "create_experiment",
     "get_experiment",
@@ -79,8 +80,10 @@ from neptune.new.internal.init import (
 from neptune.new.internal.utils.deprecation import deprecated
 from neptune.new.metadata_containers import Run
 from neptune.new.version import version
+from neptune.new.run import RunType
 
 __version__ = str(version)
+
 
 
 def _raise_legacy_client_expected(*args, **kwargs):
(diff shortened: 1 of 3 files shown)


src/neptune/new/__init__.py
```diff
@@ -36,6 +36,7 @@
     "init_project",
     "init_run",
     "Run",
+    "RunType",
```
Raalsky, line 39: Let's move it to `new/integrations/utils.py` instead.

twolodzko: But it isn't strictly related to integrations. 

Moreover, my thinking was to put it close to Run so in case it is refactored, it is harder to miss the type signature.

src/neptune/new/run.py
```diff
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
```
Raalsky, line 13: This module shouldn't be updated. It's only for backward compatibility. The `Run` is currently located in: https://github.com/neptune-ai/neptune-client/blob/master/src/neptune/new/metadata_containers/run.py

src/neptune/new/__init__.py
Raalsky: > Since this is for internal use, I assumed that there is no need to update CHANGELOG for that.

It isn't internal as it was added to the most exposed module :P. 

> But it isn't strictly related to integrations.

What is an example of something that is not the integration but is still internal?

Raalsky: > Moreover, my thinking was to put it close to Run so in case it is refactored, it is harder to miss the type signature.

Even it's close it doesn't mean that will be updated. I think IDE will handle that and no reason to concern

twolodzko: Done

Raalsky (approved):

PR #1204, opened 2023-02-08
Do not use legacy wrapper in common package

aniezurawski:


src/neptune/new/envs.py
```diff
@@ -22,7 +22,6 @@
     "NEPTUNE_ALLOW_SELF_SIGNED_CERTIFICATE",
     "NEPTUNE_NOTEBOOK_ID",
     "NEPTUNE_NOTEBOOK_PATH",
-    "NEPTUNE_RETRIES_TIMEOUT_ENV",
```
Raalsky, line 25: Add an import from common package here for backward compatibility.

Raalsky (changes requested):

Raalsky (approved):

PR #1210, opened 2023-02-09
Add R integration to e2e tests

twolodzko:


.github/workflows/integrations.yml
```diff
@@ -382,3 +382,49 @@ jobs:
...
+      NEPTUNE_PROJECT: e2e-tests/integrations
+    timeout-minutes: 75
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {os: macos-latest,   r: 'release'}
+          - {os: ubuntu-latest,  r: 'release'}
+    runs-on: ${{ matrix.os }}
+    name: 'R (${{ matrix.os }})'
```
Raalsky, line 397: Underlying miniconda action can handle different Python versions:
```
      - uses: conda-incubator/setup-miniconda@v2
        with:
          auto-update-conda: true
          python-version: ${{ matrix.python-version }}
```
Could we pass it through and test on 3.7 + 3.9/3.10?

.github/workflows/integrations.yml
```diff
@@ -382,3 +382,49 @@ jobs:
...
+
+  rlang:
+    env:
+      NEPTUNE_PROJECT: e2e-tests/integrations
+    timeout-minutes: 75
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {os: macos-latest,   r: 'release'}
```
Raalsky, line 394: Maybe Release + Some older version of R?

twolodzko: Not sure if it is worth it. We have just a few users of this integration.

.github/workflows/integrations.yml
twolodzko: We could, but do we need it? Usually, unit tests are not for testing the dependencies, and this would only run the client code using different Python versions, which should already be tested by the unit tests for the client itself. This integration only calls the public API of the client, so if the public API works, we don't need to recheck it.

.github/workflows/integrations.yml
Raalsky: Ok

.github/workflows/integrations.yml
Raalsky: Ok. I buy it

Raalsky (dismissed):

Raalsky (approved):

PR #1211, opened 2023-02-09
Fix syncing offline runs with file upload

shnela:


src/neptune/new/cli/sync.py
```diff
@@ -80,8 +81,9 @@ def sync_execution(
         container_id: UniqueId,
         container_type: ContainerType,
     ) -> None:
+        operation_storage = OperationStorage(str(execution_path))
```
aniezurawski, line 84: Why convert `Path` to `str`?

src/neptune/new/cli/sync.py
```diff
@@ -102,6 +104,7 @@ def sync_execution(
                             container_id=container_id,
                             container_type=container_type,
                             operations=batch,
+                            upload_path=operation_storage.upload_path,
```
aniezurawski, line 107: If we already have such cool abstract like OperationStorage then let's use it instead of raw path. Just add method `Path OperationStorage::get_file(UploadFile op)` or something like that and use it inside `execute_operations`. Abstractions make code clean.

src/neptune/new/internal/operation.py
```diff
@@ -197,21 +202,16 @@ def of_file(cls, value: File, attribute_path: List[str], upload_path: Path):
...
+            value._save(upload_path / tmp_file_name)
+            operation = UploadFile(path=attribute_path, ext=value.extension, tmp_file_name=tmp_file_name)
         else:
             raise ValueError(f"Unexpected FileType: {value.file_type}")
         return operation
 
-    def clean(self):
-        if self.clean_after_upload:
-            os.remove(self.file_path)
+    def clean(self, upload_path: Path):
```
aniezurawski, line 212: Why does it need to take this path as an argument if it stores path by itself? Very confusing.

src/neptune/new/internal/operation.py
```diff
@@ -185,7 +188,9 @@ def from_dict(data: dict) -> "AssignArtifact":
 class UploadFile(Operation):
 
     ext: str
-    file_path: str
+    file_path: str = None
+    tmp_file_name: str = None
+    # `clean_after_upload` is for backward compatibility and should be removed in the future
```
aniezurawski, line 193: Could you explain? When can we remove it? Can we remove it already?

aniezurawski (changes requested):

src/neptune/new/cli/sync.py
shnela: Because `OperationStorage` requires string, but ok. I can make `OperationStorage` initializer more flexible.

src/neptune/new/internal/operation.py
shnela: In case when we actually want to remove file we store only file name.

src/neptune/new/internal/backends/hosted_neptune_backend.py
```diff
@@ -522,7 +526,7 @@ def _execute_upload_operations(
                     swagger_client=self.leaderboard_client,
                     container_id=container_id,
                     attribute=path_to_str(op.path),
-                    source=op.file_path,
+                    source=op.get_absolute_path(operation_storage),
```
aniezurawski, line 529: Looks like both - `clear` and `get_absolute_path` should be defined in OperationStorage. Operation should be a simple serializable dataclass with minimum logic. And it would be better to pass operation object instead of raw path as an argument.

src/neptune/new/internal/operation.py
```diff
@@ -197,21 +203,16 @@ def of_file(cls, value: File, attribute_path: List[str], upload_path: Path):
...
-            value._save(tmp_file_path)
-            operation = UploadFile(
-                path=attribute_path,
-                ext=value.extension,
-                file_path=os.path.abspath(tmp_file_path),
-                clean_after_upload=True,
-            )
+            tmp_file_name = cls.get_tmp_file_name(attribute_path, value.extension)
+            value._save(upload_path / tmp_file_name)
+            operation = UploadFile(path=attribute_path, ext=value.extension, tmp_file_name=tmp_file_name)
```
aniezurawski, line 208: `clean_after_upload` is missing. Is it intentional?

src/neptune/new/internal/operation.py
```diff
@@ -185,7 +189,9 @@ def from_dict(data: dict) -> "AssignArtifact":
 class UploadFile(Operation):
 
     ext: str
-    file_path: str
+    file_path: str = None
+    tmp_file_name: str = None
+    # `clean_after_upload` is for backward compatibility and should be removed in the future
     clean_after_upload: bool = False
 
     @classmethod
```
aniezurawski, line 197: Maybe pass `OperationStorage` instead of `upload_path`?

aniezurawski (changes requested):

src/neptune/new/internal/operation.py
shnela: Yes. Now we determine if operation should be cleaned by existence of `tmp_file_name` value.

`clean_after_upload` is deprecated flag leaved only for handling of old format operations if user invoke sync on old data.

src/neptune/new/internal/operation.py
shnela: Ok, why not.

src/neptune/new/internal/backends/hosted_neptune_backend.py
shnela: But if we define `clear` and `get_absolute_path` in `OperationStorage` we'll have to pass `Operation` (in particular `UploadFile`) to those functions because we would like to clear or get path of particular operation.

At the same time we need `upload_path` of `OperationStorage` when creating operation (when we create `UploadFile` of `in-memory` file we also save file to `.neptune` catalog).

It'll require `OperationStorage` to be some kind of proxy on `UploadFile` operation irresponsible also for creating it.

If you're ok with that I can do that, what do you think?

aniezurawski (approved):

aniezurawski (approved):

aniezurawski (approved):

PR #1215, opened 2023-02-13
Disabled automatic casting to strings

Raalsky:

### DIFF
--- a/src/neptune/types/type_casting.py
+++ b/src/neptune/types/type_casting.py
@@ -83,8 +83,8 @@ def cast_value(value: Any) -> Value:
         return Float(value)
     elif is_dict_like(value):
         return Namespace(value)
-    elif from_stringify_value or (is_collection(value) and all(is_stringify_value(elem) for elem in value)):
-        return String(repr(value))
+    elif (is_collection(value) and all(is_stringify_value(elem) for elem in value)) or from_stringify_value:
+        return String(str(value))
     else:
         raise TypeError("Value of unsupported type {}".format(type(value)))
 
--- a/src/neptune/types/type_casting.py
+++ b/src/neptune/types/type_casting.py
@@ -25,6 +25,7 @@ from typing import (
 
 from neptune.internal.utils import (
     is_bool,
+    is_collection,
     is_dict_like,
     is_float,
     is_float_like,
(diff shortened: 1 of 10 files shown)


src/neptune/types/type_casting.py
```diff
@@ -121,15 +113,7 @@ def cast_value_for_extend(values: Union[Namespace, Series, Collection[Any]]) ->
...
-                message="The object you're logging will be implicitly cast to a string."
-                " We'll end support of this behavior in `neptune-client==1.0.0`."
-                " To log the object as a string, use `str(object)` or"
-                " `stringify_unsupported(collection)` for collections and dictionaries."
-                " For details, see https://docs.neptune.ai/setup/neptune-client_1-0_release_changes"
-            )
+    elif from_stringify_value:
         return StringSeries(values=values)
     else:
         raise TypeError("Value of unsupported type List[{}]".format(type(sample_val)))
```
Raalsky, line 119: @normandy7 after 1.0.0 this will raise `TypeError` if anyone will try to assign an unsupported type (including all that were converted to string in `0.x.x` and anything else that were unsupported in prior versions as well). Maybe we would like to improve it anyhow? This also applies to `log` and `extend` & `append` methods and could be specialized to the called method (`append` and `extend` are combined).

shnela (approved):

src/neptune/types/type_casting.py
normandy7: @Raalsky Yes, I think we can use a similar message as before. I think it might be best to create a docs page for this error and how to work around it, then we can simply add a link.

Would something like this work?

```
            warn_once(
                message="The type of the object you're logging is not supported by"
                " Neptune. To log the object as a string, use `str(object)` or"
                " `stringify_unsupported(collection)` for collections and dictionaries."
                " For details, see https://docs.neptune.ai/help/value_of_unsupported_type"
            )
```

I can open a separate PR with this if needed.

Raalsky: This couldn't be a warning. Probably a new exception. I'll take care of it later today :wink:

Raalsky: https://github.com/neptune-ai/neptune-client/pull/1229

PR #1228, opened 2023-02-15
Update methods to have mainly keyword arguments

shnela:


src/neptune/attributes/namespace.py
```diff
@@ -91,14 +92,14 @@ def to_dict(self) -> Dict[str, Any]:
                 result[key] = value
         return result
 
-    def assign(self, value: Union[NamespaceVal, dict, Mapping], wait: bool = False):
+    def assign(self, value: Union[NamespaceVal, dict, Mapping], *, wait: bool = False):
```
AleksanderWWW, line 95: Just out of curiosity - do we need dict, when we have Mapping?

AleksanderWWW (approved):

src/neptune/attributes/namespace.py
shnela: IDK

AleksanderWWW (approved):

PR #1233, opened 2023-02-15
Change run status to Active / Inactive

shnela:


src/neptune/legacy/internal/api_clients/hosted_api_clients/hosted_alpha_leaderboard_api_client.py
```diff
@@ -1009,7 +1009,7 @@ def get_leaderboard_entries(
         min_running_time=None,
     ):
         if states is not None:
-            states = [state if state == "running" else "idle" for state in states]
+            states = [state if state == "Active" else "Inactive" for state in states]
```
Raalsky, line 1012: Too far, let's leave `legacy` unchanged :wink:

src/neptune/internal/init/project.py
```diff
@@ -34,8 +34,8 @@
 
 
 def init_project(
-    *,
     project: Optional[str] = None,
+    *,
```
Raalsky, line 38: Did you applied it here: https://github.com/neptune-ai/neptune-client/pull/1228 as well?

tests/unit/neptune/new/internal/backends/test_nql.py
```diff
@@ -54,10 +54,10 @@ def test_attributes(self):
                     name="sys/state",
                     type=NQLAttributeType.EXPERIMENT_STATE,
                     operator=NQLAttributeOperator.EQUALS,
-                    value="running",
+                    value="Active",
```
Raalsky, line 57: This shouldn't be updated (calling an filtering API)

src/neptune/metadata_containers/project.py
```diff
@@ -199,8 +199,8 @@ def fetch_runs_table(
                 Matching any element of the list is sufficient to pass the criterion.
                 Defaults to `None`.
             state: Run state, or list of states.
-                Example: `"running"` or `["idle", "running"]`.
-                Possible values: "idle", "running".
+                Example: `"Active"` or `["Inactive", "Active"]`.
```
Raalsky, line 202: Ok but we have to translate it in two places:
- On NQL creation (a query to Leaderboard filters)
- On table entries creation (the result of `fetch_runs_table`)

src/neptune/legacy/internal/api_clients/hosted_api_clients/hosted_alpha_leaderboard_api_client.py
shnela: Oh, sorry...

tests/unit/neptune/new/internal/backends/test_nql.py
shnela: Hm, I assumed that we adapt to some new API.

src/neptune/internal/init/project.py
shnela: No, I've messed with branches.
But let's add it here.

src/neptune/metadata_containers/project.py
shnela: Ok...
So I've created intermediate `RunState` enum which is responsible for translating strings.

src/neptune/internal/init/project.py
Raalsky: Ok

src/neptune/metadata_containers/project.py
```diff
@@ -124,16 +125,17 @@ def _prepare_nql_query(ids, states, owners, tags):
...
+            api_states = list(map(lambda s: RunState(s).to_api(), states))
             query_items.append(
                 NQLQueryAggregate(
                     items=[
                         NQLQueryAttribute(
                             name="sys/state",
                             type=NQLAttributeType.EXPERIMENT_STATE,
                             operator=NQLAttributeOperator.EQUALS,
-                            value=state,
+                            value=api_state,
```
Raalsky, line 136: What about just `value=RunState(state).to_api()` instead of separate list?

src/neptune/metadata_containers/run_state.py
```diff
@@ -0,0 +1,45 @@
+#
```
Raalsky, line 1: How about moving it directly to `metadata_containers_table` or to `internal` subpackage?

src/neptune/metadata_containers/project.py
shnela: Good point.

I've implemented it higher and then moved the code here.

src/neptune/metadata_containers/run_state.py
shnela: +1

Raalsky (approved):

PR #1234, opened 2023-02-15
Remove properties and constants

AleksanderWWW:

### DIFF
--- a/src/neptune/constants.py
+++ b/src/neptune/constants.py
@@ -14,10 +14,8 @@
 # limitations under the License.
 #
 __all__ = [
-    "ANONYMOUS",
     "ANONYMOUS_API_TOKEN",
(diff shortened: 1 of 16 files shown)


Raalsky (approved):

src/neptune/constants.py
```diff
@@ -26,16 +24,13 @@
 
 """Constants used by Neptune"""
 
-ANONYMOUS = "ANONYMOUS"
```
normandy7, line 29: Does this remove the option to pass the literal string as the argument (`api_token="ANONYMOUS"`)?

AleksanderWWW: It seems so. Tried doing that and got `neptune.common.exceptions.NeptuneInvalidApiTokenException`.

PR #1246, opened 2023-02-18
Made `Run`, `Model`, `ModelVersion` and `Project` made first class citizens

Raalsky:
This allows us to call `Run` and any other object in-place of `init_run`:
```python
with Run() as run:
    run['a'] = 99.9
```

---


twolodzko:
I didn't review the implementation, just looked at the API, and see no reason why we shouldn't have it.

AleksanderWWW (approved):
Looks good to me on the code front. Doc strings probably need to be updated, as they only point to using `init_*` aliases

Raalsky:
@normandy7, this is mostly our internal code deduplication etc. but as a result, this adds a feature to use `Run`/`Model`/`ModelVersion`/`Project` in addition to standard `init_run`. There is no requirement to expose it currently in docs/docstrings. For example:
```python
with Run() as run:
    run['a'] = 99.9

# or

run = Run()
```

normandy7 (approved):
I agree no docs action should be required for now. We may update the class docstrings in the future, should we decide to "officially" include the constructor approach in the API. Thanks!

PR #1254, opened 2023-02-21
Proper input value type verification for `append` method

Raalsky:

### DIFF
--- a/src/neptune/handler.py
+++ b/src/neptune/handler.py
@@ -681,12 +681,7 @@ class ExtendUtils:
         so work can be delegated to `extend` method."""
         if isinstance(value, Namespace) or is_dict_like(value):
             return {k: ExtendUtils.validate_and_transform_to_extend_format(v) for k, v in value.items()}
-        elif is_collection(value):
-            raise NeptuneUserApiInputException(
-                "Value cannot be a collection, if you want to `append` multiple values at once use `extend` method."
-            )
-        else:
-            return [value]
+        return [value]
 
     @staticmethod
     def validate_values_for_extend(values, steps, timestamps):
--- a/src/neptune/handler.py
+++ b/src/neptune/handler.py
@@ -380,7 +380,7 @@ class Handler:
         if timestamp is not None:
             timestamp = [timestamp]
 
-        value = ExtendUtils.validate_and_transform_to_extend_format(value)
+        value = ExtendUtils.transform_to_extend_format(value)
         self.extend(value, steps=step, timestamps=timestamp, wait=wait, **kwargs)
 
     @check_protected_paths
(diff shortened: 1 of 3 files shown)


src/neptune/handler.py
```diff
@@ -681,12 +681,7 @@ def validate_and_transform_to_extend_format(value):
         so work can be delegated to `extend` method."""
```
shnela, line 681: Can we change `validate_and_transform_to_extend_format` to `transform_to_extend_format`?

tests/unit/neptune/new/test_handler.py
```diff
@@ -270,15 +271,15 @@ def test_log_many_values(self):
 
     def test_append_many_values_cause_error(self):
```
shnela, line 272: It sould not be called `test_append_many_values_cause_error`, now error is caused by non-supported value.
We could test some `set`, `tuple` and instance of custom class here as well.

src/neptune/handler.py
Raalsky: Done

tests/unit/neptune/new/test_handler.py
Raalsky: Done

shnela (approved):

Raalsky:
@normandy7, we had a bug that whenever the user was calling the `.append(...)` method with list/tuple/list we were raising an exception about provided collection (probably to differentiate it from `extend`). In the spec, the `append` method should accept a dictionary (of namespaces) or any other value as a single. There was no need for such a check.

normandy7:
> @normandy7, we had a bug that whenever the user was calling the `.append(...)` method with list/tuple/list we were raising an exception about provided collection (probably to differentiate it from `extend`). In the spec, the `append` method should accept a dictionary (of namespaces) or any other value as a single. There was no need for such a check. Probably no action is needed on your side.

Got it. So this was not possible before:

```
run["train"].append({"acc": acc, "loss": loss})
```

but should be OK after this fix?

Raalsky:
Not exactly.:
`run["train"].append({"acc": acc, "loss": loss})`:
- Before: Works as expected
- After this PR: Still works


`run['a'].append((1, 2))` / `run['a'].append([3, 4])` / `run['a'].append({5, 6})`:
- Before: raises an exception about being collection and invalid type provided
- After this PR: If we'll publish it somehow before 1.0.0 then it would be just translated to string without any warning. After 1.0.0 this will warn with the default `NeptuneUnsupportedType` warning.

normandy7:
> * After this PR: If we'll publish it somehow before 1.0.0 then it would be just translated to string without any warning. After 1.0.0 this will warn with the default `NeptuneUnsupportedType` warning.

Ah, makes sense.

PR #1255, opened 2023-02-21
Warning instead of an exception for unsupported types

Raalsky:

### DIFF
--- a/src/neptune/common/deprecation.py
+++ b/src/neptune/common/deprecation.py
@@ -13,7 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-__all__ = ["warn_once", "NeptuneDeprecationWarning", "NeptuneWarning"]
+__all__ = [
+    "warn_once",
+    "warn_about_unsupported_type",
+    "NeptuneDeprecationWarning",
+    "NeptuneWarning",
+]
 
 import os
 import traceback
@@ -27,6 +32,10 @@ class NeptuneDeprecationWarning(DeprecationWarning):
 
 
 class NeptuneWarning(Warning):
+    pass
+
+
+class NeptuneUnsupportedType(Warning):
     pass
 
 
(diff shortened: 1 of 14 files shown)


src/neptune/common/deprecation.py
```diff
@@ -44,11 +53,22 @@ def get_user_code_stack_level():
...
             message=message,
             category=exception,
             stacklevel=get_user_code_stack_level(),
         )
         warned_once.add(message)
+
+
+def warn_about_unsupported_type(type_str: str):
+    warn_once(
+        message=f"""The type of the object you're logging is not supported by Neptune ({type_str}).
```
Raalsky, line 71: @normandy7, that's for you. Remember that warnings should be more compact than exceptions :wink:

normandy7: So you're you're saying the current message should be condensed? I can even shorten the docs URL :)

**Q:** If it's no longer a named exception, will the user see `NeptuneUnsupportedType` in the output? (Asking for docs search purposes)

Raalsky: > So you're you're saying the current message should be condensed? 
Not really. Could be even longer. I was referencing that in case of exceptions there are some "fancy" things like newlines, separators etc.

Raalsky: >Q: If it's no longer a named exception, will the user see NeptuneUnsupportedType in the output? (Asking for docs search purposes)

Currently it looks like this:
```
>>> class Obj:
...     pass
...
>>> run['a'] = Obj()
<stdin>:1: NeptuneUnsupportedType: The type of the object you're logging is not supported by Neptune (<class '__main__.Obj'>).
        Need help? -> https://docs.neptune.ai/help/value_of_unsupported_type
```

normandy7 (approved):

AleksanderWWW (approved):

PR #1280, opened 2023-03-01
Draft: Added Assignable interface

Raalsky:

- [ ] Did you **asked the docs owner** to review all the updated user-facing interfaces?


src/neptune/metadata_containers/abstract.py
```diff
@@ -0,0 +1,62 @@
...
+    def __getitem__(self, path: str) -> "Handler":
+        ...
+
+    @abstractmethod
+    def __setitem__(self, key: str, value) -> None:
+        ...
+
+    @abstractmethod
+    def __delitem__(self, path) -> None:
+        ...
```
twolodzko, line 62: Could we add `get_root_object` to the interface? + to `Run` as something like

```python
def get_root_object(self):
   return self
```

This is something we use in many integrations and having it as a part of interface would help to avoid unnecessary type checks.

src/neptune/integrations/utils.py
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-__all__ = ["expect_not_an_experiment", "join_paths", "verify_type", "RunType"]
+__all__ = ["expect_not_an_experiment", "join_paths", "verify_type", "RunType", "SupportsNamespaces"]
```
twolodzko, line 16: Do we want it to be importable from `neptune.integration.utils`? Why not just use `neptune.metadata_containers.abstract` ? Even if we want to keep the `neptune.metadata_containers.abstract` private, maybe it'd make sense to have it in some kind of "types" or "interfaces" module, it's not something I'll look for in `integration.utils`.

twolodzko (approved):
I like it, but left few small comments on possible improvements.

src/neptune/integrations/utils.py
Raalsky: Sure. I've exposed it via `neptune.typing`. What do you think?

src/neptune/metadata_containers/abstract.py
Raalsky: As you wish. Done

twolodzko (approved):
LGTM

PR #1287, opened 2023-03-03
Integrations type hints

Raalsky:

- [ ] Did you **asked the docs owner** to review all the updated user-facing interfaces?


src/neptune/integrations/transformers/__init__.py
```diff
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+# mypy: disable-error-code=import
```
AleksanderWWW, line 16: Is it necessary here? We're not using `... import *` syntax here

PR #1305, opened 2023-03-21
Fix-long-filenames for docker arguments

jahaniam:
Hi Neptune team, 
We're doing a quick prototype using your library and during the prototyping, we have faced the following easy to fix issue.

If you use a container and pass very long arguments to the container (in our case the orchestrator does it and we have no control over it), the input argument is captured within `sys.argv` and later in the code it will result in exception error `OSError: [Errno 36] File name too long`

In pathlib documents it also mentions that path.is_file() can throw errors like permissions ( in our case filename being too long).

Solution:
you should use error handling. This won't usually happen in the normal execution and only happens in container-based orchestrators as one of the argv could be very long.

We're currently blocked by this issue. We'd appreciate it if you can generate a quick bugfix release from this PR. This allows us to use your library in our prototype going forward.

PS. 
1. Is there documentation on how I can locally build this package?
2. I wasn't sure about the new versioning so I didn't change the CHANGELOG. Feel free to do it as you see fit. 

Thanks,
Ali

Testcase:
```
long_str = str(10**255)
path = Path(long_str)
print(path.is_file()) -> Throws an Error filename too long. 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [UNRELEASED] neptune 1.1.1
+
+### Fixes
+- Fixed handling errors in case of too long filenames provided with `sys.argv` ([#1305](https://github.com/neptune-ai/neptune-client/pull/1305))
+
 ## neptune 1.1.0
 
 ### Features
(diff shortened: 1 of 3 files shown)


Raalsky:
Hey @jahaniam, thanks a lot for submitting this PR :wink:. I much appreciate that! 

> I wasn't sure about the new versioning so I didn't change the CHANGELOG. Feel free to do it as you see fit.

Sure, I'll update a CHANGELOG in a moment.

> Is there documentation on how I can locally build this package?

It should be as easy as `pip install -e ".[dev]"`. We're going to prepare a CONTRIBUTING guide, but we don't have something publicly available.

Raalsky (approved):

jahaniam:
@Raalsky Thanks for looking into it!

Raalsky:
Released in `1.1.1`. Thanks a lot :wink:

PR #1318, opened 2023-04-05
Support for timestamps and steps in Series constructor

AleksanderWWW:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Changes
 - Neptune objects and universal methods covered with docstrings ([#1309](https://github.com/neptune-ai/neptune-client/pull/1309))
+- Series objects accept `timestamps` and `steps` in their constructors ([#1318](https://github.com/neptune-ai/neptune-client/pull/1318))
 
 ## neptune 1.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Changes
 - Neptune objects and universal methods covered with docstrings ([#1309](https://github.com/neptune-ai/neptune-client/pull/1309))
+
+### Features
 - Series objects accept `timestamps` and `steps` in their constructors ([#1318](https://github.com/neptune-ai/neptune-client/pull/1318))
 
 ## neptune 1.1.1
--- a/src/neptune/attributes/series/file_series.py
+++ b/src/neptune/attributes/series/file_series.py
@@ -72,9 +72,6 @@ class FileSeries(Series[Val, Data, LogOperation], max_batch_size=1, operation_cl
     def _is_value_type(self, value) -> bool:
         return isinstance(value, FileSeriesVal)
 
-    def _handle_stringified_value(self, value):
-        return value.value
-
     @staticmethod
     def _get_base64_image_content(file: File) -> str:
         if file.file_type is FileType.LOCAL_FILE:
--- a/src/neptune/attributes/series/file_series.py
+++ b/src/neptune/attributes/series/file_series.py
@@ -72,6 +72,9 @@ class FileSeries(Series[Val, Data, LogOperation], max_batch_size=1, operation_cl
     def _is_value_type(self, value) -> bool:
         return isinstance(value, FileSeriesVal)
 
+    def _handle_stringified_value(self, value):
+        return value.value
+
     @staticmethod
     def _get_base64_image_content(file: File) -> str:
         if file.file_type is FileType.LOCAL_FILE:
--- a/src/neptune/attributes/series/series.py
+++ b/src/neptune/attributes/series/series.py
@@ -17,7 +17,6 @@ __all__ = ["Series"]
 
 import abc
 from typing import (
-    Any,
     Collection,
     Generic,
     Iterable,
@@ -78,10 +77,8 @@ class Series(Attribute, Generic[ValTV, DataTV, LogOperationTV]):
     def _is_value_type(self, value) -> bool:
         pass
 
-    @abc.abstractmethod
-    def _handle_stringified_value(self, value: Any) -> Any:
-        # type hints to be discussed
-        ...
+    def _handle_stringified_value(self, value):
+        return value.value
 
     def assign(self, value, wait: bool = False) -> None:
         if not self._is_value_type(value):
--- a/src/neptune/attributes/series/series.py
+++ b/src/neptune/attributes/series/series.py
@@ -16,8 +16,6 @@
 __all__ = ["Series"]
 
 import abc
-import time
-from itertools import cycle
 from typing import (
     Collection,
     Generic,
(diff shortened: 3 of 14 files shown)


src/neptune/attributes/series/series.py
```diff
@@ -116,30 +108,26 @@ def log(
...
 
-        steps = None if step is None else [step]
-        timestamps = None if timestamp is None else [timestamp] * len(value)
+        if is_stringify_value(value):
+            value = self._handle_stringified_value(value)
 
-        ops = self._get_log_operations_from_value(value, steps=steps, timestamps=timestamps)
+        if is_collection(value):
+            if step is not None and len(value) > 1:
+                raise ValueError("Collection of values are not supported for explicitly defined 'step'.")
```
normandy7, line 121: Verb to agree with "collection"

```suggestion
                raise ValueError("Collection of values is not supported for explicitly defined 'step'.")
```

normandy7 (commented):
Thanks for tagging me!

So you can do

```
s = SomeSeries(value=..., step=..., timestamp=...)
```

or

```
s = SomeSeries(values=..., steps=..., timestamps=...)
```

and everything will work fine, as long as you pass single values to the former and equal-length collections to the latter? With only `value(s)` being required

AleksanderWWW:
> Thanks for tagging me!
> 
> So you can do
> 
> ```
> s = SomeSeries(value=..., step=..., timestamp=...)
> ```
> 
> or
> 
> ```
> s = SomeSeries(values=..., steps=..., timestamps=...)
> ```
> 
> and everything will work fine, as long as you pass single values to the former and equal-length collections to the latter? With only `value(s)` being required

Hey, no, only `steps` and `timestamps` are viable keyword arguments in a Series constructor

normandy7:
> Hey, no, only `steps` and `timestamps` are viable keyword arguments in a Series constructor

Got it; was lacking the context for this change. Seems it's just an internal fix/enabler for now. Thanks

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### Changes
 - Neptune objects and universal methods covered with docstrings ([#1309](https://github.com/neptune-ai/neptune-client/pull/1309))
+- Series objects accept `timestamps` and `steps` in their constructors ([#1318](https://github.com/neptune-ai/neptune-client/pull/1318))
```
Raalsky, line 5: Isin't it a feature?

src/neptune/attributes/series/file_series.py
```diff
@@ -72,6 +72,9 @@ def _data_to_value(self, values: Iterable, **kwargs) -> Val:
     def _is_value_type(self, value) -> bool:
         return isinstance(value, FileSeriesVal)
 
+    def _handle_stringified_value(self, value):
```
Raalsky, line 75: As this method will be common for all serieses except String series, how about moving it to the abstract `Series` class?

src/neptune/internal/types/stringify_value.py
```diff
@@ -39,6 +39,12 @@ class StringifyValue:
     def __init__(self, value: Any):
         self.__value = value
 
+    def __len__(self) -> int:
```
Raalsky, line 42: I don't like it :P. `StringifyValue` may be a single value as well and it doesn't make any sense. Think of it as some kind of a "serialization", if you need to access the underlying value -> unpack it.

src/neptune/internal/types/stringify_value.py
```diff
@@ -39,6 +39,9 @@ class StringifyValue:
     def __init__(self, value: Any):
         self.__value = value
 
+    def __getitem__(self, item):
```
Raalsky, line 42: The same applies here, you cannot assume that it's an array all the time :wink:.

AleksanderWWW: Ok, but shall I leave them in `tests/unit/neptune/new/test_stringify_unsupported.py`? There we have a type hint of `str` in the constructor, so those methods will make sense for an underlying value of type string.

src/neptune/attributes/series/series.py
```diff
@@ -50,23 +47,12 @@ def __init_subclass__(cls, max_batch_size: int, operation_cls: type(LogOperation
         cls.max_batch_size = max_batch_size
         cls.operation_cls = operation_cls
 
-    def clear(self, *, wait: bool = False) -> None:
+    def clear(self, wait: bool = False) -> None:
```
Raalsky, line 50: Why do we need to remove the `*` here?

tests/unit/neptune/new/attributes/series/test_float_series.py
```diff
@@ -51,3 +52,7 @@ def test_log(self):
             values = list(var.fetch_values()["value"].array)
             expected = list(range(0, 5000))
             self.assertEqual(len(set(expected)), len(set(values)))
+
+    def test_handle_stringified_value(self):
```
Raalsky, line 56: This looks a little bit weird as we're mocking some internal value. Isin't it already covered in stringify value related test?

tests/unit/neptune/new/test_stringify_unsupported.py
```diff
@@ -44,6 +44,12 @@ class Obj:
     def __init__(self, name: str = "A"):
         self._name = name
 
+    def __len__(self) -> int:
```
Raalsky, line 47: Use a list of objects instead in tests. `Obj` should be just a single value and really similar to dataclass (or even we can made it a dataclass, but no need to do to my knowledge)

src/neptune/internal/types/stringify_value.py
Raalsky: No, `Obj` should be a really simple value like `int` or `string` but not a built-in :wink:. `len` and index-based access looks like an array/list.

src/neptune/attributes/series/series.py
AleksanderWWW: By mistake! My bad

src/neptune/attributes/series/string_series.py
```diff
@@ -65,20 +65,31 @@ def _get_log_operations_from_value(
...
         if kwargs:
             logger.warning("Warning: unexpected arguments (%s) in StringSeries", kwargs)
 
-        return StringSeriesVal(values)
+        return StringSeriesVal(values, steps=steps, timestamps=timestamps)
 
     def _is_value_type(self, value) -> bool:
         return isinstance(value, StringSeriesVal)
 
+    def _handle_stringified_value(self, value):
```
Raalsky, line 85: Just a code style suggestion:
```python
    def _handle_stringified_value(self, value):
        if is_collection(value.value):
            return list(map(str, value.value))
        return str(value.value)
```

Raalsky:
E2E tests: https://github.com/neptune-ai/neptune-client/actions/runs/4691318134

Raalsky (approved):

PR #1333, opened 2023-04-28
Invite users to the workspace programmatically

AleksanderWWW:

### DIFF
--- a/src/neptune/management/internal/api.py
+++ b/src/neptune/management/internal/api.py
@@ -578,7 +578,7 @@ def invite_to_workspace(
     role: WorkspaceMemberRole = WorkspaceMemberRole.MEMBER,
     add_to_all_projects: bool = False,
 ) -> None:
-    """Created invitation to Neptune workspace.
+    """Creates invitation to Neptune workspace.
 
     Args:
         username: username of the user to invite.
--- a/src/neptune/management/internal/api.py
+++ b/src/neptune/management/internal/api.py
@@ -581,27 +581,28 @@ def invite_to_workspace(
     """Creates invitation to Neptune workspace.
 
     Args:
-        username: username of the user to invite.
-        email: email of the user to invite.
-            Note: at least one of the above parameters are needed.
-            If neither the username nor the email is passed, will raise ValueError.
+        username: Username of the user to invite.
+        email: Email address of the user to invite.
+            Note: at least one of the above parameters is needed.
             If both are filled email will be ignored.
         workspace: Name of your Neptune workspace.
         api_token: Account's API token.
             If None, the value of the NEPTUNE_API_TOKEN environment variable is used.
             Note: To keep your token secure, use the NEPTUNE_API_TOKEN environment variable rather than placing your
             API token in plain text in your source code.
-        role: the workspace role that is to be granted to the invited user.
-        add_to_all_projects: whether to add the user to all projects in the workspace.
-    Returns:
-        None
+        role: The workspace role that is to be granted to the invited user.
+        You can choose between the following values:
+        - Administrator: `WorkspaceMemberRole.ADMIN`
+        - Member: `WorkspaceMemberRole.MEMBER`
+        add_to_all_projects: Whether to add the user to all projects in the workspace.
 
     Example:
         >>> from neptune import management
+        >>> from management import WorkspaceMemberRole
         >>> management.invite_to_workspace(
         ...     username="user",
         ...     workspace="ml-team",
-        ...     role=management.WorkspaceMemberRole.ADMIN
+        ...     role=WorkspaceMemberRole.ADMIN,
         ... )
 
     You may also want to check the management API reference:
--- a/src/neptune/management/internal/api.py
+++ b/src/neptune/management/internal/api.py
@@ -629,15 +629,7 @@ def invite_to_workspace(
         raise ValueError("Neither `username` nor `email` arguments filled. At least one needs to be passed")
 
     if isinstance(role, str):
-        if role.lower() not in {"admin", "member"}:
-            raise ValueError(f"Unrecognized role: {role}")
-
-        role = role.lower()
-
-        role = role if role != "admin" else "owner"
-
-    if isinstance(role, WorkspaceMemberRole):
-        role = role.value
+        role = WorkspaceMemberRole(role)
 
     params = {
         "newOrganizationInvitations": {
@@ -645,7 +637,7 @@ def invite_to_workspace(
                 {
                     "invitee": invitee,
                     "invitationType": invitation_type,
-                    "roleGrant": role,
+                    "roleGrant": role.to_api(),
                     "addToAllProjects": add_to_all_projects,
                 }
             ],
--- a/src/neptune/management/internal/api.py
+++ b/src/neptune/management/internal/api.py
@@ -575,16 +575,17 @@ def invite_to_workspace(
     email: Optional[str] = None,
     workspace: str,
     api_token: Optional[str] = None,
-    role: WorkspaceMemberRole = WorkspaceMemberRole.MEMBER,
+    role: Union[WorkspaceMemberRole, str] = WorkspaceMemberRole.MEMBER,
     add_to_all_projects: bool = False,
 ) -> None:
     """Creates invitation to Neptune workspace.
 
     Args:
-        username: Username of the user to invite.
-        email: Email address of the user to invite.
-            Note: at least one of the above parameters is needed.
-            If both are filled email will be ignored.
+        username: username of the user to invite.
+        email: email of the user to invite.
+            Note: at least one of the above parameters are needed.
+            If neither the username nor the email is passed, will raise ValueError.
+            If both are filled, will raise ValueError.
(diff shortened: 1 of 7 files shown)


src/neptune/management/internal/api.py
```diff
@@ -563,6 +568,85 @@ def get_workspace_service_account_list(workspace: str, *, api_token: Optional[st
...
+    add_to_all_projects: bool = False,
+) -> None:
+    """Creates invitation to Neptune workspace.
+
+    Args:
+        username: username of the user to invite.
+        email: email of the user to invite.
+            Note: at least one of the above parameters are needed.
+            If neither the username nor the email is passed, will raise ValueError.
+            If both are filled email will be ignored.
```
normandy7, line 588:
```suggestion
        username: Username of the user to invite.
        email: Email address of the user to invite.
            Note: at least one of the above parameters is needed.
            If both are filled email will be ignored.
```

src/neptune/management/internal/api.py
```diff
@@ -563,6 +568,85 @@ def get_workspace_service_account_list(workspace: str, *, api_token: Optional[st
...
+        workspace: Name of your Neptune workspace.
+        api_token: Account's API token.
+            If None, the value of the NEPTUNE_API_TOKEN environment variable is used.
+            Note: To keep your token secure, use the NEPTUNE_API_TOKEN environment variable rather than placing your
+            API token in plain text in your source code.
+        role: the workspace role that is to be granted to the invited user.
+        add_to_all_projects: whether to add the user to all projects in the workspace.
+    Returns:
+        None
```
normandy7, line 597:
```suggestion
        role: The workspace role that is to be granted to the invited user.
        You can choose between the following values:
        - Administrator: `WorkspaceMemberRole.ADMIN`
        - Member: `WorkspaceMemberRole.MEMBER`
        add_to_all_projects: Whether to add the user to all projects in the workspace.
```

src/neptune/management/internal/api.py
```diff
@@ -563,6 +568,85 @@ def get_workspace_service_account_list(workspace: str, *, api_token: Optional[st
...
+
+    Example:
+        >>> from neptune import management
+        >>> management.invite_to_workspace(
+        ...     username="user",
+        ...     workspace="ml-team",
+        ...     role=management.WorkspaceMemberRole.ADMIN
```
normandy7, line 604:
```suggestion
        >>> from neptune import management
        >>> from management import WorkspaceMemberRole
        >>> management.invite_to_workspace(
        ...     username="user",
        ...     workspace="ml-team",
        ...     role=WorkspaceMemberRole.ADMIN,
```

normandy7 (commented):
For the role, can the end-user pass `"admin"`/`"member"`, or you need to use the enums?

AleksanderWWW:
> For the role, can the end-user pass `"admin"`/`"member"`, or you need to use the enums?

The spec wasn't clear about this, so in this design I decided that the enum will be required.
This way the code is cleaner imo and there's no need to check separately if the value passed by a user is correct

src/neptune/management/internal/api.py
```diff
@@ -563,6 +568,86 @@ def get_workspace_service_account_list(workspace: str, *, api_token: Optional[st
...
 
 
+@with_api_exceptions_handler
+def invite_to_workspace(
+    *,
+    username: Optional[str] = None,
+    email: Optional[str] = None,
+    workspace: str,
+    api_token: Optional[str] = None,
+    role: WorkspaceMemberRole = WorkspaceMemberRole.MEMBER,
```
Raalsky, line 578: From the spec:
> No, they still can pass raw string value. It's about typing the argument :)

I think we should support raw `string` values as well.

src/neptune/management/internal/api.py
```diff
@@ -563,6 +568,86 @@ def get_workspace_service_account_list(workspace: str, *, api_token: Optional[st
...
+    https://docs.neptune.ai/api/management
+    """
+    verify_type("workspace", workspace, str)
+    verify_type("role", role, WorkspaceMemberRole)
+    verify_type("add_to_all_projects", add_to_all_projects, bool)
+    verify_type("username", username, (str, type(None)))
+    verify_type("email", email, (str, type(None)))
+    verify_type("api_token", api_token, (str, type(None)))
+
+    if username:
```
Raalsky, line 618: We should raise an exception in case of setting both `username` and `email`.

tests/unit/neptune/new/internal/backends/test_hosted_client.py
```diff
@@ -115,6 +116,29 @@ def test_project_listing(self, swagger_client_factory):
...
+    def test_invite_to_workspace(self, swagger_client_factory):
+        swagger_client = self._get_swagger_client_mock(swagger_client_factory)
+
+        invite_to_workspace(
+            username="tester1",
+            workspace="org2",
+            api_token=API_TOKEN,
+        )
+
+        self.assertEqual(swagger_client.api.createOrganizationInvitations.call_count, 1)
```
Raalsky, line 128: Let's cleanup a little bit here. Create a structure like `given`/`when`/`then` (could be separated with comments as in case of other tests). Instead of `call_args.kwargs` you should be able to call something like `called_once_with(...)` or something like that.

Raalsky (changes requested):

src/neptune/management/internal/utils.py
```diff
@@ -50,3 +51,8 @@ def normalize_project_name(name: str, workspace: Optional[str] = None):
     extracted_workspace_name, extracted_project_name = extract_project_and_workspace(name=name, workspace=workspace)
 
     return f"{extracted_workspace_name}/{extracted_project_name}"
+
+
+class WorkspaceMemberRole(Enum):
+    MEMBER = "member"
+    ADMIN = "admin"
```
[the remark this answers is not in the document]
normandy7: That is a project role.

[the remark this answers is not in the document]
Raalsky: I know it's for workspaces and from the user perspective/UI it should be `Admin` but we're passing this value in this PR directly to the our backend API. On the API level I think it's `owner` instead of `admin` and it should be mapped in-between this enum and calling an API endpoint.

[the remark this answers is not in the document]
AleksanderWWW: Are you sure about that? I was able to call API with `Admin` value and then confirmed that the correct invitation was created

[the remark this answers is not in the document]
Raalsky: Yes, I'm quite sure. UI is making a request with `owner` and I see that backend only allows `member`/`owner`.
Screenshot 2023-05-08 at 14 23 58

src/neptune/management/internal/api.py
```diff
@@ -563,6 +568,101 @@ def get_workspace_service_account_list(workspace: str, *, api_token: Optional[st
...
+        invitee = username
+        invitation_type = "user"
+    elif email:
+        invitee = email
+        invitation_type = "emailRecipient"
+    else:
+        raise ValueError("Neither `username` nor `email` arguments filled. At least one needs to be passed")
+
+    if isinstance(role, str):
+        if role.lower() not in {"admin", "owner", "member"}:
```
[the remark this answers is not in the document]
Raalsky: From user perspective we only allow `WorkspaceRole.MEMBER`/`WorksapceRole.ADMIN`/`admin` (str) / `member` (str)

src/neptune/management/internal/utils.py
```diff
@@ -50,3 +51,8 @@ def normalize_project_name(name: str, workspace: Optional[str] = None):
     extracted_workspace_name, extracted_project_name = extract_project_and_workspace(name=name, workspace=workspace)
 
     return f"{extracted_workspace_name}/{extracted_project_name}"
+
+
+class WorkspaceMemberRole(Enum):
```
Raalsky, line 56: add a method `to_api` that translates to `member`/`owner` instead.

src/neptune/management/internal/api.py
```diff
@@ -563,6 +568,101 @@ def get_workspace_service_account_list(workspace: str, *, api_token: Optional[st
...
+    if username:
+        invitee = username
+        invitation_type = "user"
+    elif email:
+        invitee = email
+        invitation_type = "emailRecipient"
+    else:
+        raise ValueError("Neither `username` nor `email` arguments filled. At least one needs to be passed")
+
+    if isinstance(role, str):
```
Raalsky, line 631: This whole logic could be replaced with calling `Enum` casting directly:
```
try:
    casted_role = WorkspaceRole(role)
except ValueError as e:
    raise ValueError(f"Unrecognized role: {role}") from e
```

AleksanderWWW: Without handling this exception what is being raised looks actually quite good already imo
`ValueError: '{value}' is not a valid WorkspaceMemberRole`

Do you think it can stay like this or is it important to have this custom error message?

Raalsky: It's ok

Raalsky (approved):

PR #1336, opened 2023-05-05
Fix crashing syncing thread when logging big integers

aniezurawski:


src/neptune/attributes/atoms/integer.py
```diff
@@ -45,5 +53,17 @@ def assign(self, value: typing.Union[IntegerVal, float, int], *, wait: bool = Fa
         if not isinstance(value, IntegerVal):
             value = IntegerVal(value)
 
+        if value.value > Integer.MAX_32_BIT_INT or value.value < Integer.MIN_32_BIT_INT:
```
AleksanderWWW, line 56: Slightly more readable syntax:
`Integer.MIN_32_BIT_INT < value.value < Integer.MAX_32_BIT_INT`

AleksanderWWW (approved):
LGTM. One optional thing to change

src/neptune/attributes/atoms/integer.py
```diff
@@ -45,5 +50,17 @@ def assign(self, value: typing.Union[IntegerVal, float, int], *, wait: bool = Fa
...
+            logger.warning(
+                "WARNING: The value you're trying to log is outside the range of 32-bit integers "
+                "(%s to %s) and will be skipped. "
+                "We'll support 64-bit integers in the future. "
+                'For now, try logging the value as a float instead: run["field"] = float(%s)',
+                Integer.MIN_32_BIT_INT,
+                Integer.MAX_32_BIT_INT,
+                value.value,
+            )
+            return
```
Raalsky, line 63: `return None` is more convenient.

Raalsky (changes requested):

src/neptune/attributes/atoms/integer.py
aniezurawski: Is it, really? For function which is meant to be a procedure? Seems misleading.

Raalsky: Ok. ChatGPT agrees with you :P.

Raalsky (approved):

Raalsky (approved):

PR #1342, opened 2023-05-10
Update role arg of invite_to_workspace() docstring

normandy7:
Update `role` arg and example for `invite_to_workspace()` docstring.

### DIFF
--- a/src/neptune/management/internal/api.py
+++ b/src/neptune/management/internal/api.py
@@ -592,18 +592,15 @@ def invite_to_workspace(
             Note: To keep your token secure, use the NEPTUNE_API_TOKEN environment variable rather than placing your
             API token in plain text in your source code.
         role: The workspace role that is to be granted to the invited user.
-        You can choose between the following values:
-        - Administrator: `WorkspaceMemberRole.ADMIN`
-        - Member: `WorkspaceMemberRole.MEMBER`
+            You can choose between the following values: "admin" and "member".
         add_to_all_projects: Whether to add the user to all projects in the workspace.
 
     Example:
         >>> from neptune import management
-        >>> from management import WorkspaceMemberRole
         >>> management.invite_to_workspace(
         ...     username="user",
         ...     workspace="ml-team",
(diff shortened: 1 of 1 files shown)


src/neptune/management/internal/api.py
```diff
@@ -592,18 +592,15 @@ def invite_to_workspace(
             Note: To keep your token secure, use the NEPTUNE_API_TOKEN environment variable rather than placing your
             API token in plain text in your source code.
         role: The workspace role that is to be granted to the invited user.
-        You can choose between the following values:
-        - Administrator: `WorkspaceMemberRole.ADMIN`
-        - Member: `WorkspaceMemberRole.MEMBER`
+            You can choose between the following values: "admin" and "member".
```
AleksanderWWW, line 595: Don't we also want to encourage using the enum?

src/neptune/management/internal/api.py
```diff
@@ -592,18 +592,15 @@ def invite_to_workspace(
...
         add_to_all_projects: Whether to add the user to all projects in the workspace.
 
     Example:
         >>> from neptune import management
-        >>> from management import WorkspaceMemberRole
         >>> management.invite_to_workspace(
         ...     username="user",
         ...     workspace="ml-team",
-        ...     role=WorkspaceMemberRole.ADMIN,
+        ...     role="admin",
```
AleksanderWWW, line 603: Shall we also somehow underscore that this is case sensitive? So e.g. `AdMIn` or `Admin` etc. won't work

src/neptune/management/internal/api.py
normandy7: Yes, my plan is to do it in the docs. Here I think it's better to showcase the most straightforward option.

src/neptune/management/internal/api.py
normandy7: Maybe in the docs. But we explicitly state the accepted values, so IMO that should be enough :)

AleksanderWWW (approved):

PR #1343, opened 2023-05-10
Bp/project visibility restricted

BartoszPrusak:
Internal fix for correct handling of the `VISIBILITY_RESTRICTED` API error.

### DIFF
--- a/tests/unit/neptune/new/internal/backends/test_hosted_client.py
+++ b/tests/unit/neptune/new/internal/backends/test_hosted_client.py
@@ -400,7 +400,7 @@ class TestHostedClient(unittest.TestCase, BackendTestMixin):
         )
 
         # then:
-        with self.assertRaisesRegex(ProjectPrivacyRestrictedException, ".*priv visibility.*"):
+        with self.assertRaisesRegex(ProjectPrivacyRestrictedException, '.*"priv" visibility.*'):
             create_project(name="org/proj", key="PRJ", visibility="priv", api_token=API_TOKEN)
 
     def test_create_project_private_not_allowed_no_details(self, swagger_client_factory):


src/neptune/internal/backends/swagger_client_wrapper.py
```diff
@@ -28,66 +29,65 @@
...
+            ProjectPrivacyRestrictedException,
+            ProjectsLimitReached,
+        )
+
         try:
             body = response.json() or dict()
         except Exception:
             body = {}
 
+        error_processors: dict[str, Callable[[dict], ManagementOperationFailure]] = {
```
aniezurawski, line 53: I don't think all these exception inherit from `ManagementOperationFailure`.

src/neptune/internal/backends/swagger_client_wrapper.py
```diff
@@ -28,66 +29,65 @@
...
+            ),
+            "LIMIT_OF_PROJECTS_REACHED": lambda _: ProjectsLimitReached(),
+            "PROJECT_KEY_COLLISION": lambda response_body: ProjectKeyCollision(
+                key=response_body.get("key", "<unknown key>")
+            ),
+            "PROJECT_KEY_INVALID": lambda response_body: ProjectKeyInvalid(
+                key=response_body.get("key", "<unknown key>"),
+                reason=response_body.get("reason", "Unknown reason"),
+            ),
+            "PROJECT_NAME_COLLISION": lambda response_body: ProjectKeyCollision(
```
aniezurawski, line 72: `ProjectNameCollision`

src/neptune/internal/backends/swagger_client_wrapper.py
```diff
@@ -28,66 +29,65 @@
...
+                key=response_body.get("key", "<unknown key>")
+            ),
+            "PROJECT_NAME_INVALID": lambda response_body: ProjectNameCollision(
```
aniezurawski, line 75: `ProjectNameInvalid`

src/neptune/internal/backends/swagger_client_wrapper.py
```diff
@@ -28,66 +29,65 @@
...
+                name=response_body.get("name", "<unknown name>")
+            ),
+            "VISIBILITY_RESTRICTED": lambda response_body: ProjectPrivacyRestrictedException(
+                requested=response_body.get("requestedValue", "the selected"),
+                allowed=response_body.get("allowedValues"),
+            ),
+            "WORKSPACE_IN_READ_ONLY_MODE": lambda response_body: NeptuneLimitExceedException(
+                reason=response_body.get("title", "Unknown reason")
+            ),
+        }
```
aniezurawski, line 85: @Raalsky @BartoszPrusak how do you feel about testing these not-happy paths e2e? Do you think it's worth the effort?

aniezurawski (changes requested):

src/neptune/internal/backends/swagger_client_wrapper.py
Raalsky: In the long term, I would like to add some kind of integration tests that will mimic the possible outcomes of a backend API (mocking responses of `requests`). That would be a much better place for such not-happy paths and e2e could be a bare minimum. Currently -> It's ok but there is no big need for it.

src/neptune/internal/backends/swagger_client_wrapper.py
BartoszPrusak: Right, I forgot about `NeptuneException`s.
Fixed.

src/neptune/internal/backends/swagger_client_wrapper.py
BartoszPrusak: Fixed

src/neptune/management/exceptions.py
```diff
@@ -178,3 +178,16 @@ class WorkspaceOrUserNotFound(ManagementOperationFailure):
 class UserAlreadyInvited(ManagementOperationFailure):
     code = 24
     description = "User '{user}' has already been invited to the workspace '{workspace}'."
+
+
+class ProjectPrivacyRestrictedException(ManagementOperationFailure):
+    code = 25
+    description = "Cannot set {requested} visibility for project. {followup}"
```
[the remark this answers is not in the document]
BartoszPrusak: I made a new commit in this MR with simplified message. Hit me up if you need to know more :)

src/neptune/internal/backends/swagger_client_wrapper.py
BartoszPrusak: I added an e2e test for the error message I introduced, but don't plan on adding more.
I talked to Raalsky and the non-happy paths will be tested in a different environment.

aniezurawski (approved):

PR #1345, opened 2023-05-12
Automatically track environment dependencies

AleksanderWWW:

### DIFF
--- a/src/neptune/internal/utils/dependency_tracking.py
+++ b/src/neptune/internal/utils/dependency_tracking.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
 
 class DependencyTrackingStrategy(ABC):
     @abstractmethod
-    def log_dependencies(self, run) -> None:
+    def log_dependencies(self, run: "Run") -> None:
         ...
 
 
--- a/src/neptune/internal/utils/dependency_tracking.py
+++ b/src/neptune/internal/utils/dependency_tracking.py
@@ -10,38 +10,38 @@ from abc import (
     ABC,
     abstractmethod,
 )
-from typing import Union
+from typing import (
+    TYPE_CHECKING,
+    Union,
+)
 
 from neptune.types import File
+
+if TYPE_CHECKING:
+    from neptune import Run
 
 
 class DependencyTrackingStrategy(ABC):
     @abstractmethod
-    def track_dependencies(self) -> None:
+    def log_dependencies(self, run) -> None:
         ...
 
 
 class InferDependenciesStrategy(DependencyTrackingStrategy):
-    def __init__(self, run):
-        self._run = run
-
-    def track_dependencies(self) -> None:
+    def log_dependencies(self, run: "Run") -> None:
         try:
             dependencies_str = subprocess.check_output(["pipreqs", "--print", "."]).decode("utf-8")
         except subprocess.SubprocessError:
             return
 
         if dependencies_str:
-            self._run["source_code/requirements"].upload(File.from_content(dependencies_str))
+            run["source_code/requirements"].upload(File.from_content(dependencies_str))
 
 
 class FileDependenciesStrategy(DependencyTrackingStrategy):
-    def __init__(self, run, path: Union[str, os.PathLike]):
-        self._run = run
+    def __init__(self, path: Union[str, os.PathLike]):
         self._path = path
 
-    def track_dependencies(self) -> None:
-        if not os.path.isfile(self._path):
-            return
-
-        self._run["source_code/files"].upload_files(self._path)
+    def log_dependencies(self, run: "Run") -> None:
+        if os.path.isfile(self._path):
+            run["source_code/files"].upload_files(os.path.basename(self._path))
--- a/src/neptune/internal/utils/dependency_tracking.py
+++ b/src/neptune/internal/utils/dependency_tracking.py
@@ -1,0 +1,47 @@
+__all__ = [
+    "DependencyTrackingStrategy",
+    "InferDependenciesStrategy",
+    "FileDependenciesStrategy",
+]
+
+import os
+import subprocess
+from abc import (
+    ABC,
+    abstractmethod,
+)
+from typing import Union
+
+from neptune.types import File
+
+
+class DependencyTrackingStrategy(ABC):
+    @abstractmethod
+    def track_dependencies(self) -> None:
+        ...
+
+
+class InferDependenciesStrategy(DependencyTrackingStrategy):
+    def __init__(self, run):
+        self._run = run
+
+    def track_dependencies(self) -> None:
+        try:
+            dependencies_str = subprocess.check_output(["pipreqs", "--print", "."]).decode("utf-8")
+        except subprocess.SubprocessError:
+            return
+
+        if dependencies_str:
+            self._run["source_code/requirements"].upload(File.from_content(dependencies_str))
+
+
+class FileDependenciesStrategy(DependencyTrackingStrategy):
+    def __init__(self, run, path: Union[str, os.PathLike]):
+        self._run = run
+        self._path = path
+
+    def track_dependencies(self) -> None:
+        if not os.path.isfile(self._path):
+            return
+
+        self._run["source_code/files"].upload_files(self._path)
--- a/src/neptune/metadata_containers/run.py
+++ b/src/neptune/metadata_containers/run.py
@@ -480,11 +480,12 @@ class Run(MetadataContainer):
 
         else:
             # uploading dependencies file provided by the user
-            if not os.path.exists(self._dependencies):
+            if not os.path.exists(self._dependencies) or not os.path.isfile(self._dependencies):
                 from neptune.internal.utils.logger import logger
 
                 logger.error(
-                    f"File {self._dependencies} not found. The dependencies will not be tracked in the current run."
+                    f"File '{self._dependencies}' not found or is not a file. The dependencies will not be tracked in "
+                    f"the current run."
                 )
                 return
 
--- a/src/neptune/metadata_containers/run.py
+++ b/src/neptune/metadata_containers/run.py
@@ -84,6 +84,7 @@ from neptune.internal.utils.traceback_job import TracebackJob
 from neptune.internal.websockets.websocket_signals_background_job import WebsocketSignalsBackgroundJob
 from neptune.metadata_containers import MetadataContainer
 from neptune.types import (
+    File,
     GitRef,
     StringSeries,
 )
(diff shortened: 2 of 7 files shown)


src/neptune/metadata_containers/run.py
```diff
@@ -218,6 +220,9 @@ def __init__(
                 If None, Neptune looks for a repository in the path of the script that is executed.
                 To specify a different location, set to GitRef(repository_path="path/to/repo").
                 To turn off Git tracking for the run, set to GitRef.DISABLED.
+            dependencies: If you want to track the project dependencies pass a path to your dependency file.
+                If None, no dependency file will be uploaded.
+                If 'infer' is passed, Neptune will generate and upload a requirements file using `pipreqs`.
```
normandy7, line 225:
```suggestion
            dependencies: To track the project dependencies, pass a path to your dependency file.
                If None, no dependency file is uploaded.
                If you pass `"infer"`, Neptune uses [pipreqs](https://pypi.org/project/pipreqs) to generate and upload
                a requirements file. The file is named "neptune-tracked-dependencies.txt".
```

normandy7 (commented):
Looks good otherwise.

normandy7 (approved):

src/neptune/metadata_containers/run.py
```diff
@@ -451,6 +463,35 @@ def _write_initial_attributes(self):
             # upload default sources ONLY if creating a new run
             upload_source_code(source_files=self._source_files, run=self)
 
+    def _track_dependencies(self) -> None:
+        if self._dependencies == "infer":
+            # using pipreqs here
+            file_name = "neptune-tracked-dependencies.txt"
+            status_code = subprocess.check_call(["pipreqs", "--force", "--savepath", file_name, "."])
```
Raalsky, line 470: How about using this package directly instead of console script? https://github.com/bndr/pipreqs/blob/master/pipreqs/pipreqs.py

src/neptune/metadata_containers/run.py
```diff
@@ -451,6 +463,35 @@ def _write_initial_attributes(self):
             # upload default sources ONLY if creating a new run
             upload_source_code(source_files=self._source_files, run=self)
 
+    def _track_dependencies(self) -> None:
+        if self._dependencies == "infer":
+            # using pipreqs here
+            file_name = "neptune-tracked-dependencies.txt"
```
Raalsky, line 469: I would like to see it via uploading an in-memory file instead of using some temp file on disk.

src/neptune/metadata_containers/run.py
```diff
@@ -363,6 +373,9 @@ def __init__(
 
         super().__init__(project=project, api_token=api_token, mode=mode, flush_period=flush_period, proxies=proxies)
 
+        if self._dependencies:
```
Raalsky, line 376: It should be placed in `_write_initial_attributes`

src/neptune/metadata_containers/run.py
```diff
@@ -451,6 +464,35 @@ def _write_initial_attributes(self):
...
+
+            dependencies_str = proc.communicate()[0].decode(encoding="utf-8")
+
+            if not dependencies_str:
+                from neptune.internal.utils.logger import logger
+
+                logger.error("Could not generate requirements. Call to 'pipreqs' returned an empty string.")
+                return
+
+            self["source_code/files/requirements"].upload(File.from_content(dependencies_str))
```
Raalsky, line 481: I think it could be problematic if we're going to upload some source files under `source_code/files`. Did you check that?

normandy7 (dismissed):

src/neptune/metadata_containers/run.py
```diff
@@ -451,6 +461,43 @@ def _write_initial_attributes(self):
             # upload default sources ONLY if creating a new run
             upload_source_code(source_files=self._source_files, run=self)
 
+        if self._dependencies:
+            self._track_dependencies()
+
+    def _track_dependencies(self) -> None:
```
Raalsky, line 467: Let's extract this logic to separate class that would be easier to test. Here you could only check that it calls some class methods correctly and in another test suite ensure that it works for `infer`/path-specific.

src/neptune/metadata_containers/run.py
```diff
@@ -451,6 +461,43 @@ def _write_initial_attributes(self):
...
+        if self._dependencies == "infer":
+            from neptune.internal.utils.logger import logger
+
+            logger.info("INFO: Calling pipreqs to generate requirements based on project imports.")
```
Raalsky, line 471: I'm not sure that we need it or anything like that.

src/neptune/metadata_containers/run.py
```diff
@@ -451,6 +461,43 @@ def _write_initial_attributes(self):
...
+            # using pipreqs here
+            try:
+                proc = subprocess.Popen(["pipreqs", "--print", "."], stdout=subprocess.PIPE)
```
Raalsky, line 474: Check this: https://stackoverflow.com/a/4760517
This current solution is probably a bit too complicated and low-level

src/neptune/metadata_containers/run.py
```diff
@@ -451,6 +461,43 @@ def _write_initial_attributes(self):
...
+            except subprocess.SubprocessError as exp:
+                logger.error("ERROR: {}".format(exp))
+                dependencies_str = ""
+
+            if not dependencies_str:
+
+                logger.error("ERROR: Could not generate requirements. Call to 'pipreqs' returned an empty string.")
+                return
+
+            self["source_code/requirements.txt"].upload(File.from_content(dependencies_str))
```
Raalsky, line 486: `from_content` has a key parameter `extension` that could be set to `txt`. Ensure that this will not produce `requirements.txt.txt` file after downloading. Probably we should rename the namespace to just `requirements`.

src/neptune/metadata_containers/run.py
```diff
@@ -451,6 +461,43 @@ def _write_initial_attributes(self):
...
+
+        else:
+            # uploading dependencies file provided by the user
+            if not os.path.exists(self._dependencies) or not os.path.isfile(self._dependencies):
+                from neptune.internal.utils.logger import logger
```
Raalsky, line 491: Always prefer the top level imports (at the beginning of a file).

src/neptune/metadata_containers/run.py
AleksanderWWW: [from docs of `from_content`]
extension – Extension of the created file. File will be used for interpreting the type of content for visualization. If None it will be bin for binary content and txt for text content. Defaults to None.

Since we upload a string, then I guess there's no need to add it explicitly and just rename namespace to requirements

src/neptune/internal/utils/dependency_tracking.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class DependencyTrackingStrategy(ABC):
+    @abstractmethod
+    def track_dependencies(self) -> None:
+        ...
+
+
+class InferDependenciesStrategy(DependencyTrackingStrategy):
+    def __init__(self, run):
```
Raalsky, line 25: Type hints?

src/neptune/internal/utils/dependency_tracking.py
```diff
@@ -0,0 +1,47 @@
...
+    abstractmethod,
+)
+from typing import Union
+
+from neptune.types import File
+
+
+class DependencyTrackingStrategy(ABC):
+    @abstractmethod
+    def track_dependencies(self) -> None:
```
Raalsky, line 20: -> `log_dependencies`?

src/neptune/internal/utils/dependency_tracking.py
Raalsky: + `run` is common for both strategies, maybe move it to parameter of `track_dependencies`?

tests/unit/neptune/new/internal/utils/test_dependency_tracking.py
```diff
@@ -0,0 +1,46 @@
+import unittest
```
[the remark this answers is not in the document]
AleksanderWWW: Added for inferring dependencies. How do we handle dummy files in test suits? Or is there another way to test uploading dependency file in e2e tests?

src/neptune/internal/utils/dependency_tracking.py
AleksanderWWW: Added with the TYPE_CHECKING pattern to avoid circular imports

src/neptune/internal/utils/dependency_tracking.py
```diff
@@ -0,0 +1,47 @@
...
+
+class InferDependenciesStrategy(DependencyTrackingStrategy):
+    def log_dependencies(self, run: "Run") -> None:
+        try:
+            dependencies_str = subprocess.check_output(["pipreqs", "--print", "."]).decode("utf-8")
+        except subprocess.SubprocessError:
+            return
+
+        if dependencies_str:
+            run["source_code/requirements"].upload(File.from_content(dependencies_str))
```
Raalsky, line 38: Shouldn't we add `extension='.txt'`?

src/neptune/internal/utils/dependency_tracking.py
```diff
@@ -0,0 +1,47 @@
...
+
+from neptune.types import File
+
+if TYPE_CHECKING:
+    from neptune import Run
+
+
+class DependencyTrackingStrategy(ABC):
+    @abstractmethod
+    def log_dependencies(self, run) -> None:
```
Raalsky, line 26: `run: "Run"`

tests/e2e/standard/test_init.py
```diff
@@ -92,6 +94,36 @@ def test_git_disabled(self, environment):
         exp.sync()
         assert not exp.exists("source_code/git")
 
+    def test_infer_dependencies(self, environment):
+        exp = neptune.init_run(
```
Raalsky, line 98: This will produce a atexit error as run wasn't stopped. Use contextmanager syntax here (`with neptune.init_run(....) as exp:`

src/neptune/metadata_containers/run.py
Raalsky: Ok

src/neptune/internal/utils/dependency_tracking.py
AleksanderWWW: I think it's the default if the argument of `from_content` is a string

tests/e2e/standard/test_init.py
AleksanderWWW: Shall I correct it in other tests that were already here, that don't implement it like that? Like `test_git_disabled`

Raalsky: I would be grateful.

Raalsky (approved):

PR #1350, opened 2023-05-23
Track uncommitted changes (diff/patch)

AleksanderWWW:


normandy7 (approved):

tests/e2e/standard/test_init.py
```diff
@@ -122,6 +124,24 @@ def test_upload_dependency_file(self, environment):
             with zipped.open(filename, "r") as file:
                 assert file.read().decode(encoding="utf-8") == "some-dependency==1.0.0"
 
+    def test_tracking_uncommitted_changes(self, environment):
```
Raalsky, line 127: Could you please add similar initialization setup to other tests that was skipped? For instance check line 76 with ```@pytest.mark.skip("In CI we are running this in root directory with git repository")```

AleksanderWWW: Sure

AleksanderWWW: Running `grep -r -n pytest.mark.skip --include="*.py" tests/e2e` only found this one relevant test. E2e seem to be passing after the fix

src/neptune/metadata_containers/run.py
```diff
@@ -472,6 +475,11 @@ def _write_initial_attributes(self):
 
             dependency_strategy.log_dependencies(run=self)
 
+        track_uncommitted_changes(
+            git_ref=self._git_ref or GitRef(),
```
Raalsky, line 479: What if someone doesn't want to track this? Why are we not taking into account the `GitRef.Disabled`?

tests/e2e/standard/test_init.py
```diff
@@ -71,16 +73,22 @@ def test_git_client_repository(self, environment):
             exp.sync()
             assert exp.exists("source_code/git")
 
-    @pytest.mark.skip("In CI we are running this in root directory with git repository")
     def test_git_default(self, environment):
+        try:
```
Raalsky, line 77: This logic could be moved to some common pytest fixture

tests/e2e/standard/test_init.py
```diff
@@ -122,6 +130,24 @@ def test_upload_dependency_file(self, environment):
...
+
+        with open("some_file.txt", "w") as fp:
+            fp.write("some-content\n")
+
+        repo.git.add("some_file.txt")
+
+        assert repo.is_dirty()
+        with neptune.init_run(project=environment.project) as run:
+            run.sync()
+            assert run.exists("source_code/diff")
```
Raalsky, line 149: How about checking the content of a diff? We prepared a `some_file.txt` so it shouldn't be that hard to achieve :wink:

tests/unit/neptune/new/client/test_run.py
```diff
@@ -231,3 +232,12 @@ def test_infer_dependency_strategy_called(self, mock_infer_method):
     def test_file_dependency_strategy_called(self, mock_file_method):
         with init_run(mode="debug", dependencies="some_file_path.txt"):
             mock_file_method.assert_called_once()
+
+    @patch("neptune.metadata_containers.run.track_uncommitted_changes")
+    def test_track_uncommitted_changes_called(self, mock_track_changes):
+        git_ref = GitRef()
+        with init_run(mode="debug", git_ref=git_ref) as run:
```
Raalsky, line 239: + what about testing the default behavior whenever the `git_ref` is `None`?

tests/unit/neptune/new/internal/utils/test_git.py
```diff
@@ -59,3 +61,84 @@ def test_getting_git_info(self, repo_mock):
...
+        assert diff is None
+        assert tracker.upstream_commit_sha is None
+        repo_mock.git.diff.assert_not_called()
+
+    @patch("neptune.internal.utils.git.File")
+    @patch("neptune.metadata_containers.Run")
+    def test_track_uncommitted_changes(self, mock_run, mock_file):
+        mock_tracker = MagicMock()
+
+        with patch("neptune.internal.utils.git.DiffTracker.from_git_ref", return_value=mock_tracker):
```
Raalsky, line 139: Check lines 36 and 37 in this file. This and similar mocks could be replaced with a simplified syntax (I think so).

src/neptune/metadata_containers/run.py
AleksanderWWW: Check out `DiffTracker.from_git_ref` in `src/neptune/internal/utils/git.py`

tests/unit/neptune/new/internal/utils/test_git.py
AleksanderWWW: I think that this way I won't be able to access the mock_tracker object to check the args with which it's methods were called.

tests/unit/neptune/new/internal/utils/test_git.py
```diff
@@ -59,3 +61,84 @@ def test_getting_git_info(self, repo_mock):
             branch="master",
             remotes=[],
         )
+
+
+class TestDiffTracker:
```
Raalsky, line 66: Can we structurize all tests below with `// given`, `// when` etc.?

tests/unit/neptune/new/internal/utils/test_git.py
Raalsky: You're probably right :wink:. We should be able to simplify it but I'm struggling figuring it out.

src/neptune/internal/utils/git.py
```diff
@@ -13,23 +13,40 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-__all__ = ["to_git_info", "GitInfo"]
+__all__ = [
+    "to_git_info",
+    "GitInfo",
+    "track_uncommitted_changes",
+    "DiffTracker",
```
Raalsky, line 20: No need to expose it. Part of the implementation I think.

src/neptune/internal/utils/git.py
```diff
@@ -90,3 +107,95 @@ def to_git_info(git_ref: Union[GitRef, GitRefDisabled]) -> Optional[GitInfo]:
...
+        if git_ref == GitRef.DISABLED:
+            return None
+
+        initial_repo_path = git_ref.resolve_path()
+        if initial_repo_path is None:
+            return None
+        try:
+            repo = get_git_repo(repo_path=initial_repo_path)
+        except (NoSuchPathError, InvalidGitRepositoryError):
+            repo = None
```
Raalsky, line 133: `return None`?

src/neptune/internal/utils/git.py
```diff
@@ -90,3 +107,95 @@ def to_git_info(git_ref: Union[GitRef, GitRefDisabled]) -> Optional[GitInfo]:
...
+
+        if repo:
```
Raalsky, line 135: After the above suggestion -> no need to check it (`get_git_repo` returns always `Repo`)

src/neptune/internal/utils/git.py
```diff
@@ -90,3 +107,95 @@ def to_git_info(git_ref: Union[GitRef, GitRefDisabled]) -> Optional[GitInfo]:
         )
     except:  # noqa: E722
         return None
+
+
+class DiffTracker:
+    def __init__(self, repo: git.Repo):
+        self._repo = repo
+        self._head = self._repo.head
```
Raalsky, line 115: Could we add type hints here (`HEAD` and `Optional[str]` for `self._upstream_commit_sha`)?

src/neptune/internal/utils/git.py
```diff
@@ -90,3 +106,94 @@ def to_git_info(git_ref: Union[GitRef, GitRefDisabled]) -> Optional[GitInfo]:
...
+        if tracking_branch:
+            return tracking_branch.commit
+
+        return self._search_for_most_recent_ancestor()
+
+    def _search_for_most_recent_ancestor(self) -> Optional[git.Commit]:
+        most_recent_ancestor: Optional[git.Commit] = None
+
+        try:
+            for branch in self._repo.branches:
```
Raalsky, line 173: Probably `branches` are deprecated and IDE says that`self._repo.heads` is recommended.

src/neptune/internal/utils/git.py
```diff
@@ -90,3 +106,94 @@ def to_git_info(git_ref: Union[GitRef, GitRefDisabled]) -> Optional[GitInfo]:
...
+
+            self._upstream_commit_sha = upstream_commit.hexsha
+
+            try:
+                return self._repo.git.diff(upstream_commit.hexsha)
+            except GitCommandError:
+                return
+
+    def _get_relevant_upstream_commit(self) -> Optional[git.Commit]:
+        try:
```
Raalsky, line 159: These few lines could be extracted to some `get_tracking_branch_commit` function :wink:

src/neptune/internal/utils/git.py
```diff
@@ -90,3 +106,94 @@ def to_git_info(git_ref: Union[GitRef, GitRefDisabled]) -> Optional[GitInfo]:
         )
     except:  # noqa: E722
         return None
+
+
+class DiffTracker:
```
Raalsky, line 111: This class does definitely too many things at once.  Try to prepare a dataclass that will handle the contents of diff and commit sha + a function that accepts a GitRef and returns this dataclass. `track_uncommitted_changes` shouldn't be aware of dirty vs non-dirty and `DiffTracker` currently handles data, computation and result at the same time.

Raalsky (approved):

PR #1373, opened 2023-06-27
Fixed support of stringify value in series attributes with step

Raalsky:

### DIFF
--- a/tests/unit/neptune/new/test_stringify_unsupported.py
+++ b/tests/unit/neptune/new/test_stringify_unsupported.py
@@ -439,8 +439,8 @@ class TestStringifyUnsupported:
 
     def test_append__float(self, run):
         with assert_no_warnings():
-            run["stringified"].append(stringify_unsupported(1.0))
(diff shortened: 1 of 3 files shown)


tests/unit/neptune/new/test_stringify_unsupported.py
```diff
@@ -439,8 +439,8 @@ def test_extend__dict(self, run):
 
     def test_append__float(self, run):
         with assert_no_warnings():
-            run["stringified"].append(stringify_unsupported(1.0))
```
BartoszPrusak, line 442: I think it'd be great to test both `append()` with and without the `step` argument. WDYT?

BartoszPrusak (dismissed):

tests/unit/neptune/new/test_stringify_unsupported.py
Raalsky: Yes, I was thinking about doing it but I was definitely too lazy :P. Give me a minute.

BartoszPrusak (approved):

PR #1374, opened 2023-06-27
Raise an error when saving dict with blank key 

BartoszPrusak:
Saving a dict() or a Namespace that has an entry under key `""` is not allowed in Neptune. This change adds an appropriate error message for such case.

Saving a dict in a run that has `""` as a key led to `object has no attribute 'items'` error.\
That's because saving `run["foo"] = {"bar": 123}` actually saves the item as `run["foo/bar"] = 123`. Therefore, `run["foo"] = {"": 0.2}` is actually saved as `run["foo/"] = 0.2`.\
Since Neptune interprets `"foo"` and `"foo/"` as the same key, this code is understood as writing a dict and another value type under the same key, which is illegal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Fix SSL errors after forking process ([#1353](https://github.com/neptune-ai/neptune-client/pull/1353))
 - Fixed support of stringify value in series attributes with step ([#1373](https://github.com/neptune-ai/neptune-client/pull/1373))
 - `dict`s and `Namespace`s that are written to runs and contain an empty string "" key now produce a warning and drop
-  the entry with such key instead of raising an
+  the entry with such a key instead of raising an
   exception ([#1374](https://github.com/neptune-ai/neptune-client/pull/1374))
 
 ### Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixes
 - Fixed GitPython `is_dirty` failing on Windows ([#1371](https://github.com/neptune-ai/neptune-client/pull/1371))
 - Fix SSL errors after forking process ([#1353](https://github.com/neptune-ai/neptune-client/pull/1353))
+- Using an empty key in dict saved to a container (e.g. a run) now raises a ValueError ([#1374](https://github.com/neptune-ai/neptune-client/pull/1374))
 
 ### Changes
 - Added support of writing to archived project exception ([#1355](https://github.com/neptune-ai/neptune-client/pull/1355))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ### Fixes
 - Fixed GitPython `is_dirty` failing on Windows ([#1371](https://github.com/neptune-ai/neptune-client/pull/1371))
 - Fix SSL errors after forking process ([#1353](https://github.com/neptune-ai/neptune-client/pull/1353))
-- Using an empty key in dict saved to a container (e.g. a run) now raises a ValueError ([#1374](https://github.com/neptune-ai/neptune-client/pull/1374))
+- `dict`s and `Namespace`s that are written to runs and contain an empty string "" key now produce a warning and drop
+  the entry with such key instead of raising an
+  exception ([#1374](https://github.com/neptune-ai/neptune-client/pull/1374))
 
 ### Changes
 - Added support of writing to archived project exception ([#1355](https://github.com/neptune-ai/neptune-client/pull/1355))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixes
 - Fixed GitPython `is_dirty` failing on Windows ([#1371](https://github.com/neptune-ai/neptune-client/pull/1371))
 - Fix SSL errors after forking process ([#1353](https://github.com/neptune-ai/neptune-client/pull/1353))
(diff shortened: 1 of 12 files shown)


src/neptune/attributes/namespace.py
```diff
@@ -42,6 +43,8 @@
 )
 from neptune.types.namespace import Namespace as NamespaceVal
 
+_logger = logging.getLogger(__name__)
```
Raalsky, line 46: Let's use `from neptune.internal.utils.logger import logger` instead.

src/neptune/attributes/namespace.py
```diff
@@ -99,6 +102,14 @@ def assign(self, value: Union[NamespaceVal, dict, Mapping], *, wait: bool = Fals
             value = NamespaceVal(value)
 
         for k, v in value.value.items():
+            if len(parse_path(k)) == 0:
```
Raalsky, line 105: It's Python my dear:
```python
>>> if not []:
...     print("It's working!")
...
It's working!
```

src/neptune/attributes/namespace.py
```diff
@@ -99,6 +102,14 @@ def assign(self, value: Union[NamespaceVal, dict, Mapping], *, wait: bool = Fals
...
+                _logger.warning(
+                    'Key "%s" can\'t be used in Namespaces and dicts stored in Neptune. Please use a non-empty key '
```
Raalsky, line 107: Let's use f-strings instead :wink:

tests/unit/neptune/new/test_stringify_unsupported.py
```diff
@@ -302,6 +302,35 @@ def test_assign__dict(self, run):
 
         assert run["regular"].fetch() == run["stringified"].fetch()
 
+    def test_assign__drops_dict_entry_with_practically_empty_key(self, caplog, run):
+        with assert_unsupported_warning():
+            run["unsupported"] = {"": Obj(), "x": Obj()}
+        with assert_unsupported_warning():
+            run["unsupported"] = {"///": Obj(), "x": Obj()}
+        logs = caplog.get_records("call")
```
Raalsky, line 310: This looks like a pretty good pattern (catching warning) -> Could we extract it to context-manager? Check `assert_no_warnings` as a reference.

tests/unit/neptune/new/test_stringify_unsupported.py
```diff
@@ -302,6 +302,35 @@ def test_assign__dict(self, run):
 
         assert run["regular"].fetch() == run["stringified"].fetch()
 
+    def test_assign__drops_dict_entry_with_practically_empty_key(self, caplog, run):
```
Raalsky, line 305: `stringify_unsupported` looks like doesn't have an impact here. Could we move this tests somewhere else?

src/neptune/attributes/namespace.py
Raalsky, line 105: I'm not sure about this value verification here. Some of the types that we have (`Artifact`, `File`, and a few ones) does have a verification inside of a constructor (`__init__`). Could we move this directly to the constructor of `Namespace` as well? In addition, let's just filter out all the unwanted values and log a combined warning like `"", "/" could not be used as a name of an attribute`

BartoszPrusak: I know, but I was on the fence on how many people are ok with that syntax :|

Raalsky: That's quite common for Python. https://stackoverflow.com/a/53522

tests/unit/neptune/new/test_stringify_unsupported.py
BartoszPrusak: Done, moved to `test_handler.py`

src/neptune/attributes/namespace.py
BartoszPrusak: Done and done

tests/unit/neptune/new/test_stringify_unsupported.py
BartoszPrusak: Done

src/neptune/attributes/namespace.py
BartoszPrusak: Done

src/neptune/attributes/namespace.py
BartoszPrusak: Done

src/neptune/attributes/namespace.py
BartoszPrusak: Done

Raalsky (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 - Fixed GitPython `is_dirty` failing on Windows ([#1371](https://github.com/neptune-ai/neptune-client/pull/1371))
 - Fix SSL errors after forking process ([#1353](https://github.com/neptune-ai/neptune-client/pull/1353))
 - Fixed support of stringify value in series attributes with step ([#1373](https://github.com/neptune-ai/neptune-client/pull/1373))
+- `dict`s and `Namespace`s that are written to runs and contain an empty string "" key now produce a warning and drop
+  the entry with such key instead of raising an
```
normandy7, line 8:
```suggestion
  the entry with such a key instead of raising an
```

normandy7 (commented):
Nitpick; looks OK otherwise.

CHANGELOG.md
BartoszPrusak: Done

normandy7 (approved):

PR #1380, opened 2023-06-28
Wait for elastic index after experiment creation in trashing e2e

jdudycz:
After switching to new elasticsearch client code in leaderboard it seems that we need to wait a bit for index refresh after creating test experiments

### DIFF
--- a/tests/e2e/management/test_management.py
+++ b/tests/e2e/management/test_management.py
@@ -13,9 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from time import sleep
-from typing import Dict
+from typing import (
+    Callable,
+    Dict,
+    List,
+)
 
+import backoff
 import pytest
 
 from neptune import init_model_version
@@ -43,6 +47,7 @@ from neptune.management.exceptions import (
     WorkspaceOrUserNotFound,
 )
 from neptune.management.internal.utils import normalize_project_name
+from neptune.metadata_containers.metadata_containers_table import Table
 from tests.e2e.base import (
     BaseE2ETest,
     fake,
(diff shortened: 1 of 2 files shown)


tests/e2e/management/test_management.py
```diff
@@ -415,6 +415,8 @@ def test_trash_runs_and_models(self, project, environment):
         run2_id = initialize_container(ContainerType.RUN, project=environment.project)["sys/id"].fetch()
         model1_id = initialize_container(ContainerType.MODEL, project=environment.project)["sys/id"].fetch()
         model2_id = initialize_container(ContainerType.MODEL, project=environment.project)["sys/id"].fetch()
+        # wait for the elasticsearch cache to fill
+        sleep(5)
```
aniezurawski, line 419: @Raalsky what do you think about creating some `eventualy(predicate, timeout)` method with expotentional retries?

jdudycz: I've created a task for that https://neptune-labs.atlassian.net/browse/NPT-13399

Raalsky: I've already been using `backoff` package in a few integrations tests. Something like:
```python
@backoff.on_exception(backoff.expo, AssertionError, max_value=1, max_time=60)
def test_....():
    ....
```

Raalsky: I think that there is something like a constant for the number of seconds to wait.

jdudycz: Okay, will do

jdudycz: Done

aniezurawski (approved):

PR #1385, opened 2023-06-30
add warning for dependencies tracking if file not present

AleksanderWWW:

### DIFF
--- a/src/neptune/internal/utils/dependency_tracking.py
+++ b/src/neptune/internal/utils/dependency_tracking.py
@@ -7,6 +7,7 @@ __all__ = [
 import os
 import subprocess
 import sys
+import warnings
 from abc import (
     ABC,
     abstractmethod,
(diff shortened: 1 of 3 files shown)


twolodzko (approved):

src/neptune/internal/utils/dependency_tracking.py
```diff
@@ -46,3 +47,5 @@ def __init__(self, path: Union[str, os.PathLike]):
     def log_dependencies(self, run: "Run") -> None:
         if os.path.isfile(self._path):
             run["source_code/files"].upload_files(os.path.basename(self._path))
+        else:
+            warnings.warn(f"File '{self._path}' does not exist - skipping dependency file upload.")
```
twolodzko, line 51: Should this be a warning or an error?

parthpankajtiwary: @twolodzko do you consider the issue big enough to halt the subsequent experimentation? I was thinking a warning should be sufficient and would allow users to ensure the file is present in subsequent runs.

twolodzko: @parthpankajtiwary people usually don't pay attention to warnings. So with a warning, we are making the assumption that it is ok for the users if they provide an incorrect name of the file and log nothing and discover this only sometime later when browsing the metadata and noticing that the requirements are missing. I would say it would hurt more if they notice this too late vs if they notice it when running the code.

PR #1389, opened 2023-07-04
Misc adjustments to dependency tracking

AleksanderWWW:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
-## [UNRELEASED] neptune 1.3.2
+##  [UNRELEASED] neptune 1.3.3
+
+### Changes
+- Dependency tracking feature will log an error if a given file path doesn't exist ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
+- Use `pip list --format=freeze` instead of `pip freeze` in dependency tracking ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
+- Log both uploaded and inferred requirements to the same namespace ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
+
+## neptune 1.3.2
 
 ### Fixes
 - Fixed GitPython `is_dirty` failing on Windows ([#1371](https://github.com/neptune-ai/neptune-client/pull/1371))
@@ -11,7 +18,6 @@
 
 ### Changes
 - Added support of writing to archived project exception ([#1355](https://github.com/neptune-ai/neptune-client/pull/1355))
-- Dependency tracking feature will raise warning if a given file path doesn't exist ([#1385](https://github.com/neptune-ai/neptune-client/pull/1385))
 
 ## neptune 1.3.1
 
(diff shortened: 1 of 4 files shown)


src/neptune/internal/utils/dependency_tracking.py
```diff
@@ -4,10 +4,10 @@
     "FileDependenciesStrategy",
 ]
 
+import logging
```
Raalsky, line 7: We have our custom logger somewhere. It has some settings to simplify filtering etc. I think it's:
```python
from neptune.internal.utils.logger import logger
```

src/neptune/internal/utils/dependency_tracking.py
```diff
@@ -32,7 +32,9 @@ def log_dependencies(self, run: "Run") -> None:
 class InferDependenciesStrategy(DependencyTrackingStrategy):
     def log_dependencies(self, run: "Run") -> None:
         try:
-            dependencies_str = subprocess.check_output([sys.executable, "-m", "pip", "freeze"]).decode("utf-8")
+            dependencies_str = subprocess.check_output([sys.executable, "-m", "pip", "list", "--format=freeze"]).decode(
```
Raalsky, line 35: Have you tried the `importlib` approach? I've been suggesting something like:
```python
import sys

if sys.version_info >= (3, 8):
    from importlib.metadata import distributions
else:
    from importlib_metadata import distributions


def list_installed_packages():
    dists = list(sorted(distributions(), key=lambda d: d.metadata['Name']))

    for dist in dists:
        name, version = dist.metadata['Name'], dist.metadata['Version']
        print(f'{name}=={version}')


def main():
    list_installed_packages()


if __name__ == '__main__':
    main()
```

AleksanderWWW: @SiddhantSadangi

src/neptune/internal/utils/dependency_tracking.py
AleksanderWWW: I tried that at first and it doesn't show the level - just the message. I could change the logging format, but that would affect other usages of this logger throughout the code base, right?

src/neptune/internal/utils/dependency_tracking.py
SiddhantSadangi: Just tried. This seems to work very well!

Lets do this instead of subprocess

Raalsky (approved):

CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
-## [UNRELEASED] neptune 1.3.2
+##  [UNRELEASED] neptune 1.3.3
+
+### Changes
+- Dependency tracking feature will log an error if a given file path doesn't exist ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
+- Use `pip list --format=freeze` instead of `pip freeze` in dependency tracking ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
```
twolodzko, line 5: Not sure if it was fixed, but this note is misleading as we're not using it here.

PR #1391, opened 2023-07-07
make the path to .neptune folder absolute

AleksanderWWW:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixes
 - Fixed operation processor bug if current working directory is different from the script directory ([#1391](https://github.com/neptune-ai/neptune-client/pull/1391))
+
 ## neptune 1.3.2
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-##  neptune 1.3.3
+## [UNRELEASED] neptune 1.3.3
+
+### Fixes
+- Fixed operation processor bug if current working directory is different from the script directory ([#1391](https://github.com/neptune-ai/neptune-client/pull/1391))
+
+##  neptune 1.3.3rc0
 
 ### Changes
 - Dependency tracking feature will log an error if a given file path doesn't exist ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ### Changes
 - Dependency tracking feature will log an error if a given file path doesn't exist ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
-- Use `pip list --format=freeze` instead of `pip freeze` in dependency tracking ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
+- Use `importlib` instead of `pip freeze` in dependency tracking ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
 - Log both uploaded and inferred requirements to the same namespace ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
 
+### Fixes
+- Fixed operation processor bug if current working directory is different from the script directory ([#1391](https://github.com/neptune-ai/neptune-client/pull/1391))
 ## neptune 1.3.2
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-##  [UNRELEASED] neptune 1.3.3
+##  neptune 1.3.3
 
 ### Changes
 - Dependency tracking feature will log an error if a given file path doesn't exist ([#1389](https://github.com/neptune-ai/neptune-client/pull/1389))
@@ -7,6 +7,7 @@
 
 ### Fixes
 - Fixed operation processor bug if current working directory is different from the script directory ([#1391](https://github.com/neptune-ai/neptune-client/pull/1391))
+- Tracking uncommitted changes and dependencies will be skipped in case of any exception, to not disturb the run initialization ([#1395](https://github.com/neptune-ai/neptune-client/pull/1395))
 
 ## neptune 1.3.2
 
--- a/src/neptune/internal/operation_processors/operation_storage.py
+++ b/src/neptune/internal/operation_processors/operation_storage.py
@@ -64,9 +64,16 @@ class OperationStorage:
     def close(self):
         shutil.rmtree(self.data_path, ignore_errors=True)
 
-        parent = self.data_path.parent
+        try:
+            parent = self.data_path.parent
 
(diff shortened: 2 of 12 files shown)


src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -34,29 +34,46 @@ def __init__(self, data_path: Union[str, Path]):
             self._data_path = data_path
         else:
             self._data_path = Path(data_path)
-
         # initialize directories
         os.makedirs(self.data_path, exist_ok=True)
         os.makedirs(self.upload_path, exist_ok=True)
 
+        cwd = Path(os.getcwd())
+        self._data_path_abs = cwd / self._data_path
```
Raalsky, line 42: Is that much different from `self._data_path.resolve()`?

src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -34,29 +34,46 @@ def __init__(self, data_path: Union[str, Path]):
...
     @property
     def data_path(self) -> Path:
         return self._data_path
 
+    @property
+    def data_path_abs(self) -> Path:
+        return self._data_path_abs
+
     @property
     def upload_path(self) -> Path:
```
Raalsky, line 53: There are only two usages of this in `operation.py`. It could be changed to an absolute path right here.

src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -34,29 +34,46 @@ def __init__(self, data_path: Union[str, Path]):
...
 
     def close(self):
         shutil.rmtree(self.data_path, ignore_errors=True)
 
-        parent = self.data_path.parent
+        try:
+            parent = self.data_path.parent
+
+            files = os.listdir(parent)
+        except FileNotFoundError:
```
Raalsky, line 71: Try to write a test for the behavior that is missing (failing) - the one with changeing the current working directory.

src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -34,29 +34,46 @@ def __init__(self, data_path: Union[str, Path]):
...
+
     @staticmethod
     def _get_container_dir(type_dir: str, container_id: UniqueId, container_type: ContainerType):
         return f"{NEPTUNE_DATA_DIRECTORY}/{type_dir}/{container_type.create_dir_name(container_id)}"
 
     def close(self):
         shutil.rmtree(self.data_path, ignore_errors=True)
 
-        parent = self.data_path.parent
+        try:
```
Raalsky, line 67: I think that it relies on relative paths as we would like users to be able to move the `.neptune` directory from one machine to another and still be able to upload the metadata.

src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -34,11 +34,12 @@ def __init__(self, data_path: Union[str, Path]):
             self._data_path = data_path
```
Raalsky, line 34: We can get rid of it with just `self._data_path = Path(data_path)` (always).

src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -34,11 +34,12 @@ def __init__(self, data_path: Union[str, Path]):
...
         else:
             self._data_path = Path(data_path)
-
         # initialize directories
         os.makedirs(self.data_path, exist_ok=True)
```
Raalsky, line 38: I think `os.makedirs` accepts both `str` and `Path` so this could be even before the conversion and first assignment of `self_data_path`.

src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -34,11 +34,12 @@ def __init__(self, data_path: Union[str, Path]):
...
         os.makedirs(self.upload_path, exist_ok=True)
 
+        self._data_path = self._data_path.resolve()
```
Raalsky, line 41: In general, I don't like this "expanding" pattern with reassigning to the value. Let's use `= Path(data_path).resolve()`

src/neptune/internal/operation_processors/operation_storage.py
AleksanderWWW: So let's say I have
```
def __init__(self, data_path: Union[str, Path]):
    os.makedirs(data_path, exist_ok=True)
    
    self._data_path = Path(data_path).resolve()
    
    os.makedirs(self.upload_path, exist_ok=True)

@property
def data_path(self) -> Path:
    return self._data_path

@property
def upload_path(self) -> Path:
    return self.data_path / "upload_path"
```

> `os.makedirs(self.upload_path, exist_ok=True)`
will result in PermissionError, as now the upload path will be absolute.

CHANGELOG.md
```diff
@@ -2,10 +2,11 @@
```
Raalsky, line 2: Could we mark somehow that we released `1.3.3rc0` and `1.3.3` still awaits release?

src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -29,15 +28,13 @@
...
-            self._data_path = data_path
-        else:
-            self._data_path = Path(data_path)
+    UPLOAD_PATH: str = "upload_path"
 
-        # initialize directories
-        os.makedirs(self.data_path, exist_ok=True)
-        os.makedirs(self.upload_path, exist_ok=True)
+    def __init__(self, data_path: Path):
+        self._data_path = Path(data_path).resolve()
```
Raalsky, line 34: `data_path.resolve()` 😉

tests/unit/neptune/new/client/test_run.py
```diff
@@ -117,7 +116,7 @@ def test_entrypoint(self):
             self.assertEqual(exp["source_code/entrypoint"].fetch(), "main_dir/main.py")
 
         with init_run(mode="debug", source_files=["internal/*"]) as exp:
-            self.assertEqual(exp["source_code/entrypoint"].fetch(), "main.py")
+            self.assertEqual(exp["source_code/entrypoint"].fetch(), "../main.py")
```
Raalsky, line 119: Why do we have to change it?

AleksanderWWW: I had to remove the `os.getcwd` patch, as it was messing up path resolving in operation storage. Without it, the entrypoint in this case evaluates to `../main.py`. I investigated it and this patch was a little strange, as it also changed the behaviour of `get_common_root` and produced a rather unexpected output. 

The `../main.py` output makes more sense to me, cause the 'common_root' of the source_code var is `cwd + internal/`. The entrypoint path is `cwd + main.py`. So the relative path of entrypoint, where the start is common_root is `../main.py`. Check out `internal/utils/source_code.py` file, `upload_source_code` func

Raalsky (approved):

PR #1398, opened 2023-07-12
fix jsonschema version

AleksanderWWW:


AleksanderWWW:
If it's ok I'd actually leave two python versions, as many packages are no longer taking 3.7 into account in their newest releases (new version of jsonschema that was the reason for our test failures worked well on 3.7 for example)

Issue #1396, opened 2023-07-11
BUG: Neptune v1.3.2 is slow when using forked data loaders in Linux


issamemari:
### Describe the bug

I have a training script that uses the Trainer class from the HuggingFace Transformers library to train a text classification model. The Trainer class uses a NeptuneCallback to log training metrics to Neptune.

I noticed that my training script is inexplicably slow, so I profiled it using py-spy and found that `neptune/internal/threading/daemon.py` is taking more than 90% of the execution time of my script. I tried disabling the Neptune callback and saw a dramatic improvement in the speed of the training script.

### Reproduction

Here's my training code. I have the Neptune client installed and project name and API key in the environment variables so the Trainer class automatically send logs to Neptune using the NeptuneCallback class.

```python
import logging

import hydra
from transformers import Trainer, TrainingArguments

from dataset import LukalficationDataset
from dataset.classes import LEVEL1, LEVEL2, LEVEL3
from metrics import MetricsLogger
from model import LukalificationModel
from preprocessing import MultiLabelOneHotEncoder, Processor

logger = logging.getLogger(__name__)


@hydra.main(config_path=".", config_name="config.yaml", version_base="1.2")
def main(config):
    logger.info("Creating dataset")
    train_dataset = LukalficationDataset(**config.dataset.train)
    eval_dataset = LukalficationDataset(**config.dataset.eval)

    logger.info("Creating preprocessor")
    processor = Processor(
        tokenizer_config=config.tokenizer,
        label_encoders=[
            MultiLabelOneHotEncoder(LEVEL1),
            MultiLabelOneHotEncoder(LEVEL2),
            MultiLabelOneHotEncoder(LEVEL3),
        ],
    )

    logger.info("Creating model")
    model = LukalificationModel([len(LEVEL1), len(LEVEL2), len(LEVEL3)], **config.model)

    training_args = TrainingArguments(**config.training)

    logger.info("Training model")
    trainer = Trainer(
        model=model,
        args=training_args,
        train_dataset=train_dataset,
        eval_dataset=eval_dataset,
        data_collator=processor.data_collator,
        compute_metrics=MetricsLogger(
            label_encoders=processor.label_encoders
        ).compute_metrics,
    )

    trainer.train()

    logger.info("Training complete")


if __name__ == "__main__":
    main()
```

### Expected behavior

I expect logging to Neptune to not impact the speed of my training script.

### Environment
**The output of `pip list`:**
```
Package                    Version
[...]

PR #1402, opened 2023-07-19
fix: Retry request when ChunkedEncodingError occurred.

PatrykGala:

### DIFF
--- a/src/neptune/common/backends/utils.py
+++ b/src/neptune/common/backends/utils.py
@@ -34,6 +34,7 @@ from bravado.exception import (
     HTTPTooManyRequests,
     HTTPUnauthorized,
 )
+from requests.exceptions import ChunkedEncodingError
 from urllib3.exceptions import NewConnectionError
 
 from neptune.api.exceptions_utils import handle_json_errors
@@ -100,6 +101,7 @@ def with_api_exceptions_handler(func):
                 HTTPTooManyRequests,
                 HTTPInternalServerError,
                 NewConnectionError,
+                ChunkedEncodingError,
             ) as e:
                 time.sleep(min(2 ** min(10, retry), MAX_RETRY_TIME))
                 last_exception = e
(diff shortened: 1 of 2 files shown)


src/neptune/common/backends/utils.py
```diff
@@ -100,6 +101,7 @@ def wrapper(*args, **kwargs):
                 HTTPTooManyRequests,
                 HTTPInternalServerError,
                 NewConnectionError,
+                ChunkedEncodingError,
```
aniezurawski, line 104: How about retrying only once like in case of SSL error?

aniezurawski: Also some very explanatory comment would be welcome.

PatrykGala: I suppose, that client closes the connection before forking and after resuming - client doesn't receive a response from server.

This error is similar to `HTTPRequestTimeout or BravadoConnectionError` - and we can retry this request.

aniezurawski: It's hard to believe that it throws some very random exception when the connection is just closed. Can we test this hypothesis?

mcochner:
Do you know if there is an ETA on including this in a new release of `neptune-client`? We observe issues with the stability of `neptune-client` and (quite infrequently but usually once a day) it dies on `ChunkedEncodingError`.

aniezurawski:
@mcochner see these response: https://github.com/neptune-ai/neptune-client/issues/1396#issuecomment-1649826972
These issues are related and this workaround should help with this error as well.

PatrykGala:
> Do you know if there is an ETA on including this in a new release of `neptune-client`? We observe issues with the stability of `neptune-client` and (quite infrequently but usually once a day) it dies on `ChunkedEncodingError`.

Hi @mcochner, 
I have been trying to reproduce this error, but without success. Could you please check your logs for any occurrences of ChunkedEncodingError along with timestamps? I would like to verify this exception with our backend logs.

aniezurawski (approved):

PR #1409, opened 2023-07-26
allow users to pass neptune data directory as env var

AleksanderWWW:

### DIFF
--- a/src/neptune/internal/operation_processors/async_operation_processor.py
+++ b/src/neptune/internal/operation_processors/async_operation_processor.py
@@ -84,10 +84,10 @@ class AsyncOperationProcessor(OperationProcessor):
     @staticmethod
     def _init_data_path(container_id: UniqueId, container_type: ContainerType) -> Path:
         now = datetime.now()
-        container_dir = get_container_dir(ASYNC_DIRECTORY, container_id, container_type)
-        data_path = f"{container_dir}/exec-{now.timestamp()}-{now.strftime('%Y-%m-%d_%H.%M.%S.%f')}-{os.getpid()}"
-        data_path = data_path.replace(" ", "_").replace(":", ".")
-        return Path(data_path)
+        process_path = f"exec-{now.timestamp()}-{now.strftime('%Y-%m-%d_%H.%M.%S.%f')}-{os.getpid()}".replace(
+            " ", "_"
+        ).replace(":", ".")
+        return get_container_dir(ASYNC_DIRECTORY, container_id, container_type, process_path)
 
     def enqueue_operation(self, op: Operation, *, wait: bool) -> None:
         self._last_version = self._queue.put(op)
--- a/src/neptune/internal/operation_processors/async_operation_processor.py
+++ b/src/neptune/internal/operation_processors/async_operation_processor.py
@@ -29,10 +29,7 @@ from typing import (
     Optional,
 )
 
-from neptune.constants import (
-    ASYNC_DIRECTORY,
-    NEPTUNE_DATA_DIRECTORY,
-)
+from neptune.constants import ASYNC_DIRECTORY
 from neptune.envs import NEPTUNE_SYNC_AFTER_STOP_TIMEOUT
 from neptune.exceptions import NeptuneSynchronizationAlreadyStoppedException
 from neptune.internal.backends.neptune_backend import NeptuneBackend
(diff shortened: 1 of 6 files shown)


src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -27,6 +25,14 @@
 from neptune.internal.utils.logger import logger
 
 
+def get_container_dir(type_dir: str, container_id: UniqueId, container_type: ContainerType) -> str:
```
AleksanderWWW, line 28: This func is basically the clue of this PR. The rest is just adjusting the existing code for its usage

src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -27,6 +25,14 @@
...
+    neptune_data_dir_env = os.getenv("NEPTUNE_DATA_DIRECTORY")
+
+    neptune_data_dir = neptune_data_dir_env if neptune_data_dir_env else NEPTUNE_DATA_DIRECTORY
```
Raalsky, line 31: `getenv` accepts default value https://docs.python.org/3.7/library/os.html#os.getenv

src/neptune/internal/operation_processors/operation_storage.py
```diff
@@ -27,6 +25,14 @@
...
+
+    return f"{neptune_data_dir}/{type_dir}/{container_type.create_dir_name(container_id)}"
```
Raalsky, line 33: How about `Path` as a return type?

src/neptune/internal/operation_processors/offline_operation_processor.py
```diff
@@ -44,7 +44,8 @@ def __init__(self, container_id: UniqueId, container_type: ContainerType, lock:
 
     @staticmethod
     def _init_data_path(container_id: UniqueId, container_type: ContainerType) -> Path:
-        return Path(f"{NEPTUNE_DATA_DIRECTORY}/{OFFLINE_DIRECTORY}/{container_type.create_dir_name(container_id)}")
+        container_dir = get_container_dir(OFFLINE_DIRECTORY, container_id, container_type)
```
Raalsky, line 47: This could be simplified after changing the return type of `get_container_id` to `Path`

src/neptune/internal/operation_processors/operation_storage.py
Raalsky, line 28: How about adding an optional custom subdirectory (like we do in `async` & `sync` `_init_data_path`). I mean we can simplify something like:
```python
        now = datetime.now()
        container_dir = get_container_dir(ASYNC_DIRECTORY, container_id, container_type)
        data_path = f"{container_dir}/exec-{now.timestamp()}-{now.strftime('%Y-%m-%d_%H.%M.%S.%f')}-{os.getpid()}"
        data_path = data_path.replace(" ", "_").replace(":", ".")
        return Path(data_path)
```
with:
```python
        now = datetime.now()
        process_path = f"exec-{now.timestamp()}-{now.strftime('%Y-%m-%d_%H.%M.%S.%f')}-{os.getpid()}"
        process_path = process_path.replace(" ", "_").replace(":", ".")
        return get_container_dir(ASYNC_DIRECTORY, container_id, container_type, process_path)
```

src/neptune/internal/operation_processors/async_operation_processor.py
```diff
@@ -84,10 +84,10 @@ def __init__(
     @staticmethod
     def _init_data_path(container_id: UniqueId, container_type: ContainerType) -> Path:
         now = datetime.now()
-        container_dir = f"{NEPTUNE_DATA_DIRECTORY}/{ASYNC_DIRECTORY}/{container_type.create_dir_name(container_id)}"
-        data_path = f"{container_dir}/exec-{now.timestamp()}-{now.strftime('%Y-%m-%d_%H.%M.%S.%f')}-{os.getpid()}"
-        data_path = data_path.replace(" ", "_").replace(":", ".")
-        return Path(data_path)
+        process_path = f"exec-{now.timestamp()}-{now.strftime('%Y-%m-%d_%H.%M.%S.%f')}-{os.getpid()}".replace(
```
Raalsky, line 87: We probably don't need `.replace(" ", "_")` anymore and it's some kind of a leftover. What do you think?

PR #1410, opened 2023-07-28
fix: Use condition in ConnectionRetryWrapper. Logs exceptions on Netw…

PatrykGala:
…ork errors.


src/neptune/internal/threading/daemon.py
```diff
@@ -115,25 +117,24 @@ def __init__(self, kill_message):
         def __call__(self, func):
             @functools.wraps(func)
             def wrapper(self_: Daemon, *args, **kwargs):
-                while not self_._is_interrupted():
+                while self_._is_working():
```
aniezurawski, line 120: Are you sure about this change? Looks like it could stop the main loop early.

aniezurawski (changes requested):

PR #1421, opened 2023-08-09
filter empty directory files and compute hash w/o file metadata

artsiomtserashkovich:


src/neptune/internal/backends/hosted_artifact_operations.py
```diff
@@ -55,20 +55,36 @@ def _compute_artifact_size(artifact_file_list: List[ArtifactFileData]):
     return artifact_size
 
 
+def _filter_empty_directory_files(files: List[ArtifactFileData]) -> List[ArtifactFileData]:
+    return list(filter(lambda x: not _is_s3_empty_directory_file(x), files))
+
+
+def _is_s3_empty_directory_file(file: ArtifactFileData) -> bool:
+    return file.type == "S3" and file.size == 0
```
jdudycz, line 63: The case is this will skip all the empty files(keys) on s3, not only the dot ones. As far as i remember we want to only skip the dotfiles and leave all the other empty ones, right?

artsiomtserashkovich: Yes, the main idea is to skip files that are used by S3 to identify directories with no files inside (as S3 has a "flat" structure - it's using empty files as identification). 
After checking, I could confirm that there are not only dot-named file_path (``.``), but also with the full name of the folder (``testFolder``). 
As we are trimming now slash at the end of file_path and cant use metadata['location'], there is no way to segregate user-created empty files and such directory files.

Additionally, we have to keep the same logic of the filtering at API and at the python client.

Based on that, for now, we will go with such criteria of filtering out directory files.

jdudycz (approved):

PR #1434, opened 2023-08-18
Fix request import

HAOCHENYE:
`JSONDecodeError` is defined here:

https://github.com/psf/requests/blob/main/src/requests/compat.py.

`requests.exceptions.JSONDecodeError` does not import  `JSONDecodeError` from `requests.compat` until v2.27.0

https://github.com/psf/requests/blob/v2.27.0/requests/exceptions.py
https://github.com/psf/requests/blob/v2.26.0/requests/exceptions.py

This PR could make neptune compatible with lower version requests.

### DIFF
--- a/src/neptune/api/requests_utils.py
+++ b/src/neptune/api/requests_utils.py
@@ -21,7 +21,7 @@ from typing import (
     Dict,
 )
 
-from requests.exceptions import JSONDecodeError
+from requests.compat import JSONDecodeError
 
 if TYPE_CHECKING:
     from bravado_core.response import IncomingResponse


src/neptune/api/requests_utils.py
```diff
@@ -21,7 +21,7 @@
     Dict,
 )
 
-from requests.exceptions import JSONDecodeError
+from requests.compat import JSONDecodeError
```
AleksanderWWW, line 24: How about 
```
try:
   from requests.exceptions import JSONDecodeError
except ImportError:
   from requests.compat import JSONDecodeError
```
?
Should be a little more readable - we want to underscore that we are using exceptions from requests library. `compat` doesn't tell much about the nature of the package.
What do you think?

AleksanderWWW (commented):
Thank you for the PR @HAOCHENYE . One question from my side :)

HAOCHENYE:
https://github.com/psf/requests/blob/v2.27.0/requests/compat.py is a compatible layer to import different function, class or modules, and JsonDecoderError is not an exception defined in request.  I think maybe it is not necessary to import it from the `request.exceptions`.  Besides, I think the `try ... import ` behavior is a little bit dangerous since it is hard to say whether the ImportError is triggered by `JSONDecodeError`

AleksanderWWW:
@HAOCHENYE could you also update CHANGELOG?

HAOCHENYE:
Sorry for not reading the `pull_request_template.md`. I've updated the CHANGELOG.md now.

AleksanderWWW:
No worries :smile: thank you!

AleksanderWWW (approved):

PR #1443, opened 2023-08-24
cast integers outside of 32bits to float in StringifyValue

AleksanderWWW:

### DIFF
--- a/src/neptune/internal/types/stringify_value.py
+++ b/src/neptune/internal/types/stringify_value.py
@@ -34,13 +34,16 @@ __all__ = ["StringifyValue", "extract_if_stringify_value"]
 
 from typing import Any
 
-from neptune.attributes import Integer
+from neptune.constants import (
+    MAX_32_BIT_INT,
+    MIN_32_BIT_INT,
+)
 
 
 class StringifyValue:
     def __init__(self, value: Any):
         # check if it's an integer outside 32bit range and cast it to float
-        if isinstance(value, int) and (value > Integer.MAX_32_BIT_INT or value < Integer.MIN_32_BIT_INT):
+        if isinstance(value, int) and (value > MAX_32_BIT_INT or value < MIN_32_BIT_INT):
             value = float(value)
 
         self.__value = value
--- a/src/neptune/internal/types/stringify_value.py
+++ b/src/neptune/internal/types/stringify_value.py
@@ -34,9 +34,15 @@ __all__ = ["StringifyValue", "extract_if_stringify_value"]
 
 from typing import Any
 
+from neptune.attributes import Integer
+
 
 class StringifyValue:
     def __init__(self, value: Any):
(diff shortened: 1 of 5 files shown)


AleksanderWWW:
Refactored Integer class as well, as it caused circular imports

src/neptune/internal/types/stringify_value.py
```diff
@@ -34,9 +34,18 @@
...
+    MAX_32_BIT_INT,
+    MIN_32_BIT_INT,
+)
+
 
 class StringifyValue:
     def __init__(self, value: Any):
+        # check if it's an integer outside 32bit range and cast it to float
+        if isinstance(value, int) and (value > MAX_32_BIT_INT or value < MIN_32_BIT_INT):
+            value = float(value)
```
SiddhantSadangi, line 47: Let's show a warning here saying that the value is being converted to float

src/neptune/internal/types/stringify_value.py
```diff
@@ -31,12 +31,22 @@
...
+    MIN_32_BIT_INT,
+)
+from neptune.internal.utils.logger import logger
+
 
 class StringifyValue:
     def __init__(self, value: Any):
+        # check if it's an integer outside 32bit range and cast it to float
+        if isinstance(value, int) and (value > MAX_32_BIT_INT or value < MIN_32_BIT_INT):
+            logger.info("Value '%d' of type int is outside of 32 bit range and will be casted to float", value)
```
SiddhantSadangi, line 47:
```suggestion
            logger.info("Value '%d' is outside the range of 32-bit integers (-2147483648 to 2147483647) and will be logged as float", value)
```

@normandy7 ?

normandy7: Fine by me.

normandy7 (approved):

SiddhantSadangi (dismissed):

src/neptune/internal/types/stringify_value.py
```diff
@@ -31,12 +31,27 @@
...
+)
+from neptune.internal.utils.logger import logger
+
 
 class StringifyValue:
     def __init__(self, value: Any):
+        # check if it's an integer outside 32bit range and cast it to float
+        if isinstance(value, int) and (value > MAX_32_BIT_INT or value < MIN_32_BIT_INT):
+            logger.info(
+                "Value '%d' is outside the range of 32-bit integers ('%d' to '%d') and will be logged as float",
```
aniezurawski, line 48: Why do we have to cast it to float if we want to stringify it anyway?

AleksanderWWW: Integers are not be stringified. This would serve like an additional functionality of stringify_unsupported - for unsupported types it will cast to string and for unsupported values (like ints outside of the range) it would cast those values to float

SiddhantSadangi (approved):

PR #1469, opened 2023-09-13
Fix exception handling in `ApiMethodWrapper.handle_neptune_http_errors`

AleksanderWWW:
…tune_http_errors


src/neptune/internal/backends/hosted_file_operations.py
```diff
@@ -380,7 +380,7 @@ def upload_raw_data(
         HTTPPaymentRequired.status_code,
     ):
         raise NeptuneLimitExceedException(reason=response.json().get("title", "Unknown reason"))
-    response.raise_for_status()
```
AleksanderWWW, line 383: Not needed as this is now handled in `ApiMethodWrapper.handle_neptune_http_errors`

src/neptune/api/exceptions_utils.py
```diff
@@ -34,6 +34,9 @@ def handle_json_errors(
 
     if error_processor:
         raise error_processor(content) from source_exception
-    elif default_exception:
+
+    if default_exception and isinstance(source_exception, BaseException):
         raise default_exception from source_exception
-    raise source_exception
+
+    if isinstance(source_exception, BaseException):
```
aniezurawski, line 41: Just make `source_exception` an `Optional` (`source_exception: Optional[Exception]`) and check if it's not None.

src/neptune/internal/backends/swagger_client_wrapper.py
```diff
@@ -94,6 +97,17 @@ def handle_neptune_http_errors(response, exception: Optional[HTTPError] = None):
             source_exception=exception,
         )
 
+        try:
+            response.raise_for_status()
+        except AttributeError:
+            pass
+
+        if exception is None:
```
aniezurawski, line 105: This if seems to be unnecessary.

src/neptune/internal/backends/swagger_client_wrapper.py
```diff
@@ -94,6 +97,17 @@ def handle_neptune_http_errors(response, exception: Optional[HTTPError] = None):
...
+            # raise generic HTTPError with response info
+            raise HTTPError(
```
aniezurawski, line 107: Please don't expose a bravado exception. ;)

How about using `NeptuneApiException`?

aniezurawski (changes requested):

src/neptune/api/exceptions_utils.py
```diff
@@ -35,8 +35,8 @@ def handle_json_errors(
     if error_processor:
         raise error_processor(content) from source_exception
 
-    if default_exception and isinstance(source_exception, BaseException):
+    if default_exception and source_exception is not None:
```
aniezurawski, line 38: We still may want to raise a `default_exception`. Just without a cause (`from`).

src/neptune/internal/backends/swagger_client_wrapper.py
```diff
@@ -102,11 +105,8 @@ def handle_neptune_http_errors(response, exception: Optional[HTTPError] = None):
         except AttributeError:
             pass
 
-        if exception is None:
-            # raise generic HTTPError with response info
-            raise HTTPError(
-                message=f"{response.status_code} Error: {response.reason} for {response.url}", response=response
-            )
+        # raise generic NeptuneApiException with response info
+        raise NeptuneApiException(f"{response.status_code} Error: {response.reason} for {response.url}")
```
aniezurawski, line 109: Again let's wrap it in `exept AttributeError` and just stringify all arguments if somthing goes wrong.

aniezurawski:
> @aniezurawski BTW Shouldn't we switch places of those two `if` statements?
> 
> ```
>     if response.status_code >= 300:
>         ApiMethodWrapper.handle_neptune_http_errors(response)
>     if response.status_code in (
>         HTTPUnprocessableEntity.status_code,
>         HTTPPaymentRequired.status_code,
>     ):
>         raise NeptuneLimitExceedException(reason=response.json().get("title", "Unknown reason"))
> ```
> 
> They have 422 and 402 codes.

Huh. Good point. In this case it look like `handle_neptune_http_errors` was meant to not raise anything if there is no source or default excpetion. Now I see the bug was introduced in `1dfa30a`. In that case we should just restore the old behaviour.

AleksanderWWW:
> > @aniezurawski BTW Shouldn't we switch places of those two `if` statements?
> > ```
> >     if response.status_code >= 300:
> >         ApiMethodWrapper.handle_neptune_http_errors(response)
> >     if response.status_code in (
> >         HTTPUnprocessableEntity.status_code,
> >         HTTPPaymentRequired.status_code,
> >     ):
> >         raise NeptuneLimitExceedException(reason=response.json().get("title", "Unknown reason"))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > They have 422 and 402 codes.
> 
> Huh. Good point. In this case it look like `handle_neptune_http_errors` was meant to not raise anything if there is no source or default excpetion. Now I see the bug was introduced in `1dfa30a`. In that case we should just restore the old behaviour.

I don't follow. What's then the point of calling this with `None` as exception if in that case it's not supposed to do anything? Or am I missing something?

aniezurawski (approved):

PR #1475, opened 2023-09-27
programmatically delete trashed objects

AleksanderWWW:


src/neptune/metadata_containers/project.py
```diff
@@ -275,6 +278,7 @@ def fetch_runs_table(
         owner: Optional[Union[str, Iterable[str]]] = None,
         tag: Optional[Union[str, Iterable[str]]] = None,
         columns: Optional[Iterable[str]] = None,
+        trashed: Optional[bool] = False,
```
Raalsky, line 281: IDE doesn't recognize it as documented in docstrings.

src/neptune/metadata_containers/project.py
```diff
@@ -406,7 +420,9 @@ def fetch_models_table(self, *, columns: Optional[Iterable[str]] = None) -> Tabl
                 name="sys/trashed",
                 type=NQLAttributeType.BOOLEAN,
                 operator=NQLAttributeOperator.EQUALS,
-                value=False,
-            ),
+                value=trashed,
+            )
+            if trashed is not None
+            else None,
```
Raalsky, line 426: I'm not sure here (needs to be checked) but probably `NQLEmptyQuery` seems to be more expected than `None`.

src/neptune/management/internal/api.py
```diff
@@ -914,3 +917,103 @@ def trash_objects(
 
     for error in errors:
         logger.warning(error)
+
+
+def delete_objects_from_trash(
+    project: str,
+    ids: Union[str, Iterable[str]],
+    *,
+    workspace: str = None,
```
Raalsky, line 926: Probably it should be `Optional[str]`.

src/neptune/management/internal/api.py
```diff
@@ -914,3 +917,103 @@ def trash_objects(
...
+
+    params = {
+        "projectIdentifier": project_qualified_name,
+        "experimentIdentifiers": None,
+        **DEFAULT_REQUEST_KWARGS,
+    }
+
+    qualified_name_ids = [QualifiedName(f"{workspace}/{project_name}/{container_id}") for container_id in ids]
+    for batch_ids in get_batches(qualified_name_ids, batch_size=TRASH_BATCH_SIZE):
+        params["experimentIdentifiers"] = batch_ids
```
Raalsky, line 976: Let's just prepare a `params` here (always from scratch) instead of updating a dictionary.

src/neptune/management/internal/api.py
```diff
@@ -914,3 +917,103 @@ def trash_objects(
...
+
+        Clearing trash:
+        >>> from neptune import management
+        >>> management.clear_trash(project="ml-team/classification")
+    """
+    verify_type("project", project, str)
+    verify_type("workspace", workspace, (str, type(None)))
+    verify_type("api_token", api_token, (str, type(None)))
+
+    ids = get_trashed_object_ids(project, api_token)
```
Raalsky, line 1014: This doesn't have a proper pagination to my knowledge and I thought that there was a special endpoint for cleaning all the entries in the trash. Would you mind checking that?

src/neptune/management/internal/api.py
```diff
@@ -914,3 +917,103 @@ def trash_objects(
...
+
+    if not ids:  # nothing to delete
+        return
+
+    delete_objects_from_trash(project, ids, workspace=workspace, api_token=api_token)
```
Raalsky, line 1019: Let's use keyword arguments internally as this will prevent us from future mistakes.

src/neptune/management/internal/utils.py
```diff
@@ -61,3 +64,21 @@ def to_api(self) -> str:
         if self.value == "admin":
             return "owner"
         return self.value
+
+
+def get_trashed_object_ids(project: Optional[str], api_token: Optional[str]) -> List[str]:
+    from neptune import init_project  # to avoid circular imports
+
+    with init_project(project=project, api_token=api_token) as project_obj:
```
Raalsky, line 72: Let's not use a high-level `neptune` API here. `management` and `neptune`  packages should be built on top of some common `api` package and be treated as a some kind of siblings. So let's use NQL building here. Especially that:
- this will prevent us from circular dependencies probably
- probably you'll be able to list all objects at once (runs & models), don't remember but I expect that it's possible

Raalsky (approved):

PR #1477, opened 2023-09-28
Add support for retry-after mechanism

asledz:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ### Features
 - Added support for callbacks that stop the synchronization if the lag or lack of progress exceeds a certain threshold ([#1478](https://github.com/neptune-ai/neptune-client/pull/1478))
-- Add support for `retry-after` header in HTTPTooManyRequests ([#1477](https://github.com/neptune-ai/neptune-client/pull/1477))
+
+### Changes
+- - Add support for `retry-after` header in HTTPTooManyRequests ([#1477](https://github.com/neptune-ai/neptune-client/pull/1477))
 
 ### Fixes
 - Add newline at the end of generated `.patch` while tracking uncommitted changes ([#1473](https://github.com/neptune-ai/neptune-client/pull/1473))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Features
 - Added support for `airflow` integration ([#1466](https://github.com/neptune-ai/neptune-client/pull/1466))
+- Add support for `retry-after` header in HTTPTooManyRequests ([#1477](https://github.com/neptune-ai/neptune-client/pull/1477))
 
 ### Changes
 - Add handling of project limits ([#1456](https://github.com/neptune-ai/neptune-client/pull/1456))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Add newline at the end of generated `.patch` while tracking uncommitted changes ([#1473](https://github.com/neptune-ai/neptune-client/pull/1473))
 - Clarify `NeptuneLimitExceedException` error message ([#1480](https://github.com/neptune-ai/neptune-client/pull/1480))
 
-## neptune 1.7.0
+## neptune 1.8.0
 
 ### Features
 - Added support for `airflow` integration ([#1466](https://github.com/neptune-ai/neptune-client/pull/1466))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,16 @@
 
 ### Features
 - Added support for callbacks that stop the synchronization if the lag or lack of progress exceeds a certain threshold ([#1478](https://github.com/neptune-ai/neptune-client/pull/1478))
+- Add support for `retry-after` header in HTTPTooManyRequests ([#1477](https://github.com/neptune-ai/neptune-client/pull/1477))
 
 ### Fixes
 - Add newline at the end of generated `.patch` while tracking uncommitted changes ([#1473](https://github.com/neptune-ai/neptune-client/pull/1473))
 - Clarify `NeptuneLimitExceedException` error message ([#1480](https://github.com/neptune-ai/neptune-client/pull/1480))
 
-## neptune 1.8.0
+## neptune 1.7.0
 
 ### Features
 - Added support for `airflow` integration ([#1466](https://github.com/neptune-ai/neptune-client/pull/1466))
-- Add support for `retry-after` header in HTTPTooManyRequests ([#1477](https://github.com/neptune-ai/neptune-client/pull/1477))
 
 ### Changes
 - Add handling of project limits ([#1456](https://github.com/neptune-ai/neptune-client/pull/1456))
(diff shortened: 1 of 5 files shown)


src/neptune/common/backends/utils.py
```diff
@@ -95,14 +95,22 @@ def wrapper(*args, **kwargs):
...
-                HTTPTooManyRequests,
                 HTTPInternalServerError,
                 NewConnectionError,
                 ChunkedEncodingError,
             ) as e:
                 time.sleep(min(2 ** min(10, retry), MAX_RETRY_TIME))
                 last_exception = e
                 continue
+            except HTTPTooManyRequests as e:
+                response_headers = e.response.json()["headers"]
```
Raalsky, line 106: Isin't it just a HTTP Header? This should probably be available with `.response.headers['<SOME-NAME>']`. Or I'm missing something :wink:

Raalsky, line 106: Are we really really sure that this `headers` exists in On-Prems as well?

src/neptune/common/backends/utils.py
```diff
@@ -95,14 +95,22 @@ def wrapper(*args, **kwargs):
...
+                if "retry-after" in response_headers:
+                    retry_after = int(response_headers["retry-after"][0])
+                    time.sleep(min(retry_after, MAX_RETRY_TIME))
+                else:
+                    time.sleep(min(2 ** min(10, retry), MAX_RETRY_TIME))
```
Raalsky, line 111: Not sure if it works for you but this could be probably simplified to something like:
```suggestion
                wait_time = int(headers["retry-after"][0]) if "retry-after" in headers else 2 ** min(10, retry)
                time.sleep(min(wait_time, MAX_RETRY_TIME))
```
And we should probably introduce some try-catching here to make the standard retry will be default even is case that we've introduced some breaking changes in the API.

Raalsky, line 111: `10` is a Magic number -> let's make a constant instead

src/neptune/common/backends/utils.py
asledz: Changed :D

src/neptune/common/backends/utils.py
asledz: This is pretty much how much it works, I changed it cause your solution looks much better!
Also, added try-catches, good catch 😄

asledz: True, it was like that before, but I reformatted it so that it is now always in constant 😇

asledz: This is pretty much how much it works, I changed it cause your solution looks much better!
Also, added try-catches, good catch 😄

src/neptune/common/backends/utils.py
asledz: You are right, I over complicated it here - changed! 😉

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 ### Features
 - Added support for `airflow` integration ([#1466](https://github.com/neptune-ai/neptune-client/pull/1466))
+- Add support for `retry-after` header in HTTPTooManyRequests ([#1477](https://github.com/neptune-ai/neptune-client/pull/1477))
```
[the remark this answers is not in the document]
Raalsky: I mean, wrong section :D not the name :wink:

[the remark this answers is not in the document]
asledz: Oh right, it was in a wrong place because of rebasing 🤦🏻‍♀️ Changed, sorry! 😅

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### Features
 - Added support for callbacks that stop the synchronization if the lag or lack of progress exceeds a certain threshold ([#1478](https://github.com/neptune-ai/neptune-client/pull/1478))
+- Add support for `retry-after` header in HTTPTooManyRequests ([#1477](https://github.com/neptune-ai/neptune-client/pull/1477))
```
Raalsky, line 5: Features are mostly user-exposed. This is probably a Change. What do you think?

src/neptune/common/backends/utils.py
```diff
@@ -95,12 +96,23 @@ def wrapper(*args, **kwargs):
...
                 NewConnectionError,
                 ChunkedEncodingError,
             ) as e:
-                time.sleep(min(2 ** min(10, retry), MAX_RETRY_TIME))
+                time.sleep(min(2 ** min(MAX_RETRY_MULTIPLIER, retry), MAX_RETRY_TIME))
+                last_exception = e
+                continue
+            except HTTPTooManyRequests as e:
+                try:
+                    wait_time = (
```
Raalsky, line 108: How about extraction of this logic into separate function? Maybe even something in `neptune.common` to be able to use it in both legacy and new client.

src/neptune/common/backends/utils.py
```diff
@@ -120,10 +132,21 @@ def wrapper(*args, **kwargs):
...
+                    last_exception = e
+                    continue
+                elif status_code == HTTPTooManyRequests.status_code:
+                    try:
+                        wait_time = (
+                            int(e.response.headers["retry-after"][0])
+                            if "retry-after" in e.response.headers
+                            else 2 ** min(MAX_RETRY_MULTIPLIER, retry)
+                        )
+                        time.sleep(min(wait_time, MAX_RETRY_TIME))
```
Raalsky, line 147: Can we use `time.sleep` only once?

Raalsky: Probably here we should only have something like `time.sleep(min(get_retry_from_headers_or_default(e.response), MAX_RETRY_TIME))`

asledz: This is a great idea, I added this to the code now! :D

src/neptune/common/backends/utils.py
asledz: Sounds very good, I actually put it in utils, if this is okay, cause it is a very specific use case! If you think common is a better place, lmk!

src/neptune/common/backends/utils.py
asledz: This is a great idea, I added this to the code now! :D

src/neptune/common/backends/utils.py
asledz: Sounds very good, I actually put it in utils, if this is okay, cause it is a very specific use case! If you think common is a better place, lmk!

src/neptune/common/backends/utils.py
asledz: This is a great idea, I added this to the code now! :D

src/neptune/common/backends/utils.py
asledz: Sounds very good, I actually put it in utils, if this is okay, cause it is a very specific use case! If you think common is a better place, lmk!

CHANGELOG.md
asledz: Oh, that makes sense! Moved 😄

src/neptune/common/backends/utils.py
Raalsky: Ohh, I now realized it was already in `common` module. Let's then leave it just as a function inside of the same file (`src/neptune/common/backends.utils.py`) especially that you had to redefine the constants. Let's just reuse it in legacy client.

Raalsky (approved):

PR #1484, opened 2023-10-05
DiskQueue and related proper typings, tests and minor styling updates

Raalsky:


pyproject.toml
```diff
@@ -327,7 +327,6 @@ module = [
     "neptune.internal.background_job",
     "neptune.internal.container_structure",
     "neptune.internal.credentials",
-    "neptune.internal.disk_queue",
```
BartoszPrusak, line 330: Old files are removed from toml, but new files aren't added?

tests/unit/neptune/new/internal/queue/test_disk_queue.py
```diff
@@ -0,0 +1,250 @@
...
+        ) as queue:
+            # given
+            obj = Obj(5, "test")
+            queue.put(obj)
+
+            # when
+            queue.flush()
+
+            # then
+            assert queue.get() == get_queue_element(obj, 1)
```
BartoszPrusak, line 48: Something that rarely comes up but bugs me: "expected" should be on the left, "actual" on the right.

At least that should be the case according to the message displayed when the test fails - it labels the expression on the left as "expected:" and the other as "actual:". I already wasted 15 minutes trusting these labels instead of triple-checking what each statement in a test assert means.

BartoszPrusak (approved):

pyproject.toml
Raalsky: This is a list of files to be excluded from type checking by `mypy`. This effectively means that from now we have to have proper typing across new files.

tests/unit/neptune/new/internal/queue/test_disk_queue.py
Raalsky: You're 100% right. Will change that in a moment.

Raalsky: Done

PR #1490, opened 2023-10-09
Small tweaks in error handling

aniezurawski:


src/neptune/internal/operation_processors/factory.py
```diff
@@ -57,6 +59,7 @@ def get_operation_processor(
             backend=backend,
             lock=lock,
             sleep_time=flush_period,
+            batch_size=int(os.environ.get(NEPTUNE_ASYNC_BATCH_SIZE) or "1000"),
```
Raalsky, line 62: This will not take any effect with current master 😉

aniezurawski: Why?

Raalsky (approved):

PR #1501, opened 2023-10-10
Draft: Implement exception-free mode for Metadata Containers

aniezurawski:


src/neptune/__init__.py
```diff
@@ -101,21 +101,52 @@
...
 apply_patches()
 
-init_run = Run
-init_model = Model
-init_model_version = ModelVersion
-init_project = Project
+
+def _init_safety(obj: MetadataContainer) -> Union[MetadataContainer, SafeContainer]:
+    safety_mode = os.environ.get(NEPTUNE_SAFETY_MODE)
+    if safety_mode is None or not bool(safety_mode) or str(safety_mode).lower() == "false" or int(safety_mode) == 0:
```
PatrykGala, line 129:
```
    if safety_mode is None or not bool(safety_mode) or str(safety_mode).lower() == "false" or int(safety_mode) == 0:
ValueError: invalid literal for int() with base 10: 'true'
```

src/neptune/metadata_containers/safe_container.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __getattr__(self, item: str) -> Any:
+        return getattr(self._api_method, item)
+
+
+class SafeContainer:
+    def __init__(self, container: MetadataContainer):
+        self._container = container
+
+    def __getattr__(self, item: str) -> Any:
```
PatrykGala, line 45:
```
    run["root0"] = "adas"
TypeError: 'SafeContainer' object does not support item assignment
```
it should implement:
```
    def __setitem__(self, key: str, value) -> None:
```

PR #1521, opened 2023-10-12
fix: Run _async_no_progress_callback and _async_lag_callback in new t…

PatrykGala:
…hread.

### DIFF
--- a/src/neptune/internal/operation_processors/async_operation_processor.py
+++ b/src/neptune/internal/operation_processors/async_operation_processor.py
@@ -153,7 +153,7 @@ class AsyncOperationProcessor(OperationProcessor):
 
         with self._lock:
             if not self._lag_exceeded:
-                self._async_lag_callback()
+                threading.Thread(target=self._async_lag_callback, daemon=True).start()
                 self._lag_exceeded = True
 
     def _check_no_progress(self) -> None:
@@ -162,7 +162,7 @@ class AsyncOperationProcessor(OperationProcessor):
 
         with self._lock:
             if self._should_call_no_progress_callback:
-                self._async_no_progress_callback()
+                threading.Thread(target=self._async_no_progress_callback, daemon=True).start()
                 self._should_call_no_progress_callback = False
 
     def _wait_for_queue_empty(self, initial_queue_size: int, seconds: Optional[float]) -> None:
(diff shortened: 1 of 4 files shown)


src/neptune/internal/operation_processors/async_operation_processor.py
```diff
@@ -162,7 +162,7 @@ def _check_no_progress(self) -> None:
 
         with self._lock:
             if self._should_call_no_progress_callback:
-                self._async_no_progress_callback()
+                threading.Thread(target=self._async_no_progress_callback, daemon=True).start()
```
PatrykGala, line 165: `daemon=True` False?

Raalsky: Are you sure that Python doesn't require to call `join()`? I heard that sometimes it may be needed (it wasn't Python). Stackoverflow says it's fine: https://stackoverflow.com/questions/49912013/python-what-if-we-call-thread-start-and-leave-it-without-join-or-close but just to be sure :wink:

Raalsky: > A daemon thread is a thread that runs in the background and does not prevent the program from exiting. In other words, if there are only daemon threads left in a Python program, the program will exit even if those threads are still running. Non-daemon threads, on the other hand, will keep the program alive until they complete their execution.

src/neptune/internal/operation_processors/async_operation_processor.py
```diff
@@ -153,7 +153,7 @@ def _check_lag(self) -> None:
```
Raalsky, line 153: CHANGELOG

src/neptune/internal/operation_processors/async_operation_processor.py
PatrykGala: With join the caller will wait and hold the lock indefinitely.

PatrykGala: I know the theory, all background jobs have `daemon=True`, and I think that callbacks shouldn't block the exit of program.

Raalsky: > All background jobs have daemon=True, and I think that callbacks shouldn't block the exit of program.

We should probably ask someone from Product Team here :<

Raalsky: > With join the caller will wait and hold the lock indefinitely.

Yeah, I know. I was thinking about joining at the end of a script but I've looked for something like that and it's fine that we're not calling join at all.

Raalsky: But let's proceed with this fix. This could be changed anytime.

Raalsky (approved):

PR #1529, opened 2023-10-16
Minor tweaks and tests refactors in neptune.cli

Raalsky:

### DIFF
--- a/src/neptune/cli/clear.py
+++ b/src/neptune/cli/clear.py
@@ -47,6 +47,8 @@ class ClearRunner(AbstractBackendRunner):
 
         ClearRunner.remove_sync_containers(path)
 
+        ClearRunner.remove_synced_data(container_manager=container_manager, synced_containers=synced_containers)
+
         offline_containers = get_offline_dirs(path)
         if clear_eventual and (offline_containers or unsynced_containers):
             self.log_junk_metadata(offline_containers, unsynced_containers)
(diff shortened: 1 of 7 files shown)


src/neptune/cli/clear.py
```diff
@@ -87,11 +89,17 @@ def remove_data(
...
     def remove_containers(paths: List[Path]) -> None:
         for path in paths:
-            try:
-                shutil.rmtree(path)
-                logger.info(f"Deleted: {path}")
-            except OSError:
-                logger.warn(f"Cannot remove directory: {path}")
+            if path.exists():
+                try:
+                    shutil.rmtree(path)
```
artsiomtserashkovich, line 102: What do you think about make it the same as https://github.com/neptune-ai/neptune-client/blob/fe39583f7aa580fa040f5b709946abcdeedae9ca/src/neptune/cli/clear.py#L64? (``ignoreErrors=True`` or left only one implementation)

Raalsky: That's completely different thing. Sync that's an Experiment (run, model etc.) created in `sync` mode (every operation = immediate request to server) vs async that wasn't cleaned up properly (mode `async` that synced all the operations but there are leftovers)

artsiomtserashkovich: In that particular case, I meant diff at error handling - in one case we are using the bultin flag to suppress and in another, we are using try...except -> what do you think about make it the same?

Raalsky: Hmmmm. We probably wanted here (I've copied and pasted it from some other cleanups) to be more explicit. With `sync` we're not losing anything etc and that's a "bulk" delete of all containers so quite different usage.

artsiomtserashkovich (approved):

PR #1542, opened 2023-10-23
add warnings for unsupported floats

asledz:


src/neptune/attributes/atoms/float.py
```diff
@@ -45,5 +50,16 @@ def assign(self, value: typing.Union[FloatVal, float, int], *, wait: bool = Fals
         if not isinstance(value, FloatVal):
             value = FloatVal(value)
 
+        if is_unsupported_float(value.value):
+            warn_once(
+                message=f"WARNING: The value you're trying to log is a nonstandard float value ({str(value.value)}) "
```
Raalsky, line 55: Remember to ask @normandy7 to review the warning content.

src/neptune/types/series/float_series.py
```diff
@@ -96,3 +103,16 @@ def unit(self):
 
     def __str__(self):
         return "FloatSeries({})".format(str(self.values))
+
+    def clear_of_unsupported_values(self):
+        cleared_values = []
```
Raalsky, line 108: list comprehension + filter is probably faster. Use `append` in Python only in rare scenarios.

tests/unit/neptune/new/test_stringify_unsupported.py
```diff
@@ -531,3 +533,20 @@ def test_integers_outside_32bits(self, run):
 
         with assert_no_warnings():
             run["data"] = data
+
+    def test_float_warnings(self, run):
```
Raalsky, line 537: Let's test it somewhere else, I don't think it's that related to stringify value

src/neptune/utils.py
```diff
@@ -14,8 +14,9 @@
 # limitations under the License.
 #
 """Utility functions to support ML metadata logging with neptune.ai."""
-__all__ = ["stringify_unsupported", "stop_synchronization_callback"]
+__all__ = ["stringify_unsupported", "stop_synchronization_callback", "is_unsupported_float"]
```
Raalsky, line 17: That's a module for user-exposed utils. This function is much more internal.

tests/unit/neptune/new/test_stringify_unsupported.py
```diff
@@ -531,3 +533,20 @@ def test_integers_outside_32bits(self, run):
...
+        with pytest.warns(NeptuneUnsupportedValue):
+            run["infinity"] = float("inf")
+            run["neg-infinity"] = float("-inf")
```
Raalsky, line 540: What will happen here? Do we log an operation to assign value or like in series we're going to not log a unsupported value?

src/neptune/types/series/float_series.py
asledz: Right! Changed that :D

tests/unit/neptune/new/test_stringify_unsupported.py
asledz: Moved the tests 😇

src/neptune/utils.py
asledz: True, deleted from here 😄

tests/unit/neptune/new/test_stringify_unsupported.py
asledz: Like in series:
- throw the warning to user, that is not a supported value
- skip the value

src/neptune/types/series/float_series.py
```diff
@@ -55,6 +60,8 @@ def __init__(
         self._max = max
         self._unit = unit
 
+        self.clear_of_unsupported_values()
```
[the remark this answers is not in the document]
asledz: Done

src/neptune/attributes/atoms/float.py
asledz: It is copied from previous PR, which she reviewed 😄 but asked her to doublecheck, that everything is correct

src/neptune/attributes/atoms/float.py
```diff
@@ -45,5 +50,16 @@ def assign(self, value: typing.Union[FloatVal, float, int], *, wait: bool = Fals
...
+                f"that is not currently supported. "
+                f"We'll add support for this type of value in the future. "
+                f"For now, try logging the value as stringify_unsupported: "
```
normandy7, line 58:
```suggestion
                f"For now, you can use utils.stringify_unsupported() to log one or more values as strings: "
```

normandy7 (dismissed):

CHANGELOG.md
```diff
@@ -21,7 +21,7 @@
 - Minor tweaks to `neptune.cli` and cleaning leftovers after async Experiments ([#1529](https://github.com/neptune-ai/neptune-client/pull/1529))
 - Pin `simplejson` required version to below `3.19` ([#1535](https://github.com/neptune-ai/neptune-client/pull/1535))
 - Added `experimental` mode that supports partitioned operations queue ([#1524](https://github.com/neptune-ai/neptune-client/pull/1524))
-
+- Warn about unsupported float values (infinity, negative infinity, NaN) in logging floats ([#1542](https://github.com/neptune-ai/neptune-client/pull/1542))
```
Raalsky, line 24: Skip unsupported float values?

asledz: True, that is missing -> changed to skip and warn (I think it is important, because it is not a silent skip)

tests/unit/neptune/new/attributes/series/test_float_series.py
```diff
@@ -51,3 +54,18 @@ def test_log(self):
             values = list(var.fetch_values()["value"].array)
             expected = list(range(0, 5000))
             self.assertEqual(len(set(expected)), len(set(values)))
+
+    def test_float_warnings(self):
+        run = init_run(mode="debug")
+
+        with pytest.warns(NeptuneUnsupportedValue):
+            run["train"].append({"supported_1": 1, "supported_2": 2})
```
Raalsky, line 62: Let's test multiple values to the same namespaces:
```python
>>> run['f'].extend([1.5, 2.3, str(float("nan")), 4.7])
```
or:
```python
>>> run['b'].extend([str(float("inf")), 5.5, 2.3, 6.7])
```

tests/unit/neptune/new/attributes/atoms/test_float.py
```diff
@@ -60,3 +63,12 @@ def test_get(self):
...
             self.assertEqual(5, var.fetch())
+
+    def test_float_warnings(self):
+        run = init_run(mode="debug")
+        with pytest.warns(NeptuneUnsupportedValue):
+            run["infinity"] = float("inf")
+            run["neg-infinity"] = float("-inf")
+            run["nan"] = float("nan")
+
+        run.stop()
```
Raalsky, line 74: Can we assert here that these particular namespaces weren't created?

src/neptune/utils.py
```diff
@@ -16,6 +16,7 @@
 """Utility functions to support ML metadata logging with neptune.ai."""
 __all__ = ["stringify_unsupported", "stop_synchronization_callback"]
```
Raalsky, line 17: Let's extend this list with your function

src/neptune/types/series/float_series.py
```diff
@@ -96,3 +103,18 @@ def unit(self):
 
     def __str__(self):
         return "FloatSeries({})".format(str(self.values))
+
+    def clear_of_unsupported_values(self):
+        cleared_values = [
```
Raalsky, line 108: How about `steps` and `timestamps` here? As we're getting rid of particular value we have to delete corresponding timestamp and step as well

tests/unit/neptune/new/attributes/series/test_float_series.py
asledz: Added 😄

tests/unit/neptune/new/attributes/atoms/test_float.py
asledz: Added! 😄

src/neptune/utils.py
asledz: Extended back 😅

src/neptune/types/series/float_series.py
asledz: True, missed that! both of those lists are now cleared 😄

tests/unit/neptune/new/attributes/atoms/test_float.py
Raalsky: `run.exists("infinity")`? `__dict__` is internal property and I'm not sure that our abstractions handle it properly

src/neptune/utils.py
Raalsky: Looks like it's still here :D.

asledz: Moved back because of the comment you left later on, but decided to move to internal common module 😉

src/neptune/types/series/float_series.py
```diff
@@ -96,3 +103,22 @@ def unit(self):
 
     def __str__(self):
         return "FloatSeries({})".format(str(self.values))
+
+    def clear_of_unsupported_values(self):
```
Raalsky, line 107: How about something more generic like:
```python
def filter_unsupported_values(values, steps, timestamps, filter_by):
    filtered = [
        (value, step, timestamp) for value, step, timestamp in zip(values, steps, timestamps) if filter_by(value)
    ]
    return (
        [value for value, _, _ in filtered],
        [step for _, step, _ in filtered],
        [timestamp for _, _, timestamp in filtered]
    )
```
and usage:
```python
        if steps is None:
            filled_steps = list(islice(cycle([None]), len(values)))
        else:
            assert len(values) == len(steps)
            filled_steps = steps

        if timestamps is None:
            filled_timestamps = list(islice(cycle([time.time()]), len(values)))
        else:
            assert len(values) == len(timestamps)
            filled_timestamps = timestamps

        clean_values, self._steps, self._timestamps = filter_unsupported_values(
            values=values,
            steps=filled_steps,
            timestamps=filled_timestamps,
            filter_by=is_unsupported_float
        )
        self._values = [float(value) for value in clean_values]
```

But it wasn't tested :D.

src/neptune/utils.py
Raalsky: Sorry for confusing you, one thing is the proper placing another is to remember about `__all__` filling up.

src/neptune/types/series/float_series.py
asledz: That looks pretty! But we still need to warn user about unsupported, please take a look if my solution for that looks as pretty as this 😇

src/neptune/attributes/atoms/float.py
```diff
@@ -45,5 +50,16 @@ def assign(self, value: typing.Union[FloatVal, float, int], *, wait: bool = Fals
         if not isinstance(value, FloatVal):
             value = FloatVal(value)
 
+        if is_unsupported_float(value.value):
+            warn_once(
+                message=f"WARNING: The value you're trying to log is a nonstandard float value ({str(value.value)}) "
+                f"that is not currently supported. "
+                f"We'll add support for this type of value in the future. "
+                f"For now, you can use utils.stringify_unsupported() to log one or more values as strings: "
```
Raalsky, line 58: That's not true :wink:. `stringify_unsupported` doesn't help here or I don't see how it should work with float series.

asledz: I mean, isn't it? it is applicable for atom floats. What might be a bit misleading maybe is "one or more values as strings" part, cause it is not useful in float_series. 
Is this is the part you say is not true? 😅 
(I am not the best with warning messages 😅 )

src/neptune/types/series/float_series.py
```diff
@@ -96,3 +103,22 @@ def unit(self):
...
+        values = [
+            value
+            if not is_unsupported_float(value)
+            else warn_once(
```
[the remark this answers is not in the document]
asledz: Yep, fixed with the generic result

src/neptune/attributes/atoms/float.py
normandy7: Since I contributed to the message: My assumption was that this doesn't show for float series, but rather if the user tries to log a single value (or collection of values, hence "one or more"), in which case `stringify_unsupported` could be used.

Raalsky: Keep in mind that `stringify_unsupported` means that if there will be some unsupported value (based on context as it's lazy-evaluated) it's going to be cased to a String. String may not be proper type in all of the context like -> trying to append a String value to FloatSeries.

To summarize about `stringify_unsupported`:
- It should work with atoms in most of the cases as `stringify_unsupported` is going to be either a proper value or a String; but in case there was a float value previously or something similar it may raise a MetadataInconsistency/ValueError as we're trying to do something similar to:
```python
>>> run['a'] = 4.2
>>> run['a'] = 'unsupported'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/neptune/Repos/neptune-client/src/neptune/metadata_containers/metadata_container.py", line 105, in inner_fun
    return fun(self, *args, **kwargs)
  File "/Users/neptune/Repos/neptune-client/src/neptune/metadata_containers/metadata_container.py", line 312, in __setitem__
    self.__getitem__(key).assign(value)
  File "/Users/neptune/Repos/neptune-client/src/neptune/handler.py", line 88, in inner_fun
    return fun(self, *args, **kwargs)
  File "/Users/neptune/Repos/neptune-client/src/neptune/handler.py", line 222, in assign
    attr.process_assignment(value, wait=wait)
  File "/Users/neptune/Repos/neptune-client/src/neptune/attributes/attribute.py", line 66, in process_assignment
    return self.assign(value, wait=wait)
  File "/Users/neptune/Repos/neptune-client/src/neptune/attributes/atoms/float.py", line 51, in assign
    value = FloatVal(value)
  File "/Users/neptune/Repos/neptune-client/src/neptune/types/atoms/float.py", line 39, in __init__
    self.value = float(extract_if_stringify_value(value))
ValueError: could not convert string to float: 'unsupported'
```
but I don't have a good example yet.

- It shouldn't work for float/image series as `.extend(stringify_unsupported([5.4, 6.7, float(inf), 7.8]))` in the base case will lead to something like `.extend([5.4, 6.7, "inf", 7.8])` so it's the same as trying to append a String to a FloatSeries (should lead to an exception like MetadataInconsistency)

Raalsky: Btw. Another thing that I spotted is that we're going to show double warning message with:
```python
>>> run['a'] = stringify_unsupported(float('inf'))
The value you're trying to log is an unsupported float value (inf) and will be logged as string. We'll add support for these types of values in the future.
<stdin>:1: NeptuneUnsupportedValue: WARNING: The value you're trying to log is a nonstandard float value (inf) that is not currently supported. We'll add support for this type of value in the future. For now, you can use utils.stringify_unsupported() to log one or more values as strings: run['field'] = stringify_unsupported(float(inf)
```
Is it expected?

src/neptune/common/utils.py
```diff
@@ -268,3 +268,9 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         pass
+
+
+def is_unsupported_float(value: float) -> bool:
```
Raalsky, line 273: `src/neptune/common` means that it's a part of both `neptune.legacy` and regular `neptune` packages. I don't think there are any usage of it in legacy client. Let's move it to something like `src/neptune/internal/types/utils.py`

src/neptune/types/series/float_series.py
```diff
@@ -96,3 +110,23 @@ def unit(self):
 
     def __str__(self):
         return "FloatSeries({})".format(str(self.values))
+
+    def warn_about_unsupported(self, values):
+        for value in values:
+            warn_once(
```
Raalsky, line 116: You can hide it under the `filter_by` implementation -> Instead of calling `is_unsupported_float` directly you may wrap it with something like:
```python
def is_unsupported_float_with_warn(x):
    if is_unsupported_float(x):
        warn_once(...)
        return ....
    return ...
```

asledz: Done!

src/neptune/common/utils.py
asledz: Moved

src/neptune/attributes/atoms/float.py
asledz: About the double warning -> 
true, not expected. But I was thinking, using stringify_unsupported should warn us once, so I deleted stringify_unsupported warning, now only the NeptuneUnsupportedValue will be displayed. 
I think it is a good solution for end user, @normandy7 and @Raalsky what do you think? 🤔

src/neptune/attributes/atoms/float.py
```diff
@@ -45,5 +50,16 @@ def assign(self, value: typing.Union[FloatVal, float, int], *, wait: bool = Fals
...
+                f"run['field'] = stringify_unsupported(float({str(value.value)})",
```
normandy7, line 59: Add missing closing parenthesis

```suggestion
                f"run['field'] = stringify_unsupported(float({str(value.value)}))",
```

CHANGELOG.md
```diff
@@ -22,7 +22,7 @@
 - Minor tweaks to `neptune.cli` and cleaning leftovers after async Experiments ([#1529](https://github.com/neptune-ai/neptune-client/pull/1529))
 - Pin `simplejson` required version to below `3.19` ([#1535](https://github.com/neptune-ai/neptune-client/pull/1535))
```
Raalsky, line 23: Could you remove this entry as well?

asledz: Removed!

src/neptune/attributes/atoms/float.py
asledz: Added

src/neptune/attributes/atoms/float.py
asledz: So now we would have:
Scenario 1: (single float, single warning)
 `run['a'] = float("inf")`
Result 1:
`NeptuneUnsupportedValue: WARNING: The value you're trying to log is a nonstandard float value (inf) that is not currently supported. We'll add support for this type of value in the future. For now, you can use utils.stringify_unsupported() to log one or more values as strings: run['field'] = stringify_unsupported(float(inf)
  run['a'] = float("inf")`
  
Scenario 2: (float in list, warning for float in list, logging correct float)
`run['f'].extend([float("inf"), 4.0])`
Result 2:
`NeptuneUnsupportedValue: WARNING: A value you're trying to log (`inf`) will be skipped because it's a non-standard float value that is not currently supported.
  run['f'].extend([float("inf"), 4.0])`
  
 Scenario 3 (stringify unsupported):
 `run['a'] = stringify_unsupported(float("inf"))`
 Result 3:
 No warning, logged as string -> and I think it makes sense, since this is what user wanted to do in the first place.

src/neptune/types/series/float_series.py
```diff
@@ -56,16 +61,24 @@ def __init__(
...
 
         if timestamps is None:
-            self._timestamps = cycle([time.time()])
+            filled_timestamps = cycle([time.time()])
         else:
             assert len(values) == len(timestamps)
-            self._timestamps = timestamps
+            filled_timestamps = timestamps
+
+        clean_values, self._steps, self._timestamps = self.filter_unsupported_values(
```
Raalsky, line 75: For not it's just to think of: Do you think we can refactor it someone to not to create an empty FloatSeries whenever someone will but just a single unsupported value?
I mean the following code produces an empty FloatSeries:
```python
>>> run['c'].append(float('nan'))
<stdin>:1: NeptuneUnsupportedValue: WARNING: A value you're trying to log (`nan`) will be skipped because it's a non-standard float value that is not currently supported.
>>> run.get_structure()['c']
<neptune.attributes.series.float_series.FloatSeries object at 0x13ff83d90>
```

asledz: I think it requires a bit more rework, I would leave it for a separate pr, after trying to make that change and not looking so obvious 🤔

src/neptune/attributes/atoms/float.py
Raalsky: I rather agree with the current behavior :wink:

src/neptune/types/series/float_series.py
Raalsky: Ok. Posted to Jira then

Raalsky (approved):

src/neptune/attributes/atoms/float.py
normandy7: For the record, I also agreed with the behavior 😄

Raalsky: > For the record

@asledz Just an FYI the general process is to wait for a double approval from some dev and @normandy7 if was involved/asked for review.

asledz: Oh sorry! Will wait in future, I just assumed a 👍🏻 emoji meant I am free to merging! 😅 Won't do that again 😇

PR #1545, opened 2023-10-26
Added support for plugins/extensions

Raalsky:


AleksanderWWW (approved):

AleksanderWWW:
Just a general non-blocking question.
Patches are located in `neptune.common` package. Extensions will be located in `neptune.internal`. Why is there a difference in location between those two?

Raalsky:
> Just a general non-blocking question.
> Patches are located in `neptune.common` package. Extensions will be located in `neptune.internal`. Why is there a difference in location between those two?

common in neptune.common means that package/module is shared between current and legacy neptune clients. We would like to extend the behaviour of the recent client.

PR #1549, opened 2023-10-30
Rework disk usage limitation

artsiomtserashkovich:


src/neptune/envs.py
```diff
@@ -31,8 +31,7 @@
     "NEPTUNE_FETCH_TABLE_STEP_SIZE",
     "NEPTUNE_SYNC_AFTER_STOP_TIMEOUT",
     "NEPTUNE_REQUEST_TIMEOUT",
-    "NEPTUNE_MAX_DISK_UTILIZATION",
-    "NEPTUNE_NON_RAISING_ON_DISK_ISSUE",
+    "NEPTUNE_LIMIT_DISK_UTILIZATION",
```
Raalsky, line 34: @normandy7 Any suggestions here? Probably we would like to document that. The idea is that Neptune will stop saving data after reaching some disk usage percentage OR any other disk-related issues.

normandy7: So it's an option that limits disk utilization, for multiple possible reasons, and you enable it by passing a valid disk usage %-value?

If this environment variable is not used and some problem occurs with saving the data, I guess that leads to an exception. What's the case when using this environment variable? The execution continues but Neptune will just print a warning and not do anything else (except continue syncing already logged/saved stuff with the servers, depending on the mode)?

artsiomtserashkovich: > So it's an option that limits disk utilization, for multiple possible reasons, and you enable it by passing a valid disk usage %-value?

Yes, to enable it, the user has to provide a percentage of max disk utilization (e.g. `NEPTUNE_LIMIT_DISK_UTILIZATION=42`), by default checking disk utilization is turned off. 

> If this environment variable is not used

As mentioned before, by default checking for disk usage is turned off

> some problem occurs with saving the data

As it will be turned off, the handling of errors will be as same as it's now

>  What's the case when using this environment variable? The execution continues but Neptune will just print a warning and not do anything else (except continue syncing already logged/saved stuff with the servers, depending on the mode)?

If the limit is defined, it will not allow to client use the disk when some threshold is reached. It will print a warning (https://github.com/neptune-ai/neptune-client/blob/6ebb77827ac3c3537dd6c50eeeff7ffbeaaa1c99/src/neptune/internal/utils/disk_utilization.py#L92) and not call method that's decorated with checking (now mostly decorated methods that's saving something on disk). Error in this case wont be thrown

normandy7: So from the end-user perspective, if the threshold is reached, the Neptune logging will just stop? I'm asking because I'm also not that familiar with the default behavior, as we don't have much documentation around disk usage mechanics.

A couple of other remarks/questions:

1. Since you pass the threshold and not True/False, maybe `NEPTUNE_DISK_UTILIZATION_LIMIT` would work better?
1. Does "utilization" mean something different than "usage" here? If there is no technical difference, `NEPTUNE_DISK_USAGE_LIMIT` (or `NEPTUNE_LIMIT_DISK_USAGE`) is maybe a bit simpler.

artsiomtserashkovich: > So from the end-user perspective, if the threshold is reached, the Neptune logging will just stop

If you mean logging of attributes data to server -> yes, when the threshold is reached logging will stop

> If there is no technical difference

Nope, in our there is no difference :) 
So, will rename it to `NEPTUNE_DISK_USAGE_LIMIT`

normandy7: I'm just hesitant about the variable's double role, as it both enables a mechanism and sets a threshold related to the mechanism. `NEPTUNE_DISK_USAGE_LIMIT` is a good name for setting the limit, but it does not say anything about the difference in behavior.

But it would be difficult to express all of this in a single variable name, and maybe this double role is not so strange to devs.

Raalsky: From what I know, we had some recommendations for clients about using this feature so we need to have a good reason for changing it. I like the idea but probably it will be hard to explain it properly in docs and probably a lot of ppl will be unhappy with changing some environment variables behavior.

src/neptune/internal/utils/disk_utilization.py
```diff
@@ -0,0 +1,102 @@
...
+                warn_once(
+                    "Encountered disk issue during utilization check. Neptune will not save your data.",
+                    exception=NeptuneWarning,
+                )
+                return
+
+            if current_utilization >= max_disk_utilization:
+                warn_once(
+                    f"Max disk utilization {max_disk_utilization}% exceeded with {current_utilization}. "
+                    + "Neptune will not be save your data.",
```
normandy7, line 93: Would propose to make this a bit clearer.

```suggestion
                    f"Disk usage is at {current_utilization}%, which exceeds the maximum allowed utilization "
                    + "of {max_disk_utilization}%. Neptune will not save your data.",
```

src/neptune/envs.py
artsiomtserashkovich: Revert to previous naming

normandy7: - `NEPTUNE_MAX_DISK_UTILIZATION` -- sounds fine (or `NEPTUNE_MAX_DISK_USAGE`).
- `NEPTUNE_NON_RAISING_ON_DISK_ISSUE` -- unless it follows some convention that I don't know of, it sounds a bit strange. How about `NEPTUNE_CONTINUE_ON_DISK_ISSUE`?

artsiomtserashkovich: @normandy7, I'd like to change it, but as I know clients already using it and it's main blocker for renaming. @Raalsky , correct me if I wrong

src/neptune/internal/utils/disk_utilization.py
```diff
@@ -0,0 +1,104 @@
...
+
+def get_max_disk_utilization_from_env() -> Optional[float]:
+    env_limit_disk_utilization = os.getenv(NEPTUNE_MAX_DISK_UTILIZATION)
+
+    if env_limit_disk_utilization is None:
+        return None
+
+    try:
+        limit_disk_utilization = float(env_limit_disk_utilization)
+        assert 0 < limit_disk_utilization <= 100
```
AleksanderWWW, line 65: As far as I know, using `assert` statements as an exception handling method is a bad practice. Can we change it to raise `ValueError` if the conditions are not met?

AleksanderWWW (changes requested):

src/neptune/internal/utils/disk_utilization.py
artsiomtserashkovich: As I see we are using `assert` in many different places. E.g. https://github.com/neptune-ai/neptune-client/blob/3bab4d6336d53af8d054962b8e44e991956a6c53/src/neptune/metadata_containers/model_version.py#L282

AleksanderWWW:
```
if limit_disk_utilization <= 0 or limit_disk_utilization > 100:
    raise ValueError
```

then it will be caught and handled in the `except clause`. What do you think?

AleksanderWWW: If someone launches our code with the `-o` flag then the `asserts` are not executed, which is dangerous
https://stackoverflow.com/questions/2830358/what-are-the-implications-of-running-python-with-the-optimize-flag

src/neptune/envs.py
normandy7: Ah, I understand now. It wasn't obvious to me that the scope of the PR changed. In that case, just re-request a review from me if there's something more I should look at :)

src/neptune/internal/utils/disk_utilization.py
artsiomtserashkovich: Done

AleksanderWWW (approved):

AleksanderWWW (approved):

PR #1567, opened 2023-11-15
No synchronization callbacks reworked

Raalsky:


src/neptune/metadata_containers/run.py
```diff
@@ -99,6 +99,9 @@
 from neptune.types.atoms.git_ref import GitRefDisabled
 from neptune.types.mode import Mode
 
+if TYPE_CHECKING:
```
artsiomtserashkovich, line 102: Is there any specific reason why import code that are related to signals wrapped by TYPE_CHECKING?

artsiomtserashkovich (dismissed):

src/neptune/metadata_containers/run.py
Raalsky: In general, this solves circular-imports when sometimes we're trying to import some parent objects just for mypy/typing requirements. Even if that's not a case it's a part of my personal practice to indicate that specific imports are just for typing and should be only triggered whenever `mypy` is processing the codebase.

src/neptune/internal/operation_processors/async_operation_processor.py
```diff
@@ -122,10 +121,11 @@ def _init_data_path(container_id: "UniqueId", container_type: "ContainerType") -
 
     @ensure_disk_not_overutilize
     def enqueue_operation(self, op: Operation, *, wait: bool) -> None:
-        self._last_version = self._queue.put(op)
+        if not self._accepts_operations:
+            warn_once("Not accepting operations", exception=NeptuneWarning)
```
Raalsky, line 125: @normandy7 IMO that's the only thing for you here. This will be triggered in case that callback with `run.stop()` inside was called (no progress/lag callback) and at the same time someone is still trying to log more metadata. It's quite general but I don't know any more cases that this could happen but probably they exist.

artsiomtserashkovich (approved):

src/neptune/internal/operation_processors/async_operation_processor.py
Raalsky: I was asked to force this PR further. Let me know after all if you have any suggestions here.

normandy7: @Raalsky You mean there could be cases that are unrelated to the lag/no-progress callback? But the general scenario is that the run object is available in the session, it's just no longer active?

PR #1569, opened 2023-11-17
Added timestamp of operation put to disk queue

Raalsky:

### DIFF
--- a/src/neptune/internal/disk_queue.py
+++ b/src/neptune/internal/disk_queue.py
@@ -23,6 +23,7 @@ import threading
 from dataclasses import dataclass
 from glob import glob
 from pathlib import Path
+from time import time
 from typing import (
     Callable,
     Generic,
@@ -47,6 +48,7 @@ class QueueElement(Generic[T]):
     obj: T
     ver: int
     size: int
+    at: Optional[float] = None
 
 
 class DiskQueue(Generic[T]):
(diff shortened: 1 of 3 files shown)


src/neptune/internal/disk_queue.py
```diff
@@ -47,6 +48,7 @@ class QueueElement(Generic[T]):
     obj: T
     ver: int
     size: int
+    at: Optional[float] = None
```
artsiomtserashkovich, line 51: I found it could cause some confusing. What do you think about change type from `float` to `datetime` here?

artsiomtserashkovich: We could store time in the same format, but deserialize to more "high-level" type, cause for me float looks like is more about implementation detail

Raalsky: Hmm. But it's harder to serialize/deserialize probably. Maybe we can introduce just an alias like `Timestamp` for float here.

artsiomtserashkovich: FYI: datetime has bultin method to convert from/to timestamp - https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp

Raalsky: I've added a type alias for timestamps. I've been trying to introduce `datetime`-s but it's not that straight forward and I have to worry a lot about proper timezones and it was harder to test it.

artsiomtserashkovich (approved):

PR #1571, opened 2023-11-20
download hook

AleksanderWWW:


src/neptune/internal/backends/hosted_file_operations.py
```diff
@@ -460,18 +467,29 @@ def _get_download_url(swagger_client: SwaggerClientWrapper, download_id: str):
...
         target_file = destination
+
+    total_size = int(response.headers.get("content-length", 0))
+    pre_download_hook(total_size)
     with response:
         with open(target_file, "wb") as f:
             for chunk in response.iter_content(chunk_size=1024 * 1024):
                 if chunk:
                     f.write(chunk)
+                download_iter_hook(len(chunk))
```
Raalsky, line 491: `len(chunk)` is risky outside of `if chunk:`. Use `len(chunk) if chunk else 0` or pass `chunk` to the hook directly. Not sure if `iter_content` could return `None` or something similar that will break `len(chunk)` with some `ValueError`

AleksanderWWW: Good point

Raalsky (approved):

Raalsky (approved):

PR #1587, opened 2023-12-06
Matched `fetch_*_table()` with UI capabilities

Raalsky:

### DIFF
--- a/src/neptune/internal/backends/hosted_neptune_backend.py
+++ b/src/neptune/internal/backends/hosted_neptune_backend.py
@@ -48,6 +48,10 @@ from neptune.common.exceptions import (
 )
 from neptune.common.patterns import PROJECT_QUALIFIED_NAME_PATTERN
 from neptune.core.components.operation_storage import OperationStorage
+from neptune.common.warnings import (
+    NeptuneWarning,
+    warn_once,
+)
 from neptune.envs import NEPTUNE_FETCH_TABLE_STEP_SIZE
 from neptune.exceptions import (
     AmbiguousProjectName,
@@ -149,6 +153,15 @@ if TYPE_CHECKING:
 
 
 _logger = get_logger()
+
+ATOMIC_ATTRIBUTE_TYPES = {
+    AttributeType.INT.value,
+    AttributeType.FLOAT.value,
+    AttributeType.STRING.value,
+    AttributeType.BOOL.value,
+    AttributeType.DATETIME.value,
+    AttributeType.RUN_STATE.value,
+}
 
 
 class HostedNeptuneBackend(NeptuneBackend):
@@ -1016,6 +1029,20 @@ class HostedNeptuneBackend(NeptuneBackend):
             raise FetchAttributeNotFoundException(path_to_str(path))
 
     @with_api_exceptions_handler
+    def _get_column_types(self, project_id: UniqueId, column: str, types: Optional[Iterable[str]] = None) -> List[Any]:
+        params = {
+            "projectIdentifier": project_id,
+            "search": f"/^{column}$/",  # exact regex match
+            "type": types,
+            "params": {},
+            **DEFAULT_REQUEST_KWARGS,
+        }
+        try:
+            return self.leaderboard_client.api.searchLeaderboardAttributes(**params).response().result.entries
+        except HTTPNotFound as e:
+            raise ProjectNotFound(project_id=project_id) from e
+
+    @with_api_exceptions_handler
     def search_leaderboard_entries(
         self,
         project_id: UniqueId,
@@ -1023,6 +1050,7 @@ class HostedNeptuneBackend(NeptuneBackend):
         query: Optional[NQLQuery] = None,
         columns: Optional[Iterable[str]] = None,
         limit: Optional[int] = None,
+        sort_by: str = "sys/creation_time",
     ) -> Generator[LeaderboardEntry, None, None]:
         default_step_size = int(os.getenv(NEPTUNE_FETCH_TABLE_STEP_SIZE, "100"))
 
@@ -1030,6 +1058,12 @@ class HostedNeptuneBackend(NeptuneBackend):
 
         types_filter = list(map(lambda container_type: container_type.to_api(), types)) if types else None
         attributes_filter = {"attributeFilters": [{"path": column} for column in columns]} if columns else {}
+
+        if sort_by == "sys/creation_time":
+            sort_by_column_type = AttributeType.DATETIME.value
+        else:
+            sort_by_column_type_candidates = self._get_column_types(project_id, sort_by, types_filter)
+            sort_by_column_type = _get_column_type_from_entries(sort_by_column_type_candidates, sort_by)
 
         try:
             return iter_over_pages(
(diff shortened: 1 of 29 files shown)


src/neptune/metadata_containers/project.py
```diff
@@ -281,9 +287,16 @@ def fetch_runs_table(
...
+    def fetch_models_table(
+        self,
+        *,
+        columns: Optional[Iterable[str]] = None,
+        trashed: Optional[bool] = False,
+        limit: Optional[int] = None,
+    ) -> Table:
         """Retrieve models stored in the project.
 
         Args:
```
Raalsky, line 302: @normandy7 I've just discovered that `Examples` under `columns` makes the whole docstrings unrecognizable in PyCharm. That's not for now but we should probably change that to something like `Sample usages`.

src/neptune/metadata_containers/model.py
```diff
@@ -265,6 +276,12 @@ def fetch_model_versions_table(self, *, columns: Optional[Iterable[str]] = None)
                     Fields: `["params/lr", "params/batch", "val/acc"]` - these fields are included as columns.
                     Namespaces: `["params", "val"]` - all the fields inside the namespaces are included as columns.
                 If `None` (default), all the columns of the model versions table are included.
+            limit: How many entries to return at most (default: None - return all entries).
```
normandy7, line 279:
```suggestion
            limit: How many entries to return at most. If `None`, all entries are returned.
```

src/neptune/metadata_containers/model.py
```diff
@@ -265,6 +276,12 @@ def fetch_model_versions_table(self, *, columns: Optional[Iterable[str]] = None)
...
+            sort_by: Name of the column to sort the results by.
+                Must be an atomic column (string, float, datetime, integer, boolean), otherwise raises `ValueError`.
+                Default: 'sys/creation_time.
```
normandy7, line 282: 1. I think we should use "field" here instead of "column". In the UI, column names can be changed arbitrarily, but it has no bearing on the underlying field name.
1. See the previous comment about "atomic".
1. IMO no need to mention errors raised, as it represents a violation of the API in this case.
1. No need to mention defaults in the docstring. (We're not entirely consistent here yet; that's on me)

```suggestion
            sort_by: Name of the field to sort the results by.
                The field type must be simple (string, float, datetime, integer, or Boolean).
```

src/neptune/metadata_containers/model.py
```diff
@@ -265,6 +276,12 @@ def fetch_model_versions_table(self, *, columns: Optional[Iterable[str]] = None)
...
+            ascending: Whether to sort model versions in the ascending order of the sorting column values.
+                Default: False - descending order.
```
normandy7, line 284:
```suggestion
            ascending: Whether to sort model versions in ascending order of the sorting column values.
```

src/neptune/metadata_containers/model.py
```diff
@@ -255,7 +258,15 @@ def get_url(self) -> str:
...
 
-    def fetch_model_versions_table(self, *, columns: Optional[Iterable[str]] = None) -> Table:
+    def fetch_model_versions_table(
+        self,
+        *,
+        columns: Optional[Iterable[str]] = None,
+        limit: Optional[int] = None,
+        sort_by: str = "sys/creation_time",
+        ascending: bool = False,
+        progress_bar: Optional[Union[bool, Type[ProgressBarCallback]]] = None,
```
normandy7, line 268: Should `progress_bar` not be documented?

normandy7 (commented):
Some questions/comments on `fetch_model_versions_table()` that would also apply to the other fetching methods

src/neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -1063,3 +1106,28 @@ def get_model_version_url(
...
+            continue
+        if entry.type not in ATOMIC_ATTRIBUTE_TYPES:  # non-atomic type - no need to look further
+            raise ValueError(f"Column {column} used for sorting is not of atomic type.")
+        types.add(entry.type)
+
+    if types == {AttributeType.INT.value, AttributeType.FLOAT.value}:
+        return AttributeType.FLOAT.value
+
+    warn_once(
+        f"Column {column} contains more than one atomic data type. Sorting result might be inaccurate.",
```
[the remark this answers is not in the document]
Raalsky: I think that we can use "simple" instead. That's the same.

src/neptune/metadata_containers/model.py
Raalsky: We probably should document the `ProgressBarCallback` interface somehow.

src/neptune/internal/backends/hosted_neptune_backend.py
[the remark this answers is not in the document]
Raalsky: Is it fine to use phrasing "not of simple type" or should we introduce something like `complex`?

src/neptune/metadata_containers/model.py
Raalsky: 2-4 should be done :wink:. I'm struggling with 1. as you may provide namespaces instead of just fields and the full description is quite hard for me to rephrase.

src/neptune/internal/backends/hosted_neptune_backend.py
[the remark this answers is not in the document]
Raalsky: Done I think

src/neptune/metadata_containers/model.py
normandy7: Right, just making sure since it was missing from the docstrings' Args list.

I'd offer to contribute directly, but I don't know how this ProgressBar is supposed to work.

src/neptune/metadata_containers/model.py
normandy7: I'd like your take on `1` (field vs column) but I can of course submit proposals directly once the PR is mature enough.

src/neptune/metadata_containers/model.py
Raalsky: @AleksanderWWW Would you mind?

src/neptune/internal/backends/hosted_neptune_backend.py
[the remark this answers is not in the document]
Raalsky: Missed comment above, applied :wink: Or at least I think so.

src/neptune/metadata_containers/model.py
AleksanderWWW: :eyes:

AleksanderWWW: @normandy7 will try to propose docstrings for the `ProgressBarCallback` in this PR today

AleksanderWWW: Added an example using `click`, though I don't have a strong opinion if such a full example is necessary, but thought that it might make it easier to understand what the callback is about. Let me know what you think please

normandy7: Thanks, will do by EOD!

normandy7: The example was helpful, thank you. Maybe it's not "necessary" to include in the docstring in full, but I don't think it hurts.

I'll push some suggestions shortly; do let me know if you have any remarks.

normandy7 (approved):

AleksanderWWW (approved):
# :partying_face: :champagne:

PR #1593, opened 2023-12-08
Add limit parameter to fetch_*_table

AleksanderWWW:

### DIFF
--- a/src/neptune/internal/backends/hosted_neptune_backend.py
+++ b/src/neptune/internal/backends/hosted_neptune_backend.py
@@ -1022,8 +1022,11 @@ class HostedNeptuneBackend(NeptuneBackend):
         types: Optional[Iterable[ContainerType]] = None,
         query: Optional[NQLQuery] = None,
         columns: Optional[Iterable[str]] = None,
+        limit: Optional[int] = None,
     ) -> Generator[LeaderboardEntry, None, None]:
         step_size = int(os.getenv(NEPTUNE_FETCH_TABLE_STEP_SIZE, "100"))
+
+        step_size = min(step_size, limit) if limit else step_size
 
         types_filter = list(map(lambda container_type: container_type.to_api(), types)) if types else None
         attributes_filter = {"attributeFilters": [{"path": column} for column in columns]} if columns else {}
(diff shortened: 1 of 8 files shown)


src/neptune/metadata_containers/metadata_container.py
```diff
@@ -665,6 +667,8 @@ def _fetch_entries(self, child_type: ContainerType, query: NQLQuery, columns: Op
             columns=columns,
         )
 
+        leaderboard_entries = itertools.islice(leaderboard_entries, limit) if limit else leaderboard_entries
```
Raalsky, line 670: I think that it's beneficial to still pass it to the pagination function as the limit may be much smaller than `step_size`. With the current approach, we're going to fetch for instance `100` entries even though the `limit` was set to `1`.

Raalsky: You can probably just set `step_size` to a minimum of the `limit` and the old step_size. It should work as well.

src/neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -1022,9 +1022,12 @@ def search_leaderboard_entries(
         types: Optional[Iterable[ContainerType]] = None,
         query: Optional[NQLQuery] = None,
         columns: Optional[Iterable[str]] = None,
+        limit: Optional[int] = None,
     ) -> Generator[LeaderboardEntry, None, None]:
         step_size = int(os.getenv(NEPTUNE_FETCH_TABLE_STEP_SIZE, "100"))
```
Raalsky, line 1027: Let's rename it to `default_step_size` then.

Raalsky:
@normandy7 we have a feature branch called `dev/fetch-ui-parity` that is going to have all the new parameters of `fetch_*_table` methods. What do you think about making just a one final docstrings review/adjustments after all of the parts will be merged?

normandy7:
> @normandy7 we have a feature branch called `dev/fetch-ui-parity` that is going to have all the new parameters of `fetch_*_table` methods. What do you think about making just a one final docstrings review/adjustments after all of the parts will be merged?

@Raalsky Alright, that works. I'd like to understand the reasoning for "step size", as it's not entirely clear to me.

Raalsky (approved):

PR #1617, opened 2024-01-18
add neptune backend protobuf calls for series values fetching.

szysad:
Additionally:
  - add unit test for new call. Includes some refractoring as new calls are ment to be new implementation of already exisiting calls.
  - add protobuf python compliled code with stubs in neptune.internal.proto with additional unit tests
  - pin types-protobuf in mypy pre-commit hook as a workaround to fresh protobuf-typing bug. More info: https://github.com/python/mypy/issues/16767
  - excluded proto files from flake8 check as they are output of protoc compiler and far from perfect syntax
  - install tensorflow dev dependency only if python version >=3.10
  - skip executing unit tests that require tensorflow if it's not installed. 

- [ ] <s>Did you **update the CHANGELOG**? (not for test updates, internal changes/refactors or CI/CD setup)</s>
- [ ] <s>Did you **ask the docs owner** to review all the user-facing changes?</s>


tests/unit/neptune/new/internal/backends/test_neptune_backend_mock.py
```diff
@@ -270,7 +268,13 @@ def test_get_string_series_values(self):
                     ret,
                 )
 
-    def test_get_float_series_values(self):
+    def test_get_string_series_values(self):
```
szysad, line 271: rename `test_get_string_series_values ` to `test_get_float_series_values`

pyproject.toml
```diff
@@ -212,6 +213,10 @@ force_grid_wrap = 2
 [tool.flake8]
 max-line-length = 120
 extend-ignore = "E203"
+exclude = [
+    # protobuf compiler output
+    "src/neptune/internal/proto/protobuf_api_series_dto_pb2.py"
```
Raalsky, line 218: Does "just" `src/neptune/internal/proto/` works? It's probably more future-proof

src/neptune/internal/backends/neptune_backend.py
```diff
@@ -211,6 +211,12 @@ def get_float_series_attribute(
     ) -> FloatSeriesAttribute:
         pass
 
+    @abc.abstractmethod
+    def get_float_series_values_proto(
```
Raalsky, line 215: I think this is not needed (no point of extending the abstract class/interface) as this is a part of implementation detail what protocol/endpoint to use. Could we just have two private methods as a part of `HostedNeptuneBackend` and keep the interface the same as it was before?

szysad: Yes we can do that. The only issue with that it requires client config support flag which was out of scope of this PR.

For now I have implemented proto support bloating `NeptuneBackend` method pool in next PR (attached on top of this one) https://github.com/neptune-ai/neptune-client/pull/1619

```
HostedNeptuneBackend now for both fetching string and float series values has methods:
get_*_series_values_json: that works as before with json dto
get_*_series_values_proto: with new implementation using protobuf
get_*_series_values: chooses proto or json implementation based on client config
```

Raalsky: > This change requires client config proto support flag which was out of scope of this PR

You can just add a private method in `HostedNeptuneBackend` without any calls of it for now. No need to worry about client config in this PR as you already mentioned.

pyproject.toml
szysad: Yep, good catch!

src/neptune/internal/backends/neptune_backend.py
szysad: At the end I just implmented `*_proto` methods as private methods of `HostedNeptuneBacked` without changing `NeptuneBackend` interface

src/neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -981,6 +1006,31 @@ def get_float_series_values(
...
+        limit: int,
+    ) -> FloatSeriesValues:
+        params = {
+            "experimentId": container_id,
+            "attribute": path_to_str(path),
+            "limit": limit,
+            "offset": offset,
+        }
+        try:
+            result = self.leaderboard_client.api.getFloatSeriesValuesProto(**params).response().result
```
Raalsky, line 1025: I think this is not going to work with the final version of our Swagger. @asledz what version are we currently using? The one with protobuf-related classes or the clean one?

szysad: Me and @asledz made some tests and verified it works fine with the final swagger version 🚀

szysad:
✅ E2E
https://github.com/neptune-ai/neptune-client/actions/runs/7655899511

PR #1620, opened 2024-01-23
Use progress bar in downloading File and FileSet

AleksanderWWW:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Added `sort_by` parameter to `fetch_*_table()` methods ([#1595](https://github.com/neptune-ai/neptune-client/pull/1595))
 - Added `ascending` parameter to `fetch_*_table()` methods ([#1602](https://github.com/neptune-ai/neptune-client/pull/1602))
 - Added `progress_bar` parameter to `fetch_*_table()` methods ([#1599](https://github.com/neptune-ai/neptune-client/pull/1599))
+- Added `progress_bar` parameter to `download` method of the `Handler` class ([#1620](https://github.com/neptune-ai/neptune-client/pull/1620))
 
 ### Fixes
 - Add direct requirement of `typing-extensions` ([#1586](https://github.com/neptune-ai/neptune-client/pull/1586))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Added `sort_by` parameter to `fetch_*_table()` methods ([#1595](https://github.com/neptune-ai/neptune-client/pull/1595))
 - Added `ascending` parameter to `fetch_*_table()` methods ([#1602](https://github.com/neptune-ai/neptune-client/pull/1602))
 - Added `progress_bar` parameter to `fetch_*_table()` methods ([#1599](https://github.com/neptune-ai/neptune-client/pull/1599))
-- Added `progress_bar` parameter to `download` method of the `Handler` class ([#1620](https://github.com/neptune-ai/neptune-client/pull/1620))
+- Added `progress_bar` parameter to `download()` method of the `Handler` class ([#1620](https://github.com/neptune-ai/neptune-client/pull/1620))
 
 ### Fixes
 - Add direct requirement of `typing-extensions` ([#1586](https://github.com/neptune-ai/neptune-client/pull/1586))
--- a/src/neptune/attributes/atoms/artifact.py
+++ b/src/neptune/attributes/atoms/artifact.py
@@ -62,7 +62,7 @@ class Artifact(Atom):
             artifact_hash,
         )
 
-    def download(self, destination: str = None):
+    def download(self, destination: str = None, **kwargs):
         self._check_feature()
         for file_definition in self.fetch_files_list():
             driver: typing.Type[ArtifactDriver] = ArtifactDriversMap.match_type(file_definition.type)
--- a/src/neptune/attributes/atoms/artifact.py
+++ b/src/neptune/attributes/atoms/artifact.py
@@ -63,7 +63,7 @@ class Artifact(Atom):
             artifact_hash,
         )
 
-    def download(self, destination: str = None, progress_bar: ProgressBarType = None):
+    def download(self, destination: str = None, progress_bar: typing.Optional[ProgressBarType] = None):
         self._check_feature()
         for file_definition in self.fetch_files_list():
             driver: typing.Type[ArtifactDriver] = ArtifactDriversMap.match_type(file_definition.type)
--- a/src/neptune/attributes/atoms/artifact.py
+++ b/src/neptune/attributes/atoms/artifact.py
@@ -30,6 +30,7 @@ from neptune.internal.operation import (
     TrackFilesToArtifact,
 )
 from neptune.types.atoms.artifact import Artifact as ArtifactVal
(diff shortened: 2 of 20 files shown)


src/neptune/attributes/atoms/artifact.py
```diff
@@ -62,7 +62,7 @@ def fetch_files_list(self) -> typing.List[ArtifactFileData]:
             artifact_hash,
         )
 
-    def download(self, destination: str = None):
+    def download(self, destination: str = None, **kwargs):
```
AleksanderWWW, line 65: This is needed due to how `download` works in `Handler`

AleksanderWWW:
### Additionally:
- Added a type alias `ProgressBarType` to save some typing
- Created a function to construct a progress bar given a progress bar type and description string

AleksanderWWW:
### E2E :white_check_mark: 
https://github.com/neptune-ai/neptune-client/actions/runs/7641402558

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Added `sort_by` parameter to `fetch_*_table()` methods ([#1595](https://github.com/neptune-ai/neptune-client/pull/1595))
 - Added `ascending` parameter to `fetch_*_table()` methods ([#1602](https://github.com/neptune-ai/neptune-client/pull/1602))
 - Added `progress_bar` parameter to `fetch_*_table()` methods ([#1599](https://github.com/neptune-ai/neptune-client/pull/1599))
+- Added `progress_bar` parameter to `download` method of the `Handler` class ([#1620](https://github.com/neptune-ai/neptune-client/pull/1620))
```
normandy7, line 10:
```suggestion
- Added `progress_bar` parameter to `download()` method of the `Handler` class ([#1620](https://github.com/neptune-ai/neptune-client/pull/1620))
```

src/neptune/handler.py
```diff
@@ -612,11 +617,16 @@ def download(self, destination: str = None) -> None:
                 composed from field name and extension (if present).
                 If `destination` is a path to a file, the file will be downloaded under the specified name.
                 Defaults to `None`.
+            progress_bar: (bool or Type of progress bar, optional): progress bar to be used while downloading assets.
+                If `None` or `True` the default tqdm-based progress bar will be used.
+                If `False` no progress bar will be used.
+                If a type of progress bar is passed, it will be used instead of the default one.
```
normandy7, line 623: How would you recommend to instruct the end-user on how to accomplish this? Is it correct to say that they can define a class that inherits from `ProgressBarCallback` and pass its type as the argument? (Asking also for docs purposes)

AleksanderWWW: Yes, just like in the example with `click` in the docstrings. You can pass this `ClickProgressBar`.
Maybe link it somehow?

normandy7: Very good; your example did make it super clear. Just wanted to double-check that I've got it right and that this alias didn't change anything on the user-end.

src/neptune/typing.py
```diff
@@ -80,3 +84,6 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
     @abc.abstractmethod
     def update(self, *, by: int, total: Optional[int] = None) -> None:
         ...
+
+
+ProgressBarType: TypeAlias = Optional[Union[bool, Type[ProgressBarCallback]]]
```
szysad, line 89: I suggest unwrapping `ProgressBarType` from `Optional` as its kinda unusual practice. I think that

```
def f(x: Optional[ProgressBarType] = None)
    pass
```
is more clear then

```
def f(x: ProgressBarType = None)
    pass
```

src/neptune/handler.py
```diff
@@ -595,7 +596,11 @@ def delete_files(self, paths: Union[str, Iterable[str]], *, wait: bool = False)
         return self._pass_call_to_attr(function_name="delete_files", paths=paths, wait=wait)
 
     @check_protected_paths
-    def download(self, destination: str = None) -> None:
+    def download(
+        self,
+        destination: str = None,
```
szysad, line 601: `destination` should have type `Optional[str]` if you are allowed to pass None.
Not sure why mypy is passing on that

src/neptune/internal/backends/hosted_file_operations.py
```diff
@@ -482,14 +479,13 @@ def _store_response_as_file(
...
-                    f.write(chunk)
-                download_iter_hook(len(chunk) if chunk else 0)
-    post_download_hook()
+    with construct_progress_bar(progress_bar, "Fetching fileset...") as bar:
+        with response:
+            with open(target_file, "wb") as f:
+                for chunk in response.iter_content(chunk_size=1024 * 1024):
+                    if chunk:
+                        f.write(chunk)
+                        bar.update(by=len(chunk), total=total_size)
```
szysad, line 488: its getting a bit nested. There is cool python trick, you can actually use `with` with multiple context managers with only one indent :))

```
    total_size = int(response.headers.get("content-length", 0))
    with (
        construct_progress_bar(progress_bar, "Fetching file...") as bar,
        response,
        open(target_file, "wb") as f
    ):
        for chunk in response.iter_content(chunk_size=1024 * 1024):
            if chunk:
                f.write(chunk)
                bar.update(by=len(chunk), total=total_size)
```

its added in python 3.10 so not sure if we can use it tho :(

but there are other ways around it in pre 3.10 versions like

```
with A() as a, \
     B() as b, \
     C() as c:
    doSomething(a,b,c)
```

more info: https://stackoverflow.com/questions/893333/multiple-variables-in-a-with-statement
python 3.10 doc: https://docs.python.org/3/whatsnew/3.10.html#parenthesized-context-managers

szysad (approved):
Looks good!
All my comments are code quality related so nothing blocking :)

src/neptune/internal/backends/hosted_file_operations.py
AleksanderWWW: Thought of it too, but unfortunately, it is not compatible with lower python versions, so we cannot support that. I can leave a TODO though to update it once we have 3.10 as the minimum version, what do you think?
Also, a workaround here could be to use `try-finally` syntax, but I'm not certain if we need that so much

src/neptune/handler.py
AleksanderWWW: Interesting. I see that it's been here for a long time, so yeah - weird. Worth looking into that. Thanks for noticing! Won't hurt to correct it anyway I guess, so let me do that quickly ;)

src/neptune/typing.py
AleksanderWWW: Good point. Done :)

src/neptune/internal/backends/hosted_file_operations.py
szysad: yep I think moving from python 3.7 to newer versions is crucial as we will keep running into issues like that.

1-st seems good, I'm not big fan of `try-finally` approach (its only nesting at the end).
How about [ExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) have you seen that? Its supported since python 3.3

AleksanderWWW: Didn't know that. Looks promising :)

szysad (approved):
LGTM! 🚀

PR #1634, opened 2024-02-06
Parse `datetime` attribute values from string to datetime

AleksanderWWW:

### DIFF
--- a/src/neptune/metadata_containers/utils.py
+++ b/src/neptune/metadata_containers/utils.py
@@ -155,6 +155,6 @@ def _parse_entry(entry: LeaderboardEntry) -> LeaderboardEntry:
                 for attribute in entry.attributes
             ],
         )
-    except TypeError:
-        # date is already in the right format
+    except (TypeError, ValueError):
+        # date is already in the right format or the parsing format is incorrect
         return entry
--- a/src/neptune/metadata_containers/utils.py
+++ b/src/neptune/metadata_containers/utils.py
@@ -165,7 +165,7 @@ def _parse_entry(entry: LeaderboardEntry) -> LeaderboardEntry:
     except ValueError:
         # the parsing format is incorrect
         warn_once(
-            "Date parsing failed. The date format is incorrect. Returning in the string format.",
+            "Date parsing failed. The date format is incorrect. Returning as string instead of datetime.",
             exception=NeptuneWarning,
         )
         return entry
--- a/src/neptune/metadata_containers/utils.py
+++ b/src/neptune/metadata_containers/utils.py
@@ -13,6 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+__all__ = [
+    "DATE_FORMAT",
+    "parse_dates",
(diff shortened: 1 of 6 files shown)


AleksanderWWW:
## E2E :white_check_mark: 
https://github.com/neptune-ai/neptune-client/actions/runs/7801924014

src/neptune/metadata_containers/utils.py
```diff
@@ -29,6 +43,8 @@
 )
 from neptune.internal.utils.run_state import RunState
 
+DATE_FORMAT: str = "%Y-%m-%dT%H:%M:%S.%fZ"
```
AleksanderWWW, line 46: I don't really like it tbh, but not sure what to do. Do we have a way to predict the output format of date objects from backend?

szysad: There is lib [dateutils.parser](https://dateutil.readthedocs.io/en/stable/parser.html) but it's external library and much slower so not worth the hassle imo.

I think its good idea to wrap datetime parsing in try catch and logging warning / error if format doesnt match datetime (ValueError is risen)

szysad (approved):

src/neptune/metadata_containers/utils.py
AleksanderWWW: Yes, that could work (I mean the try-catch). Was also thinking if this format is somehow set in the backend and perhaps it would be possible to expose it to the client with an endpoint for that purpose

src/neptune/metadata_containers/utils.py
```diff
@@ -115,3 +131,30 @@ def prepare_nql_query(
...
+                    },
+                )
+                if attribute.type == AttributeType.DATETIME
+                else attribute
+                for attribute in entry.attributes
+            ],
+        )
+    except (TypeError, ValueError):
+        # date is already in the right format or the parsing format is incorrect
+        return entry
```
szysad, line 160: Looks good, I'd only add logging that date parsing fails instead of only defaulting to str value

szysad (approved):

szysad (dismissed):
LGTM!

tests/unit/neptune/new/client/test_project.py
```diff
@@ -131,3 +138,49 @@ def test_prepare_nql_query():
...
+                    {"value": "07-02-2024"},  # different format than expected
+                )
+            ],
+        )
+    ]
+
+    parsed = list(parse_dates(entries))
+    assert parsed[0].attributes[0].properties["value"] == "07-02-2024"  # should be left unchanged due to ValueError
+    mock_warn_once.assert_called_once_with(
+        "Date parsing failed. The date format is incorrect. Returning in the string format.",
```
normandy7, line 184: Maybe something slightly more helpful. Like this, if I understand it correctly

```suggestion
        "Date parsing failed. The date format is incorrect. Returning as string instead of datetime.",
```

AleksanderWWW: Done

normandy7 (approved):

PR #1638, opened 2024-02-08
Disable progress bar for small fetchings

AleksanderWWW:

### DIFF
--- a/src/neptune/api/searching_entries.py
+++ b/src/neptune/api/searching_entries.py
@@ -160,7 +160,7 @@ def iter_over_pages(
         **kwargs,
     ).get("matchingItemCount", 0)
 
-    progress_bar = False if total < step_size else progress_bar  # disable progress bar if only one page is fetched
+    progress_bar = False if total <= step_size else progress_bar  # disable progress bar if only one page is fetched
 
     with construct_progress_bar(progress_bar, "Fetching table...") as bar:
         # beginning of the first page
--- a/src/neptune/api/searching_entries.py
+++ b/src/neptune/api/searching_entries.py
@@ -154,19 +154,19 @@ def iter_over_pages(
     searching_after = None
     last_page = None
 
-    total = 0
+    total = get_single_page(
+        limit=0,
+        offset=0,
+        **kwargs,
+    ).get("matchingItemCount", 0)
 
-    progress_bar = progress_bar if step_size >= total else None
+    progress_bar = False if total < step_size else progress_bar  # disable progress bar if only one page is fetched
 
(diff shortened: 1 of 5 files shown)


src/neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -1094,7 +1094,7 @@ def search_leaderboard_entries(
                 sort_by=sort_by,
                 ascending=ascending,
                 sort_by_column_type=sort_by_column_type,
-                progress_bar=progress_bar,
+                progress_bar=False if limit < default_step_size else progress_bar,
```
AleksanderWWW, line 1097: If `limit` is smaller than the `default_step_size` then it means that we will not fetch even a "full" single page, so no need to construct a progress bar for that

Raalsky: But as we're always fetching the total number of matching entries -> is it needed?

src/neptune/api/searching_entries.py
```diff
@@ -154,19 +154,19 @@ def iter_over_pages(
...
 
-    total = 0
+    total = get_single_page(
+        limit=0,
+        offset=0,
+        **kwargs,
+    ).get("matchingItemCount", 0)
 
-    progress_bar = progress_bar if step_size >= total else None
+    progress_bar = False if total < step_size else progress_bar  # disable progress bar if only one page is fetched
```
Raalsky, line 163: I'm not 100% sure but shouldn't it be `<=`?

src/neptune/internal/backends/hosted_file_operations.py
```diff
@@ -481,6 +481,9 @@ def _store_response_as_file(
         target_file = destination
 
     total_size = int(response.headers.get("content-length", 0))
+
+    progress_bar = False if total_size < 1024 * 1024 else progress_bar  # less than one chunk
```
Raalsky, line 485: 1. Magic numbers 🛑 
2. Looks like`content-length` may not be provided. If I understand correctly this doesn't mean that the file is empty; rather it's not possible to determine the target size. In such a case we should probably show the progress bar.

src/neptune/api/searching_entries.py
AleksanderWWW: Done

src/neptune/internal/backends/hosted_file_operations.py
AleksanderWWW: Done

Raalsky (approved):

PR #1644, opened 2024-02-09
Better handle limit in fetch_*_table

AleksanderWWW:

~~- [ ] Did you **ask the docs owner** to review all the user-facing changes?~~

### DIFF
--- a/src/neptune/api/searching_entries.py
+++ b/src/neptune/api/searching_entries.py
@@ -144,6 +144,7 @@ def find_attribute(*, entry: LeaderboardEntry, path: str) -> Optional[AttributeW
 def iter_over_pages(
     *,
     step_size: int,
+    limit: Optional[int] = None,
     sort_by: str = "sys/id",
     max_offset: int = MAX_SERVER_OFFSET,
     sort_by_column_type: Optional[str] = None,
@@ -160,7 +161,12 @@ def iter_over_pages(
         **kwargs,
     ).get("matchingItemCount", 0)
 
+    if limit is not None and total > limit:
+        total = limit
+
     progress_bar = False if total <= step_size else progress_bar  # disable progress bar if only one page is fetched
+
+    extracted_records = 0
 
     with construct_progress_bar(progress_bar, "Fetching table...") as bar:
         # beginning of the first page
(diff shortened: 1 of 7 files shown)


tests/unit/neptune/new/client/abstract_tables_test.py
```diff
@@ -217,22 +217,7 @@ def test_get_table_as_table_entries(
...
-        now = datetime.now()
-        attributes = self.build_attributes_leaderboard(now)
-
-        # and
-        empty_entry = LeaderboardEntry(str(uuid.uuid4()), [])
-        filled_entry = LeaderboardEntry(str(uuid.uuid4()), attributes)
-        search_leaderboard_entries.return_value = [empty_entry, filled_entry]
-
-        # and
-        for limit, expected_len in [(1, 1), (2, 2), (3, 2), (10_000, 2)]:
```
AleksanderWWW, line 232: Those tests no longer make sense if we don't perform limiting in the `HostedNeptuneBackend` but in the `searching_entries` module

src/neptune/api/searching_entries.py
```diff
@@ -55,6 +55,26 @@
...
+        return True
+
+    def __le__(self, other: Any) -> bool:
+        return False
+
+    def __eq__(self, other: Any) -> bool:
+        return False
+
+    def __ne__(self, other: Any) -> bool:
+        return True
```
szysad, line 75: fyi, `float('inf')` has the same interface, but is of 'float' type so might not be applicable in out case :(

szysad (approved):
Looks good!

One quick question tho.
Was there any need for this change except from getting rid of constant checking if 'limit' was none in body of this `iter_over_pages` function?

src/neptune/api/searching_entries.py
AleksanderWWW: I was really conflicted between this and `float("inf")` but mypy would complain because `limit` is `int | None`.
We can disable this particular check with `type: ignore` comment, so what I did here is a proposition that I'm open to discuss :smile:

AleksanderWWW:
> Was there any need for this change except from getting rid of constant checking if 'limit' was none in body of this iter_over_pages function?

Yes, it turned out that not checking it in `iter_over_pages` resulted in progress bars appearing even if there was only 1 out of however many records to fetch, which wasn't very pretty tbh

`Fetching table...:   0%|          | 1/402 [00:00<00:27, 14.62/s]`

src/neptune/api/searching_entries.py
szysad: I think you made the right call 😎.
We could try to simplify implementation with [functools.total_ordering](https://docs.python.org/3/library/functools.html#functools.total_ordering) but it doesn't override with superclass method overriding and we want that `int` superclass 😤

Issue #1416, opened 2023-08-07
BUG: Inconsistent behaviour betweeen `init_model` and `init_model_version`


d13g0:
### Describe the bug
I create a model with:
```
mod = neptune.init_model(
            key='MODEL',
            project = 'PROJECT',
            name = 'My Model'
        )
```

Then, I want to create a version of this model with:
```
mod_version = neptune.init_model_version(
            model='MODEL',
            project='PROJECT'

        )
```

I get this error:
```
311, in get_metadata_container
    raise MetadataContainerNotFound.of_container_type(
neptune.exceptions.ModelNotFound: Model <<user>>/PROJECT/MODEL not found.
```

However, if I add the "key" to the model name like this (e.g. 'PRE'):

```
mod_version = neptune.init_model_version(
            model='PRE-MODEL',
            project='PROJECT'
        )
```

Then the model version is created. 

### Expected behavior
Since I am providing the `project` parameter to create the model version, the code should not fail, as the proper prefix (key) can be inferred from the project name.

### Traceback
If applicable, add traceback or log output/screenshots to help explain your problem.
### Environment
**The output of `pip list`:**
```
neptune                   1.4.1
neptune-notebooks         0.9.3
neptune-pytorch           1.0.1

```
**The operating system you're using:**
MacOS
**The output of `python --version`:**
3.10.12

### Additional context
None

Blaizzy:
Hi @d13g0 

This happens because the model version method takes a combination of `project_key` and `model_key` as the model argument.
Where `project_key` can be fetched from a neptune object ID (i.e., runs, project, etc.).

**Example**:
```python

project = "workspace/project_name"
run = neptune.init_run(project=project)

# (Neptune) Initializing a Model and Model version
model_key = "PIPELINES"
project_key = run["sys/id"].fetch().split("-")[0]

# (Neptune) Creating a new model
model = neptune.init_model(key=model_key, project=project)
model.wait()

# (Neptune) Creating a new model version
model_version = neptune.init_model_version(model=f"{project_key}-{model_key}")
```

d13g0:
@Blaizzy I think what you are saying is sensible. However, I am giving you the perspective of a noob (me) trying the framework for the first time. The simpler the better for adoption, IMHO. Thanks!

SiddhantSadangi:
Hey @d13g0 ,

Your suggestion actually makes total sense!
We have discussed this internally in the past as well and will be processing your suggestions as a feature request :)

PR #1646, opened 2024-02-09
Infer full model name from project

AleksanderWWW:

Solves #1416 
@SiddhantSadangi

### DIFF
--- a/src/neptune/metadata_containers/model_version.py
+++ b/src/neptune/metadata_containers/model_version.py
@@ -222,8 +222,11 @@ class ModelVersion(MetadataContainer):
             if self._mode == Mode.READ_ONLY:
                 raise NeedExistingModelVersionForReadOnlyMode()
 
+            model_id = self._project_api_object.sys_id
+            full_model_name = self._model if self._model.split("-")[0] == model_id else f"{model_id}-{self._model}"
+
             api_model = self._backend.get_metadata_container(
-                container_id=QualifiedName(project_qualified_name + "/" + self._model),
+                container_id=QualifiedName(project_qualified_name + "/" + full_model_name),
                 expected_container_type=ContainerType.MODEL,
             )
             return self._backend.create_model_version(project_id=self._project_api_object.id, model_id=api_model.id)


SiddhantSadangi:
Just a heads up to inform relevant stakeholders as this would need all our public code snippets and examples to be updated

szysad (approved):
LGTM!

src/neptune/metadata_containers/model_version.py
```diff
@@ -222,8 +222,11 @@ def _get_or_create_api_object(self) -> ApiExperiment:
             if self._mode == Mode.READ_ONLY:
                 raise NeedExistingModelVersionForReadOnlyMode()
 
+            model_id = self._project_api_object.sys_id
```
Raalsky, line 225: Isn't it `project_key`?

src/neptune/metadata_containers/model_version.py
```diff
@@ -222,8 +222,11 @@ def _get_or_create_api_object(self) -> ApiExperiment:
...
+            full_model_name = self._model if self._model.split("-")[0] == model_id else f"{model_id}-{self._model}"
```
Raalsky, line 226: > `self._model.split("-")[0] == model_id`

How about just using `startswith(...)`?

normandy7:
> Just a heads up to inform relevant stakeholders as this would need all our public code snippets and examples to be updated

We _could_ update all our examples, but I'm not convinced we should. This change doesn't really make our somewhat confusing identifier scheme any easier to explain; it'll just remove an obstacle when the user doesn't consider that the full ID also contains the project key.

This "magic" may be helpful in most cases, but we run into a problem again if the user wants to initialize an existing model version. Maybe now they'd expect to use `init_model_version(with_id="MOD-1")`, but it won't work.

Anyway, these thoughts are not blocking. Just wanted to share that I think our ID/key scheme remains tricky to understand and use, so it could be something to rethink.

SiddhantSadangi:
> Maybe now they'd expect to use init_model_version(with_id="MOD-1"), but it won't work.

We should be updating both `init_model()` and `init_model_version()` to prepend the project key to the model/model version passed with `with_id`, if none is prepended already

normandy7:
> We should be updating both `init_model()` and `init_model_version()` to prepend the project key to the model/model version passed with `with_id`, if not prepended already by the user

[Not sure I agree](https://neptune-labs.slack.com/archives/C03HL9NKEFQ/p1707746013053339).

src/neptune/metadata_containers/model_version.py
AleksanderWWW: What if someone has a model named `ABC-ABCModel`?
Then the bare model name starts with `ABC`, but doesn't contain the ID that backend expects

Issue #1622, opened 2024-01-26
NPT-14389: `.neptune` folder is not cleaned up if multiple PyTorch Dataloaders are used


NiklasKappel:
### Describe the bug
When neptune is used to track a pytorch (lightning) project, and the project uses pytorch `DataLoader`s which specify their optional `num_workers` argument, neptune creates many persistent files in the `.neptune` directory. After the run is completed, the `.neptune` directory looks something like this:
```
tree .neptune/
.neptune/
├── async
│   └── run__9b2edd24-6733-4f30-af3a-915c49553c8c
│       ├── exec-1706255810.093006-2024-01-26_08.56.50.093006-6341
│       │   ├── data-1.log
│       │   ├── last_ack_version
│       │   ├── last_put_version
│       │   ├── metadata.json
│       │   └── upload_path
│       ├── exec-1706255810.227333-2024-01-26_08.56.50.227333-6350
│       │   ├── data-1.log
│       │   ├── last_ack_version
│       │   ├── last_put_version
│       │   ├── metadata.json
│       │   └── upload_path
│       ├── exec-1706255810.376134-2024-01-26_08.56.50.376134-6359
│       │   ├── data-1.log
│       │   ├── last_ack_version
│       │   ├── last_put_version
│       │   ├── metadata.json
│       │   └── upload_path
│       ├── exec-1706255810.500691-2024-01-26_08.56.50.500691-6368
│       │   ├── data-1.log
│       │   ├── last_ack_version
│       │   ├── last_put_version
│       │   ├── metadata.json
│       │   └── upload_path
│       ├── exec-1706255810.641591-2024-01-26_08.56.50.641591-6378
│       │   ├── data-1.log
│       │   ├── last_ack_version
│       │   ├── last_put_version
│       │   ├── metadata.json
│       │   └── upload_path
│       ├── exec-1706255810.761324-2024-01-26_08.56.50.761324-6387
│       │   ├── data-1.log
│       │   ├── last_ack_version
│       │   ├── last_put_version
│       │   ├── metadata.json
│       │   └── upload_path
│       └── exec-1706255810.893379-2024-01-26_08.56.50.893379-6396
│           ├── data-1.log
│           ├── last_ack_version
│           ├── last_put_version
│           ├── metadata.json
│           └── upload_path
└── Untitled
    └── TES-4
        └── checkpoints
            └── epoch=0-step=100.ckpt

20 directories, 29 files
```
[...]

PR #1628, opened 2024-02-02
[DO-NOT-MERGE] Proof of concept for lazy evaluated OperationProcessor when forking

Raalsky:
## Before submitting checklist

- [ ] Did you **update the CHANGELOG**? (not for test updates, internal changes/refactors or CI/CD setup)
- [ ] Did you **ask the docs owner** to review all the user-facing changes?

PR #1649, opened 2024-02-13
Lazy initialization of operation processor

szysad:
Lazy initialization of operation processor. Fixes [NPT-14389](https://github.com/neptune-ai/neptune-client/issues/1622)

- [ ] <s> Did you **ask the docs owner** to review all the user-facing changes?</s>

### DIFF
--- a/src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
+++ b/src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 __all__ = ("LazyOperationProcessorWrapper",)
 
+from pathlib import Path
 from typing import (
     Any,
     Callable,
@@ -25,6 +26,7 @@ from typing import (
     TypeVar,
 )
 
+from neptune.core.components.abstract import Resource
 from neptune.core.components.operation_storage import OperationStorage
 from neptune.internal.operation import Operation
 from neptune.internal.operation_processors.operation_processor import OperationProcessor
@@ -32,16 +34,18 @@ from neptune.internal.operation_processors.operation_processor import OperationP
 RT = TypeVar("RT")
 
 
-def trigger_op_processor(method: Callable[..., RT]) -> Callable[..., RT]:
+def trigger_evaluation(method: Callable[..., RT]) -> Callable[..., RT]:
     def _wrapper(self: LazyOperationProcessorWrapper, *args: Any, **kwargs: Any) -> RT:
         if self._operation_processor is None:
             self._operation_processor = self._operation_processor_getter()
+            if self._post_trigger_side_effect is not None:
+                self._post_trigger_side_effect()
         return method(self, *args, **kwargs)
 
     return _wrapper
 
 
-def exec_if_triggered(method: Callable[..., RT]) -> Callable[..., RT]:
+def noop_if_not_triggered(method: Callable[..., RT]) -> Callable[..., RT]:
     def _wrapper(self: LazyOperationProcessorWrapper, *args: Any, **kwargs: Any) -> RT:
         if self._operation_processor is not None:
             return method(self, *args, **kwargs)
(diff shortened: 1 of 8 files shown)


szysad:
✅ [E2E](https://github.com/neptune-ai/neptune-client/actions/runs/7885250713)

src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(
+        self,
+        operation_processor_getter: Callable[[], OperationProcessor],
+        post_trigger_side_effect: Optional[Callable[[], Any]] = None,
+    ):
+        self._operation_processor_getter = operation_processor_getter
+        self._post_trigger_side_effect = post_trigger_side_effect
+        self._operation_processor: OperationProcessor = None  # type: ignore
+
+    def evaluated(self) -> bool:
```
Raalsky, line 66: Take a look at the usage of `evaluated` in my PR: https://github.com/neptune-ai/neptune-client/pull/1628/files#diff-aee064cd8dc933c9e3a255fd43e594782bf4146bb594326cbb8e3c66a7debc62R500 . If I understand correctly, here we will always have a `stop`-related messages that are probably not needed. Especially, we rather don't want to log anything to stdout if OperationProcessor wasn't initialized.

Raalsky: + How about a `@property`?

src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
```diff
@@ -0,0 +1,112 @@
...
+
+class LazyOperationProcessorWrapper(OperationProcessor):
+    def __init__(
+        self,
+        operation_processor_getter: Callable[[], OperationProcessor],
+        post_trigger_side_effect: Optional[Callable[[], Any]] = None,
+    ):
+        self._operation_processor_getter = operation_processor_getter
+        self._post_trigger_side_effect = post_trigger_side_effect
+        self._operation_processor: OperationProcessor = None  # type: ignore
```
Raalsky, line 64: Shouldn't it be `Optional[OperationProcessor]`?

Raalsky: Or was it to simplify the implementation?

src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
szysad: I will investigate excessive stop messages as I was not aware. Maybe there is another way to get rid of them.

I don't like using `evaluated` in `metadata_container` as its outside of OperationProcessor interface and it introduces type errors if `self._op_processor` is not defined as `LazyOperationProcessorWrapper`. 

Thats another reason why I made decision discussed [here](https://github.com/neptune-ai/neptune-client/pull/1649#discussion_r1490636799)

src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
szysad: It should be `Optional[OperationProcessor]` but then I can not call `self._operation_processor.some_method()` as from typchecker perspective `self._operation_processor` might be `None`. We know that wont be the case becouse of all the decorators wrapping all of methods, but I am not aware of any clean way to implement this kind of type guard, so I just went with  typecheck supress in one line

src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
Raalsky: Yeah, for sure we should do something more feasible with `evaluated` than I did in Proof-of-Concept :wink:. Just be aware that we don't want to have any messages if the operation processor wasn't created (this brings us mostly to stop messages).

src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
Raalsky: Sure, I agree. Just wanted to know the rationale :wink:. Thanks!

src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
szysad: UPDATE: indeed fork spawned childs are cousin additional logs.

If the intended behaviour is not logging shutdown messages if `LazyOperationProcessor` was not triggered, maybe we can add `shutdown hook` that will log this messages before or after `operationProcessor.stop()` function.

We could also for each `metadata_container` process keep var to know if we are parent (responsible for logging shutdown messages) or child which should be silent.

Let me know what you think about that :D

src/neptune/metadata_containers/metadata_container.py
```diff
@@ -175,14 +178,17 @@ def __init__(
...
 
-        self._op_processor: OperationProcessor = get_operation_processor(
-            mode=mode,
-            container_id=self._id,
-            container_type=self.container_type,
-            backend=self._backend,
-            lock=self._lock,
-            flush_period=flush_period,
-            queue=self._signals_queue,
+        self._op_processor: LazyOperationProcessorWrapper = LazyOperationProcessorWrapper(
```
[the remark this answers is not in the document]
Raalsky: > I guess we could define self._op_processor of type OperationProcessor and in __init__ set it as non wrapped processor and in child fork set it as LazyWrapper so we just switch implementation of interface. What do you think about that?

Yes, as Lazy should be just a special implementation of `OperationProcessor` interface/abstract class this won't hurt as as long as the typing was set to this interface.

> there is no real difference (init will be triggered by self._write_initial_attributes() few lines later ) if I wrap op_processor in LazyWrapper here.

There is one difference called ... debugging 🤖 , or rather the complexity of it. In case of having to follow some quite low-level things with this approach, you will have to go a way deeper in a calls stack etc.

src/neptune/internal/operation_processors/lazy_operation_processor_wrapper.py
Raalsky: > If the intended behaviour is not logging shutdown messages if LazyOperationProcessor was not triggered, maybe we can add shutdown hook that will log this messages before or after operationProcessor.stop() function.

So you may assume that there is no need to print anything except the logs from `OperationProcessor.stop()` itself from a forked Run. So:
- No need to print info about background jobs shuting down
- No need to print info about URL to metadata

Does it help you?

Raalsky: This means that you don't care about evaluated/non-evaluated as it's a part of OperationProcessor implementation and rather just store info about being forked/or not that could be a part of `MetadataContainer`.

szysad: yep, ty :)

Raalsky (approved):

PR #1664, opened 2024-02-27
scale images to [0,1] range

AleksanderWWW:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## [UNRELEASED] neptune 1.10.0
 
 ### Features
-- Added auto-scaling for image logging ([#1664](https://github.com/neptune-ai/neptune-client/pull/1664))
+- Added auto-scaling pixel values for image logging ([#1664](https://github.com/neptune-ai/neptune-client/pull/1664))
 
 ### Fixes
 - Restored support for SSL verification exception ([#1661](https://github.com/neptune-ai/neptune-client/pull/1661))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
-## [UNRELEASED] neptune 1.9.2
+## [UNRELEASED] neptune 1.10.0
+
+### Features
+- Added auto-scaling for image logging ([#1664](https://github.com/neptune-ai/neptune-client/pull/1664))
 
 ### Fixes
 - Restored support for SSL verification exception ([#1661](https://github.com/neptune-ai/neptune-client/pull/1661))
--- a/src/neptune/internal/utils/images.py
+++ b/src/neptune/internal/utils/images.py
@@ -48,7 +48,7 @@ from neptune.internal.utils.logger import get_logger
 
 logger = get_logger()
 SEABORN_GRID_CLASSES = {"FacetGrid", "PairGrid", "JointGrid"}
-ALLOWED_IMG_PIXEL_RANGES = ("[0, 255]", "[0, 1]", "[-1, 1]")
+ALLOWED_IMG_PIXEL_RANGES = ("[0, 255]", "[0.0, 1.0]")
 
 try:
     from numpy import array as numpy_array
@@ -229,9 +229,6 @@ def _scale_array(array: np.ndarray) -> np.ndarray:
     if array_min >= 0 and array_max <= 1:
         return array * 255
 
-    if array_min >= -1 and array_max <= 1:
-        return (array + 1) / 2 * 255
-
     _warn_about_incorrect_image_data_range(array_min, array_max)
     return array
 
@@ -239,7 +236,7 @@ def _scale_array(array: np.ndarray) -> np.ndarray:
 def _warn_about_incorrect_image_data_range(array_min: int | float, array_max: int | float) -> None:
     msg = f"Image data is in range [{array_min}, {array_max}]."
     logger.warning(
-        "%s To be interpreted as colors correctly values in the array need to be in the %s, %s or %s range.",
+        "%s To be interpreted as colors correctly values in the array need to be in the %s or %s range.",
         msg,
         *ALLOWED_IMG_PIXEL_RANGES,
     )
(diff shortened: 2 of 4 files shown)


szysad (approved):
LGTM!

SiddhantSadangi:
@normandy7 - this would need a docs change (all instances where an image array is divided by 255 for logging to Neptune)

Raalsky:
> @normandy7 - this would need a docs change (all instances where an image array is divided by 255 for logging to Neptune)

But, isn't it optional? If I understand, this not going to have an impact on cases that the user provied already correctly scaled image and docs could be just updated anytime

SiddhantSadangi:
> But, isn't it optional? If I understand, this not going to have an impact on cases that the user provied already correctly scaled image and docs could be just updated anytime

@Raalsky - yeah, updating docs is not a blocker for the client release.

src/neptune/internal/utils/images.py
```diff
@@ -228,6 +220,39 @@ def _get_numpy_as_image(array):
...
+    max_value = np.max(array)
+
+    if min_value >= 0 and max_value <= 1:
+        return array
+
+    # Avoid division by zero if all values are the same
+    if min_value == max_value:
+        scaled_array = np.zeros_like(array)
+    else:
+        scaled_array = (array - min_value) / (max_value - min_value)
```
Raalsky, line 234: Are you sure about this approach? This will change the brightness of the image if I understand. This may work for ML-purposes to normalize images as an input to some model but I would rather just divide by `255` to make the image consistent.

szysad: or just divide by `maxabs` value. Dividing by 255 makes sense if array is of type `uint8`

src/neptune/internal/utils/images.py
```diff
@@ -196,36 +197,57 @@ def _image_content_to_html(content: bytes) -> str:
...
+    if min_value >= 0 and 1 < max_value <= 255:
+        return array
 
+    if min_value >= 0 and max_value <= 1:
+        return array * 255
+
+    if min_value >= -1 and max_value <= 1:
+        return (array + 1) / 2 * 255
+
+    return (array + 1) * 128
```
Raalsky, line 234: It's duplicated with the [-1, 1] case.

src/neptune/internal/utils/images.py
```diff
@@ -196,36 +197,57 @@ def _image_content_to_html(content: bytes) -> str:
     return "<img src='data:image/png;base64," + str_equivalent_image + "'/>"
 
 
-def _get_numpy_as_image(array):
+def _get_numpy_as_image(array: np.ndarray, autoscale: bool) -> bytes:
```
Raalsky, line 200: I'll get back to you as we probably can get rid of the `autoscale` param

SiddhantSadangi: @AleksanderWWW - [0,255]

src/neptune/internal/utils/images.py
```diff
@@ -196,38 +200,51 @@ def _image_content_to_html(content: bytes) -> str:
...
+    array_max = array.max()
+
+    if array_min >= 0 and 1 < array_max <= 255:
+        return array
+
+    if array_min >= 0 and array_max <= 1:
+        return array * 255
+
+    if array_min >= -1 and array_max <= 1:
+        return (array + 1) / 2 * 255
```
szysad, line 233: This is a very unusual image representation. Maybe it is better not to handle this case and just warn about incorrect data range

AleksanderWWW: Yeah, I think that we actually need to have a discussion about what we want to support and how to proceed with exotic image representations and warnings :face_exhaling:

SiddhantSadangi: These are normalized images in the range [-1,1].
This range is generally used as an input for models, not for rendering images.

Most users would be normalizing images, so IMO, we should take care of rendering it from our end

szysad: @SiddhantSadangi 
Most of the input transformation in deep learning is obtained by [normalization](https://pytorch.org/vision/main/generated/torchvision.transforms.Normalize.html) which aims to make image channels to have mean 0 and std 1.

Usually process looks like this: scale to range [0, 1] -> apply normalization (might be out of [0, 1 range]) -> feed into network 

I would be very coucious when it comes to scaling images for user behind the user. It might be the case that normalized images fit into range [0, 1] and naive linear scaling will disturb the image (we need to know mean and std of dataset to perform inverse transformation) confusing end user.

I suggest picking the most common representations, ex. recognized by other libraries like matplotlib. Documenting them properly  and warning user when other conventions are detected

CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
-## [UNRELEASED] neptune 1.9.2
+## [UNRELEASED] neptune 1.10.0
+
+### Features
+- Added auto-scaling for image logging ([#1664](https://github.com/neptune-ai/neptune-client/pull/1664))
```
Raalsky, line 4: We should probably mention that it's pixel range auto-scaling as scaling wrt. images may mean changing size.

src/neptune/internal/utils/images.py
```diff
@@ -45,6 +48,7 @@
 
 logger = get_logger()
 SEABORN_GRID_CLASSES = {"FacetGrid", "PairGrid", "JointGrid"}
+ALLOWED_IMG_PIXEL_RANGES = ("[0, 255]", "[0, 1]", "[-1, 1]")
```
Raalsky, line 51: -> `[0.0, 1.0]`?

src/neptune/internal/utils/images.py
Raalsky: Closing as we should proceed with the `autoscale` param just in case. @szysad had some examples that could break our autodetection that were consistent with my initial thoughts.

src/neptune/internal/utils/images.py
Raalsky: I agree with @szysad and I want to be consistent with what matplotlib does. They're supporting `[0-255]` and `[0.0-1.0]` and there is a warning and clipping to `[0, 1]` in other cases like `[-1, 1]`

Raalsky (approved):

PR #1667, opened 2024-02-29
adjust as_image docstring

normandy7:


src/neptune/types/atoms/file.py
```diff
@@ -174,15 +174,16 @@ def from_stream(stream: IOBase, *, seek: Optional[int] = 0, extension: Optional[
     def as_image(image, autoscale: bool = True) -> "File":
         """Static method for converting image objects or image-like objects to an image File value object.
 
-        This way you can upload `Matplotlib` figures, `Seaborn` figures, `PIL` images, `NumPy` arrays, as static images.
+        This way you can upload Matplotlib figures, Seaborn figures, PIL images, and NumPy arrays as static images.
 
         Args:
             image: Image-like object to be converted.
```
Raalsky, line 180: We should mention that the input image pixel should be either in `[0.0, 1.0]`(floats)  or `[0-255]` (integer). In case they have something weird they need to do the normalization on their end.

src/neptune/types/atoms/file.py
```diff
@@ -174,15 +174,17 @@ def from_stream(stream: IOBase, *, seek: Optional[int] = 0, extension: Optional[
...
-                Supported are `PyTorch` tensors, `TensorFlow/Keras` tensors, `NumPy` arrays, `PIL` images,
-                `Matplotlib` figures and `Seaborn` figures.
+                The input image pixel must be either in range [0.0, 1.0] (float) or [0, 255] (integer).
+                Supported are PyTorch tensors, TensorFlow/Keras tensors, NumPy arrays, PIL images,
+                Matplotlib figures and Seaborn figures.
+            autoscale: Whether to scale images to the [0, 1] range.
```
Raalsky, line 184:
```suggestion
            autoscale: If `True` (default) whether to try to detect pixel ranges automatically and scale it to the format acceptable by Neptune.
```

normandy7: Alright, sounds more specific. If I understand the grammar correctly, how about:

```suggestion
            autoscale: Whether Neptune should try to detect the pixel range automatically and scale it to an acceptable format.
```

Raalsky: Perfect! Maybe we should mention that it's turned on by default?

Raalsky (approved):

src/neptune/types/atoms/file.py
normandy7: In general, I've moved in the direction of not repeating types or defaults if they're evident from the function signature (more in Slack). But we can do it here if you think it's beneficial.

SiddhantSadangi: @Raalsky / @normandy7  - late to the discussion here, but IMO the current docstring does not inform the user about what the "acceptable format" is, and does not warn them that this could distort the images if they lie outside the "acceptable" range.

Since the point of `autoscale` is to better render the images on the Neptune web app, how about the below?

```
Whether Neptune should try to scale image pixel values to better render them on the app. 
This can distort images if their pixels lie outside the [0.0, 1.0] or [0, 255] range. 
To disable this auto-scaling, set the argument to `False`.
```

Raalsky: I'm leaving it up to @normandy7 :wink: Both works for me (current and proposed).

normandy7: > `This can distort images`

Replace `This` with what it actually refers to. `Scaling`?

PR #1670, opened 2024-03-02
Runtime dependencies check

AleksanderWWW:


src/neptune/exceptions.py
```diff
@@ -790,32 +791,39 @@ def __init__(self, project: QualifiedName):
...
-                integration_package_name=integration_package_name,
+                exception=self.__class__.__name__,
+                package_name=package_name,
                 framework_name=framework_name,
                 **STYLES,
             )
         )
 
 
+class NeptuneIntegrationNotInstalledException(NeptuneMissingRequirementException):
```
AleksanderWWW, line 822: @Raalsky tbh I'm not sure I see a need for a separate exception type for integrations. It makes it all more complicated. How about we just throw a `NeptuneMissingRequirementException` and always link a docs page for integrations? Also @normandy7, what do you think?

normandy7: You mean removing the installation instructions from the exception message? What are the pros/cons with this change (from an end-user perspective)? It would be ideal to have an installation command handy for copy-pasting

AleksanderWWW: No no, actually it will be just about renaming `NeptuneIntegrationNotInstalledException` to `NeptuneMissingRequirementException` and small tweaks in the message (changing 'integration' to just 'package') to make it more generic, as we'll also probably be checking installation of things like `pandas` which is not an integration per se

AleksanderWWW: The rest will stay the same - installation instruction and link to the docs. I just don't see a need to create two separate entities which only differ in names and some small message wording details

normandy7: Oh I see! Yes, that should be fine.

AleksanderWWW: I will introduce a simplification shortly and ask you to review, ok? :smiley:

normandy7: Certainly 👍🏻 I'll monitor my inbox

AleksanderWWW: There's a lot of files, but it's mostly just updating how integrations are imported. The thing to check is probably only `src/neptune/exceptions.py` :)

src/neptune/exceptions.py
```diff
@@ -790,26 +790,27 @@ def __init__(self, project: QualifiedName):
...
+    {h1}
+    ----NeptuneMissingRequirementException-----------------------------------------
+    {end}
+    Looks like package {package_name} wasn't installed.
+    To install, run:
+        {bash}pip install {package_name}{end}
+    Or:
+        {bash}pip install "neptune[{framework_name}]"{end}
+
+    You may also want to check the following docs page:
```
normandy7, line 805: Maybe this makes things slightly more clear and specific?

```suggestion
    Looks like the {package_name} package isn't installed.
    To install it, run:
        {bash}pip install {package_name}{end}
    Or install both Neptune and the integration:
        {bash}pip install "neptune[{framework_name}]"{end}

    For detailed instructions, check the integration guides:
```

szysad (dismissed):
clean, short and simple, LGTM! 🚀

src/neptune/internal/utils/requirement_check.py
```diff
@@ -0,0 +1,35 @@
...
+    "is_installed",
+]
+
+from importlib.util import find_spec
+from typing import Optional
+
+from neptune.exceptions import NeptuneMissingRequirementException
+
+
+def is_installed(requirement_name: str) -> bool:
```
Raalsky, line 27: Are we sure that this is going to be cached? Maybe can we just use some simple LRU cache directly?

PR #1674, opened 2024-03-04
Add support for fetching non-last items from a series

szysad:

### DIFF
--- a/src/neptune/attributes/series/fetchable_series.py
+++ b/src/neptune/attributes/series/fetchable_series.py
@@ -38,14 +38,6 @@ Row = TypeVar("Row", StringSeriesValues, FloatSeriesValues)
 MAX_FETCH_LIMIT = 1000
 
 
-def min_skip_none(a, b):
-    if a is None:
-        return b
-    if b is None:
-        return a
-    return min(a, b)
-
-
 class FetchableSeries(Generic[Row]):
     @abc.abstractmethod
     def _fetch_values_from_backend(self, offset, limit) -> Row:
@@ -80,27 +72,29 @@ class FetchableSeries(Generic[Row]):
         val = self._fetch_values_from_backend(offset=offset, limit=fetch_chunk_size)
 
         if limit is None:
-            limit = val.totalItemCount
+            to_load = val.totalItemCount - offset
+        else:
+            to_load = min(val.totalItemCount - offset, limit)
 
         data = val.values
         offset += len(data)
-        # limit -= len(data)
 
         # dont display progress bar if all values are fetched in one go
-        if limit - len(data) == 0:
+        left_to_fetch = to_load - len(data)
+        if left_to_fetch == 0:
             progress_bar = False
 
         path = path_to_str(self._path) if hasattr(self, "_path") else ""
         with construct_progress_bar(progress_bar, f"Fetching {path} values") as bar:
-            bar.update(by=len(data), total=limit)  # first fetch before the loop
-            while limit - len(data) != 0:
-                fetch_chunk_size = min(limit, MAX_FETCH_LIMIT)
+            bar.update(by=len(data), total=to_load)  # first fetch before the loop
+            while left_to_fetch != 0:
+                fetch_chunk_size = min(left_to_fetch, MAX_FETCH_LIMIT)
                 batch = self._fetch_values_from_backend(offset, fetch_chunk_size)
 
                 data.extend(batch.values)
                 offset += len(batch.values)
-                # limit -= len(batch.values)
-                bar.update(by=len(batch.values), total=limit)
+                left_to_fetch -= len(batch.values)
+                bar.update(by=len(batch.values), total=to_load)
 
         rows = dict((n, make_row(entry)) for (n, entry) in enumerate(data))
 
--- a/src/neptune/attributes/series/fetchable_series.py
+++ b/src/neptune/attributes/series/fetchable_series.py
@@ -35,19 +35,31 @@ from neptune.typing import ProgressBarType
 
 Row = TypeVar("Row", StringSeriesValues, FloatSeriesValues)
 
+MAX_FETCH_LIMIT = 1000
+
+
+def min_skip_none(a, b):
+    if a is None:
+        return b
+    if b is None:
+        return a
+    return min(a, b)
+
 
 class FetchableSeries(Generic[Row]):
     @abc.abstractmethod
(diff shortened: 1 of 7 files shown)


szysad:
⌛ [E2E](https://github.com/neptune-ai/neptune-client/actions/runs/8152838154)

tests/unit/neptune/new/attributes/series/test_float_series.py
```diff
@@ -22,11 +22,27 @@
...
+from neptune.internal.backends.api_model import (
+    FloatPointValue,
+    FloatSeriesValues,
+)
 from tests.unit.neptune.new.attributes.test_attribute_base import TestAttributeBase
 
 
 @patch("time.time", new=TestAttributeBase._now)
 class TestFloatSeries(TestAttributeBase):
+    def _get_float_series_values_dummy_impl(self, series: FloatSeriesValues):
```
AleksanderWWW, line 34: This should probably be a function

src/neptune/attributes/series/fetchable_series.py
```diff
@@ -57,16 +61,40 @@ def make_row(entry: Row) -> Dict[str, Union[str, float, datetime]]:
...
             return row
 
-        progress_bar = False if len(data) < limit else progress_bar
+        if offset is None:
+            offset = 0
+
+        if limit is not None:
+            fetch_chunk_size = min(limit, MAX_FETCH_LIMIT)
+        else:
+            fetch_chunk_size = MAX_FETCH_LIMIT
```
AleksanderWWW, line 70: Perhaps we could make it into one-liners? Also, I remember creating this `NoLimit` thing that could simplify the code, so that we don't have to check for `limit` being `None` each time

src/neptune/attributes/series/fetchable_series.py
```diff
@@ -57,16 +61,40 @@ def make_row(entry: Row) -> Dict[str, Union[str, float, datetime]]:
...
+        else:
+            to_load = min(val.totalItemCount - offset, limit)
+
+        data = val.values
+        offset += len(data)
+
+        # dont display progress bar if all values are fetched in one go
+        left_to_fetch = to_load - len(data)
+        if left_to_fetch == 0:
+            progress_bar = False
```
AleksanderWWW, line 85: Also a one-liner potential

src/neptune/attributes/series/fetchable_series.py
```diff
@@ -35,20 +35,24 @@
...
     @abc.abstractmethod
     def _fetch_values_from_backend(self, offset, limit) -> Row:
         pass
 
-    def fetch_values(self, *, include_timestamp: bool = True, progress_bar: Optional[ProgressBarType] = None):
+    def fetch_values(
+        self,
+        *,
+        include_timestamp: bool = True,
+        progress_bar: Optional[ProgressBarType] = None,
```
AleksanderWWW, line 50: Let's maybe move the `progress_bar` to the bottom of the function signature, as it will likely be the least-often set argument

tests/unit/neptune/new/attributes/series/test_string_series.py
```diff
@@ -45,3 +64,23 @@ def test_log(self):
...
+        limit = 42
+        words = [fake.word() for _ in range(total)]
+        series = StringSeriesValues(
+            values=[StringPointValue(step=i, value=words[i], timestampMillis=i) for i in range(total)],
+            totalItemCount=total,
+        )
+        with self._exp() as exp:
+            with patch(
+                "neptune.internal.backends.neptune_backend_mock.NeptuneBackendMock.get_string_series_values",
+                new=self._get_string_series_values_dummy_impl(series),
```
AleksanderWWW, line 81: If this is turned to a function then we could patch it in the `@patch` decorator, right? Might be a little cleaner

AleksanderWWW:
Comments are mostly styling-related :wink: 
The logic seems valid. Overall looks very good! :rocket:

tests/e2e/standard/test_series.py
```diff
@@ -56,34 +66,73 @@ def test_log(self, container: MetadataContainer, series_type: str):
...
+            expected_values,
+            last_value=values[-1],
+            fetch_kwargs=fetch_kwargs,
+        ) as (namespace, _, _, _):
+            namespace.assign(series_init(values=values, steps=steps, timestamps=timestamps))
+
     @contextmanager
-    def run_then_assert(self, container: MetadataContainer, series_type: str):
+    def run_then_assert_default(
+        self, container: MetadataContainer, series_type: str, fetch_kwargs: Dict[str, Any] = {}
```
AleksanderWWW, line 134: Mutable defaults are not a good idea. Perhaps we could pass `None` as default and then set to `{}` inside the function body?

szysad: good point!
Forgot about that completely

src/neptune/handler.py
```diff
@@ -557,7 +557,14 @@ def fetch_last(self):
         """
         return self._pass_call_to_attr(function_name="fetch_last")
 
-    def fetch_values(self, *, include_timestamp: Optional[bool] = True, progress_bar: Optional[ProgressBarType] = None):
+    def fetch_values(
+        self,
+        *,
+        include_timestamp: Optional[bool] = True,
+        progress_bar: Optional[ProgressBarType] = None,
+        offset: Optional[int] = None,
```
Raalsky, line 565: There were comments on Jira to use `begin` and `end` instead. Like Python does for lists.

normandy7 (approved):

src/neptune/handler.py
szysad: oh it's in the Jira comment. I missed that 😨

src/neptune/attributes/series/fetchable_series.py
```diff
@@ -35,20 +35,24 @@
...
+        offset: Optional[int] = None,
+        limit: Optional[int] = None,
```
AleksanderWWW, line 52: We might want to add some value checks, because when I tried with a negative limit I got
```
neptune.common.exceptions.ClientHttpError: 

----ClientHttpError-----------------------------------------------------------------------

The Neptune server returned the status 400.

The server response was:
{"headers":{},"code":400,"errorType":"BAD_REQUEST","title":"Limit must not be negative"}
```

which is probably not the most user-friendly. (Similar case is for the `offset` param)

PR #1675, opened 2024-03-05
cache requirements

AleksanderWWW:

### DIFF
--- a/src/neptune/internal/utils/requirement_check.py
+++ b/src/neptune/internal/utils/requirement_check.py
@@ -25,7 +25,7 @@ from typing import Optional
 from neptune.exceptions import NeptuneMissingRequirementException
 
 
-@lru_cache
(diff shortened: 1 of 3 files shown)


src/neptune/internal/utils/requirement_check.py
```diff
@@ -18,12 +18,14 @@
...
 ]
 
+from functools import lru_cache
 from importlib.util import find_spec
 from typing import Optional
 
 from neptune.exceptions import NeptuneMissingRequirementException
 
 
+@lru_cache
```
Raalsky, line 28: If I remember we can provide an upper limit of how many values to store. Let's use something like `32` for instance.

AleksanderWWW: Thought of it too. Default is 128, but 32 should be enough too

Raalsky (approved):

PR #1680, opened 2024-03-08
Improve date parsing

AleksanderWWW:


AleksanderWWW:
## What this PR does

The java.time classes use [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) by default when parsing/generating textual representations of date-time values.

https://stackoverflow.com/questions/8405087/what-is-this-date-format-2011-08-12t201746-384z/36798830#36798830

Instead of hard-coding the date format we can use `datetime.fromisoformat` to parse the dates

pyproject.toml
```diff
@@ -22,6 +22,7 @@ future = ">=0.17.1"
 # Utility
 packaging = "*"
 click = ">=7.0"
+isodate = "*"
```
AleksanderWWW, line 25: `isodate` already is a dependency of at least one or more of our dependencies

szysad (commented):
What is the motivation behind this change?

Previously we we used standard library parsing using date format stored in our codebase. Now we use external library which does exact the same thing but now to see what format is expected we have to take a look at `isodate` docs

src/neptune/internal/utils/iso_dates.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return datetime.datetime.strptime(date, date_format)
+
+
+def _is_valid_iso_date(date_string: str) -> bool:
+    if len(date_string.split("T")) != 2 or not date_string.endswith("Z"):
+        return False
+
+    if len(date_string) < len("2022-01-01T00:00:00Z") or len(date_string) > len("2022-01-01T00:00:00.000000Z"):
+        return False
```
szysad, line 43: how about just try-except with
```
datetime.datetime.strptime(date, DATE_FORMAT)
```
current implementation returns True for `"::::::::::T::::::::Z"`

tests/unit/neptune/new/internal/utils/test_iso_dates.py
```diff
@@ -0,0 +1,79 @@
...
+        ("2022-01-01-00:00:00Z", False),  # date and time not separated by "T"
+        ("2022-01-01T00:00:00.000Z", True),
+        ("22-01-01T00:00:00.000000Z", False),  # wrong year format (not 4 digits)
+        ("2022-01-01T00:00:00.000000Z", True),
+        ("2022-01-01T00:00:00.000000000Z", False),  # too big precision
+    ],
+)
+def test_date_string_validation(date_test: Tuple[str, bool]):
+    date, expected = date_test
+    assert _is_valid_iso_date(date) == expected
```
szysad, line 79:
```suggestion
@pytest.mark.parametrize(
    ("date", "expected"),
    [
        ("2022-01-01", False),  # no time
        ("2022-01-01T00:00Z", False),  # no seconds
        ("2022-01-01T00:00:00", False),  # no "Z" at the end
        ("2022-01-01T00:00:00Z", True),
        ("2022-01-0100:00:00ZT", False),  # "T" in wrong place
        ("2022-01-01-00:00:00Z", False),  # date and time not separated by "T"
        ("2022-01-01T00:00:00.000Z", True),
        ("22-01-01T00:00:00.000000Z", False),  # wrong year format (not 4 digits)
        ("2022-01-01T00:00:00.000000Z", True),
        ("2022-01-01T00:00:00.000000000Z", False),  # too big precision
    ],
)
def test_date_string_validation(date: str, expected: bool):
    assert _is_valid_iso_date(date) == expected
```

Just a suggestion, maybe you like this version more 😄 (if you apply this version you have to get rid of `Tuple` import tho)

src/neptune/internal/utils/iso_dates.py
AleksanderWWW: Hmm, I see your point, but on the other hand we already catch an exception that would be produced by a string like yours in the part where we actually use the `parse_date` function

tests/unit/neptune/new/internal/utils/test_iso_dates.py
AleksanderWWW: Good point!

src/neptune/internal/utils/iso_dates.py
szysad: then maybe we don't need this function at all

AleksanderWWW: That's actually true probably :laughing:  now I don't really remember what I had in mind creating it

szysad (approved):
LGTM 🚀

PR #1681, opened 2024-03-08
Mypy import checks supressed for integrations

Raalsky:


szysad (commented):
how about just ignoring all import errors on integration
```
[[tool.mypy.overrides]]
module = "neptune.integrations.*"
ignore_missing_imports = true
```

That way we could ignore import errors for all integrations and confine whole change to pyproject.toml

Raalsky:
> That way we could ignore import errors for all integrations and confine whole change to pyproject.toml

Done. Looks that with `ignore_missing_imports` you have to provide external package that doesn't have stubs instead of a module that does the imports.

szysad (approved):
Cool idea to silence import errors on libraries and not our files!

PR #1692, opened 2024-03-13
fix sending from forked container

szysad:

- [ ] <s>Did you **ask the docs owner** to review all the user-facing changes?</s>

### DIFF
--- a/src/neptune/metadata_containers/metadata_container.py
+++ b/src/neptune/metadata_containers/metadata_container.py
@@ -256,8 +256,8 @@ class MetadataContainer(AbstractContextManager, NeptuneObject):
                     flush_period=self._flush_period,
                     queue=self._signals_queue,
                 ),
-                post_trigger_side_effect=self._op_processor.start,
             )
+            self._op_processor.set_post_trigger_side_effect(self._op_processor.start)
 
             # TODO: Every implementation of background job should handle fork by itself.
             jobs = []
--- a/src/neptune/metadata_containers/metadata_container.py
+++ b/src/neptune/metadata_containers/metadata_container.py
@@ -257,8 +257,6 @@ class MetadataContainer(AbstractContextManager, NeptuneObject):
                     queue=self._signals_queue,
                 ),
             )
-            self._op_processor.set_post_trigger_side_effect(self._op_processor.start)
-
             # TODO: Every implementation of background job should handle fork by itself.
             jobs = []
             if self._mode == Mode.ASYNC:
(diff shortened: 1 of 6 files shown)


szysad:
❓ [E2E](https://github.com/neptune-ai/neptune-client/actions/runs/8325442853)

src/neptune/metadata_containers/metadata_container.py
```diff
@@ -256,8 +256,8 @@ def _handle_fork_in_child(self):
                     flush_period=self._flush_period,
                     queue=self._signals_queue,
                 ),
-                post_trigger_side_effect=self._op_processor.start,
             )
+            self._op_processor.set_post_trigger_side_effect(self._op_processor.start)
```
Raalsky, line 260: Isn't it simpler to set it as a part of the constructor?

szysad: its actually a bit tricky to do it in a constructor. Previously we had
```
# self._op_processor was already defined
self._op_processor = LazyOperationProcessorWrapper(
                operation_processor_getter=partial(
                    get_operation_processor,
                    mode=self._mode,
                    container_id=self._id,
                    container_type=self.container_type,
                    backend=self._backend,
                    lock=self._lock,
                    flush_period=self._flush_period,
                    queue=self._signals_queue,
                ),
                post_trigger_side_effect=self._op_processor.start,
            )
```

so in line `post_trigger_side_effect=self._op_processor.start`
was `self._op_processor` was pointing to already defined `op_processor` like `AsyncOperationProcessor`.
which was also used by parent process (fork model) and therefore both processes would reference the same `consumer` thread which caused errors down the line.

Now `self._op_processor` points to `LazyOperationProcessorWrapper` when triggered will create new `AsyncOperationProcessor` instance with new `consumer` thread in fork child

Raalsky: You don't have to provide it as a parameter. You can set up it internally in the `LazyOperationProcessor` as:
`self._post_trigger_side_effect=self.start`

Raalsky: I know that you cannot provide it as a parameter because you don't have an instance created :wink:

szysad: In theory I could define `post_trigger_side_effect` as function that takes one argument - `self` of `LazyOperationProcessorWrapper` and work my way around that

Raalsky: I mean that you can set this property internally as a part of a constructor of `LazyOperationProcessorWrapper` like: `self._post_trigger_side_effect=self.start`. Think that it needs to be always called and it's an "atomic" operation that you have to initialize OperationProcessor and once it's created, run `start` on init. I've prepared some suggestions here: https://github.com/neptune-ai/neptune-client/pull/1705

Raalsky (approved):

PR #1703, opened 2024-03-20
Disable background jobs and metadata logging

AleksanderWWW:

### DIFF
--- a/src/neptune/objects/run.py
+++ b/src/neptune/objects/run.py
@@ -20,9 +20,11 @@ import threading
 from platform import node as get_hostname
 from typing import (
     TYPE_CHECKING,
+    Callable,
     List,
     Optional,
     Tuple,
+    TypeVar,
     Union,
 )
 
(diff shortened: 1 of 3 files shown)


src/neptune/objects/run.py
```diff
@@ -104,6 +106,21 @@
...
+T = TypeVar("T")
+
+
+def temporarily_disabled(func: Callable[..., T]) -> Callable[..., T]:
+    def wrapper(*_, **__):
+        if func.__name__ == "_get_background_jobs":
+            return []
+        elif func.__name__ == "_write_initial_attributes":
+            return None
+        elif func.__name__ == "_write_initial_monitoring_attributes":
```
szysad, line 118: I kinda don't like an idea of entangling `temporarily_disabled` with names of function from `Run`.
We could  add extra param `temporarily_disabled` to that defines return value so `Run` method signature does not change. But if its temporary it doesnt matter I guess

szysad (approved):

PR #1713, opened 2024-03-26
Fixes e2e and unittests

Raalsky:

### DIFF
--- a/src/neptune/objects/project.py
+++ b/src/neptune/objects/project.py
@@ -152,7 +152,8 @@ class Project(NeptuneObject):
         async_no_progress_callback: Optional[NeptuneObjectCallback] = None,
         async_no_progress_threshold: float = ASYNC_NO_PROGRESS_THRESHOLD,
     ):
-        raise NeptuneUnsupportedFunctionalityException
+        if mode in {Mode.ASYNC.value, Mode.SYNC.value}:
+            raise NeptuneUnsupportedFunctionalityException
 
         verify_type("mode", mode, (str, type(None)))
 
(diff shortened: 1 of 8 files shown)


Raalsky:
E2E: https://github.com/neptune-ai/neptune-client/actions/runs/8433534272

tests/e2e/standard/test_cli.py
szysad, line 1: There is another latent bug in line 267. We are expecting exception message to fit regex of literal `"Project can't be initialized in OFFLINE mode"` but it only holds for `Project` container.
It should be `"* can't be initialized in OFFLINE mode"` or `"(Project|Model|ModelVersion) can't be initialized in OFFLINE mode"` instead.

This will start failing after we stop skipping test for containers other then `Project`

src/neptune/objects/project.py
```diff
@@ -152,7 +152,8 @@ def __init__(
         async_no_progress_callback: Optional[NeptuneObjectCallback] = None,
         async_no_progress_threshold: float = ASYNC_NO_PROGRESS_THRESHOLD,
     ):
-        raise NeptuneUnsupportedFunctionalityException
+        if mode in {Mode.ASYNC.value, Mode.SYNC.value}:
+            raise NeptuneUnsupportedFunctionalityException
```
szysad, line 156: If I understand correctly we are now only allowing `read-only` and `debug` modes.
I guess it makes sense to disable Project only in cases it has to be connected to backend

szysad (approved):
Good one 🚀
It didn't come to my mind to only disable `Project` in online modes. This way we can still keep offline unit tests 💯

src/neptune/objects/project.py
Raalsky: + We do not support offline mode for anything except Runs.

PR #1730, opened 2024-04-10
Fixed compatibility checks with pre-release versions

Raalsky:

### DIFF
--- a/src/neptune/internal/backends/utils.py
+++ b/src/neptune/internal/backends/utils.py
@@ -276,18 +276,26 @@ def create_swagger_client(url: str, http_client: HttpClient) -> SwaggerClient:
 
 
 def verify_client_version(client_config: ClientConfig, version: Version):
+    base_version = Version(f"{version.major}.{version.minor}.{version.micro}")
     version_with_patch_0 = Version(replace_patch_version(str(version)))
-    if client_config.version_info.min_compatible and client_config.version_info.min_compatible > version:
+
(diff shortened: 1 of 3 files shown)


src/neptune/internal/backends/utils.py
```diff
@@ -276,18 +276,26 @@ def create_swagger_client(url: str, http_client: HttpClient) -> SwaggerClient:
...
+
+    min_compatible = client_config.version_info.min_compatible
+    max_compatible = client_config.version_info.max_compatible
+    min_recommended = client_config.version_info.min_recommended
+
+    if min_compatible and min_compatible > base_version:
         raise NeptuneClientUpgradeRequiredError(version, min_version=client_config.version_info.min_compatible)
-    if client_config.version_info.max_compatible and client_config.version_info.max_compatible < version_with_patch_0:
+
+    if max_compatible and max_compatible < version_with_patch_0:
```
szysad, line 289: Why when checking whether client version is too big we extend possible `micro` versions?

Ex. if `version = 1.5.10` and `max_compatible = 1.5.0` we are not raising exception

szysad (approved):

src/neptune/internal/backends/utils.py
Raalsky: I don't know :<. Maybe someone set `0.10.0` instead of `0.10.9999` for some already delivered self hosted environment and that was easier to do it in client.

PR #1745, opened 2024-04-16
Remove pillow from dependencies

AleksanderWWW:

### DIFF
--- a/tests/e2e/utils.py
+++ b/tests/e2e/utils.py
@@ -38,15 +38,11 @@ from math import sqrt
 from time import perf_counter
 
 import numpy
-import pytest
 from attr import dataclass
 
 import neptune
 from neptune.internal.container_type import ContainerType
 from tests.e2e.exceptions import MissingEnvironmentVariable
-
-Image = pytest.importorskip("PIL.Image")
-PngImageFile = pytest.importorskip("PIL.PngImagePlugin.PngImageFile")
 
 
 def _remove_file_if_exists(filepath):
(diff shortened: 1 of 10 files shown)


tests/unit/neptune/new/attributes/series/test_file_series.py
szysad, line 1: why are we marking those tests as `xfail`. Shouldn't we skip if PIL not installed?

tests/unit/neptune/new/types/atoms/test_file.py
```diff
@@ -166,6 +165,8 @@ def test_stream_read_once(self):
             _ = file.content
 
     def test_as_image(self):
+        from PIL import Image
+
```
szysad, line 169: shouldn't PIL be loaded using `pytest.importskip` here too?

AleksanderWWW: Good point!

AleksanderWWW:
## E2E :hourglass_flowing_sand: 
https://github.com/neptune-ai/neptune-client/actions/runs/8717801814

tests/e2e/utils.py
szysad, line 1: whats the reason for getting rid of `PIL importskip` in favor of dynamic import

szysad (approved):

szysad (changes requested):
One more important thing. Shouldn't `Pillow` be now added to `dev_requirements.txt` so tests that use `Pillow` are not skipped?

tests/e2e/utils.py
AleksanderWWW: `importskip` seems to work only in test files that are not imported later on. Here's a `utils` file that's not directly tested. The result of using `importskip` here was an error:
https://github.com/neptune-ai/neptune-client/actions/runs/8717801814/job/23914764348

szysad (approved):

PR #1750, opened 2024-04-16
Added dependabot for neptune-client package

Raalsky:


.github/dependabot.yml
```diff
@@ -1,9 +1,18 @@
...
+
+  # New `neptune` package
   - package-ecosystem: pip
     directory: "/"
     schedule:
       interval: "weekly"
+
+  # Old `neptune-client` package
+  - package-ecosystem: pip
+    directory: ".github/"
```
AleksanderWWW, line 12: Why is the directory `.github` here and in the new neptune it's `/`?

Raalsky: That's where we're storing the package definition of `neptune-client` :D. Take a look at release workflow :wink:

Raalsky: Currently Dependabot is looking for only `neptune` package (probably) as only our `pyproject.toml` is detected from the root directory. We would like to take into account `neptune-client` as well that was stored in `.github` directory.

AleksanderWWW (approved):

PR #1751, opened 2024-04-17
Endpoints added for enchanced fields search

Raalsky:

### DIFF
--- a/src/neptune/api/pagination.py
+++ b/src/neptune/api/pagination.py
@@ -1,0 +1,60 @@
+#
+# Copyright (c) 2024, Neptune Labs Sp. z o.o.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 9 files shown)


src/neptune/api/pagination.py
```diff
@@ -0,0 +1,60 @@
...
+class WithPagination(abc.ABC):
+    next_page: Optional[NextPage]
+
+
+T = TypeVar("T", bound=WithPagination)
+Entry = TypeVar("Entry")
+
+
+class Paginatable(Protocol):
+    def __call__(self, *, next_page: Optional[NextPage] = None, **kwargs: Any) -> Any: ...
```
AleksanderWWW, line 44: If there is an assumption that a `Paginatable` will always return an `Iterator` of some kind, we could make the type hint more precise. Is that correct, or we cannot guarantee that?

AleksanderWWW (approved):

src/neptune/api/pagination.py
Raalsky: That was my initial try but I was struggling with it. You cannot create a `Protocol` with `Generic` at the same time and you have to provide a type var as an input if you would like to have something as an output. You can give it a try if you want.

PR #1755, opened 2024-04-24
Mark run as failed in case of error using ctx manager

AleksanderWWW:

### DIFF
--- a/.github/workflows/unit-in-pull-request.yml
+++ b/.github/workflows/unit-in-pull-request.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches-ignore:
-      - dev/1.x
+      - master
 
 jobs:
   test:
@@ -12,8 +12,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu, windows, macos]
+        os: [ubuntu, windows]
         python-version: ["3.7"]
+        include:
+          - python-version: "3.8"
+            os: macos
     name: 'test (${{ matrix.os }} - py${{ matrix.python-version }})'
     runs-on: ${{ matrix.os }}-latest
     steps:
--- a/.github/workflows/unit-in-pull-request.yml
+++ b/.github/workflows/unit-in-pull-request.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches-ignore:
-      - master
+      - dev/1.x
 
 jobs:
   test:
@@ -17,7 +17,6 @@ jobs:
         include:
           - os: macos
             python-version: "3.8"
-
     name: 'test (${{ matrix.os }} - py${{ matrix.python-version }})'
     runs-on: ${{ matrix.os }}-latest
     steps:
--- a/.github/workflows/unit-in-pull-request.yml
+++ b/.github/workflows/unit-in-pull-request.yml
@@ -15,8 +15,9 @@ jobs:
         os: [ubuntu, windows]
         python-version: ["3.7"]
         include:
-          - python-version: "3.8"
-            os: macos
+          - os: macos
+            python-version: "3.8"
+
     name: 'test (${{ matrix.os }} - py${{ matrix.python-version }})'
     runs-on: ${{ matrix.os }}-latest
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
+## [UNRELEASED] neptune 1.10.4
+
+### Fixes
+- Fixed run not failing in case of an exception if context manager was used ([#1755](https://github.com/neptune-ai/neptune-client/pull/1755))
+
+
 ## neptune 1.10.3
 
 ### Fixes
 - Clarified the `autoscale` parameter description in the `as_image()` docstring ([#1742](https://github.com/neptune-ai/neptune-client/pull/1742))
-- Fixed run not failing in case of an exception if context manager was used ([#1755](https://github.com/neptune-ai/neptune-client/pull/1755))
 
 ### Changes
 - Neptune now shows a warning instead of an error when the dependency tracking file is not found ([#1739](https://github.com/neptune-ai/neptune-client/pull/1739))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [UNRELEASED] neptune 1.10.3
+## neptune 1.10.3
 
 ### Fixes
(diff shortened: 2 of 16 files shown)


AleksanderWWW:
## What this PR does

The issue is that the `__exit__` method in the `MetadataContainer` takes precedence over the traceback background job. 
The result is that this:
```
run = neptune.init_run()
raise some_exception
```
marks the run as failed, but this
```
with neptune.init_run():
    raise some_exception
```
does not, because the `__exit__` doesn't contain any logic that would fail the run.

The proposition is to add the same logic as is present in the traceback job, but to avoid repeating the logic - move it to the
`MetadataContainer` instead and use in the traceback job from there, instead of creating it in that place locally.

src/neptune/metadata_containers/metadata_container.py
```diff
@@ -703,3 +706,9 @@ def _fetch_entries(
     def get_root_object(self) -> "MetadataContainer":
         """Returns the same Neptune object."""
         return self
+
+    def log_traceback(self, stacktrace_lines: List[str]) -> None:
```
Raalsky, line 710: I'm not a big fan of this approach. We moved the logic from background job to NeptuneObject but the background jobs should be rather opt-in and depends on the interface that NeptuneObject provides. Maybe we should prepare a way to trigger the handler in addition to `print_exception`

src/neptune/metadata_containers/metadata_container.py
```diff
@@ -315,7 +314,7 @@ def _write_initial_attributes(self):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if exc_tb is not None:
-            traceback.print_exception(exc_type, exc_val, exc_tb)
+            uncaught_exception_handler.trigger(exc_type, exc_val, exc_tb)
```
Raalsky, line 317: Will `traceback.print_exception` still be covered?

src/neptune/internal/utils/uncaught_exception_handler.py
```diff
@@ -40,20 +40,23 @@ def __init__(self):
...
+
+        traceback_lines = header_lines + traceback.format_tb(exc_tb) + str(exc_val).split("\n")
+        for _, handler in self._handlers.items():
+            handler(traceback_lines)
+
     def activate(self):
         with self._lock:
             this = self
 
             def exception_handler(exc_type, exc_val, exc_tb):
```
Raalsky, line 58: Maybe we should use `partial` instead to inject `self`?

src/neptune/metadata_containers/metadata_container.py
AleksanderWWW: I'm not sure it needs to be tbh - we want to have the same output whether we're using context manager or not, right?

SiddhantSadangi: In case of an error in a context, nothing is being printed to `stderr`.
Expected behaviour?
image

Raalsky: So let's trigger the uncaught exception handler + leave print_exception as it was :wink:

SiddhantSadangi (approved):

.github/workflows/unit-in-pull-request.yml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches-ignore:
-      - dev/1.x
+      - master
```
Raalsky, line 7: ?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
```
Raalsky, line 2: Rebase needed

src/neptune/internal/utils/uncaught_exception_handler.py
```diff
@@ -76,4 +74,10 @@ def unregister(self, uid: uuid.UUID):
                 del self._handlers[uid]
 
 
+def exception_handler(uncaught_exception_handler: UncaughtExceptionHandler, exc_type, exc_val, exc_tb):
```
Raalsky, line 77: - Instead of `exc_type, exc_val, exc_tb` let's use `*args` and `**kwargs`
- Let's make it a class method instead, you'll get `self` without a need to call `partial` and an access to internal property `_previous_uncaught_exception_handler` in a more proper way. Previously we had an inner function, right?

src/neptune/internal/utils/uncaught_exception_handler.py
```diff
@@ -19,6 +19,7 @@
 import threading
 import traceback
 import uuid
+from functools import partial
 from platform import node as get_hostname
 from typing import (
     TYPE_CHECKING,
```
Raalsky, line 25: `TYPE_CHECKING` is not needed in this module :wink:

src/neptune/internal/utils/uncaught_exception_handler.py
```diff
@@ -76,4 +74,10 @@ def unregister(self, uid: uuid.UUID):
                 del self._handlers[uid]
 
 
+def exception_handler(uncaught_exception_handler: UncaughtExceptionHandler, exc_type, exc_val, exc_tb):
+    uncaught_exception_handler.trigger(exc_type, exc_val, exc_tb)
+
+    uncaught_exception_handler._previous_uncaught_exception_handler(exc_type, exc_val, exc_tb)
```
Raalsky, line 80: Let's add a validation that it isn't `None`; I think that this would be detected by `mypy`, don't you wanna add proper typings? Doesn't look to be demanding :wink:

src/neptune/internal/utils/uncaught_exception_handler.py
```diff
@@ -21,59 +21,59 @@
...
+        for _, handler in self._handlers.items():
+            handler(traceback_lines)
 
-            if self._previous_uncaught_exception_handler is None:
-                self._previous_uncaught_exception_handler = sys.excepthook
-                sys.excepthook = exception_handler
+    def activate(self) -> None:
+        with self._lock:
+            if self._previous_uncaught_exception_handler is not None:
+                return
```
Raalsky, line 54: I had to verify it but looks like `return` could be called safely from context-manager: https://stackoverflow.com/a/9885287

src/neptune/internal/utils/uncaught_exception_handler.py
```diff
@@ -21,59 +21,59 @@
 import uuid
```
Raalsky, line 21: Shouldn't we remove an entry from `pyproject.toml` after applying mypy suggestions?

SiddhantSadangi (approved):

Raalsky (approved):

PR #1761, opened 2024-05-06
change handling of too long custom run id

szysad:
when too long raise NeptuneException
max length increased to 128

### DIFF
--- a/src/neptune/objects/run.py
+++ b/src/neptune/objects/run.py
@@ -47,6 +47,7 @@ from neptune.envs import (
 from neptune.exceptions import (
     InactiveRunException,
     NeedExistingRunForReadOnlyMode,
+    NeptuneException,
     NeptuneRunResumeAndCustomIdCollision,
 )
 from neptune.internal.backends.api_model import ApiExperiment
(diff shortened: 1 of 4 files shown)


src/neptune/objects/run.py
```diff
@@ -432,7 +436,9 @@ def _get_or_create_api_object(self) -> ApiExperiment:
 
             custom_run_id = self._custom_run_id
             if custom_run_id_exceeds_length(self._custom_run_id):
-                custom_run_id = None
+                raise NeptuneException(
+                    f"Given custom_run_id exceeds {CUSTOM_RUN_ID_LENGTH} characters and it will be ignored."
```
AleksanderWWW, line 440: Not sure about the message here. It says that it will be ignored, as if the program was about to continue, but in fact it will throw an exception and stop working. Maybe you could comment @normandy7 ? :smiley:

szysad: good point

szaganek: Substitute chimes in: 
I also understand _will be ignored_ as _will continue anyway_. I checked other errors/exceptions and I think there's this pattern to describe them (with backticks):

``"Parameter `custom_run_id` exceeds {CUSTOM_RUN_ID_LENGTH} characters."``

AleksanderWWW: I think this would be enough for the exception message :)

szysad: @szaganek Ignoring policy won't be continued as we are rising top level exception

szaganek:
```suggestion
                    f"Parameter `custom_run_id` exceeds {CUSTOM_RUN_ID_LENGTH} characters."
```

AleksanderWWW (approved):

szaganek (approved):

PR #1762, opened 2024-05-07
Generate custom run id if not provided

AleksanderWWW:

### DIFF
--- a/src/neptune/objects/run.py
+++ b/src/neptune/objects/run.py
@@ -17,6 +17,7 @@ __all__ = ["Run"]
 
 import os
 import threading
+import uuid
 from platform import node as get_hostname
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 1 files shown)


AleksanderWWW:
In addition I saw that the unit tests for `core.operation_processors` had been misplaced - I moved them to the correct place.

src/neptune/objects/run.py
```diff
@@ -406,6 +407,9 @@ def __init__(
         if mode == Mode.OFFLINE or mode == Mode.DEBUG:
             project = OFFLINE_PROJECT_QUALIFIED_NAME
 
+        if self._custom_run_id is None and mode != Mode.READ_ONLY:
```
szysad, line 410: why when in read only custom run id is not generated?

szysad (approved):
LGTM

src/neptune/objects/run.py
AleksanderWWW: We want to generate it only if a run is actually created - refer to the task description

PR #1770, opened 2024-05-08
init code for async op processor

AleksanderWWW:


src/neptune/core/operation_processors/async_operation_processor.py
AleksanderWWW, line 1: I'm thinking - do you think it's worth to create a separate package for async op processor and split this file up? It is a bit long

src/neptune/core/operation_processors/operation_logger.py
AleksanderWWW, line 1: This is copy-paste from `internal.operation_processors.operation_logger.py` + one method added - `set_processor_stop_signal_queue`

src/neptune/core/operation_processors/async_operation_processor.py
```diff
@@ -0,0 +1,430 @@
...
+        self.close()
+
+        # Remove local files
+        if self._queue_observer.is_queue_empty():
+            self.cleanup()
+
+    def cleanup(self) -> None:
+        super().cleanup()
+        try:
+            self._processing_resources.data_path.rmdir()
```
szysad, line 195: I think if you introduce new abstraction like `ProcessingResources` which itself is `WithResource` instance it makes more sense to delegate cleaning of path to `ProcessingResources` itself.

So just make `ProcessingResources` implement `cleanup` method in which it removes data path. And in `cleanup` of `AsyncOperationProcessor` just call `self._processing_resources.clenup()`

Also if you defined `processing_resources` method maybe its good idea to use it here.

(sorry for mansplaining lol)

szysad (changes requested):

src/neptune/core/operation_processors/async_operation_processor.py
AleksanderWWW: Good point

szysad (approved):

src/neptune/core/operation_processors/async_operation_processor/processing_resources.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.consumed_version: int = 0
+
+    @property
+    def resources(self) -> Tuple[Resource, ...]:
+        return self.metadata_file, self.operation_storage, self.disk_queue
+
+    @property
+    def data_path(self) -> Path:
+        return self._data_path
```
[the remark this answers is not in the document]
AleksanderWWW: this would change the signature of an interface method (`Resource`) + how it's right now is also the way it's handled in the original `AsyncOperationProcessor`

PR #1774, opened 2024-05-13
run workflow from custom branch

szysad:

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -19,7 +19,7 @@ jobs:
   set_neptune_ref:
     uses: ./.github/workflows/set_neptune_ref.yml
     with:
-      neptune_ref: ${{ inputs.neptune_ref }}
+      workflow_call_input_neptune_ref: ${{ inputs.neptune_ref }}
   print_inputs:
     needs: [ set_neptune_ref ]
     runs-on: ubuntu-latest
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -38,14 +38,15 @@ jobs:
       - name: Read debug variables
         run: |
           echo "github.event_name=${{ github.event_name }}"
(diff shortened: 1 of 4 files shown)


.github/workflows/unit.yml
```diff
@@ -2,6 +2,12 @@ name: unit
 
 on:
   workflow_call:
+    inputs:
```
Raalsky, line 5: Shouldn't it be workflow-level? For `worflow_dispatch` and `schedule` this won't work as far as I can understand.

szysad: Our goal here is to make those workflows reusable and then later on reuse them in other workflows that are getiting triggered manualy by schedule or any other way.
`workflow_call` makes workflow reusable and can be called with custom input.
[Example from my other branch ](https://github.com/neptune-ai/neptune-client/blob/4438f30304d72756fb59094e78d1d37ffd48499d/.github/workflows/e2e.yml#L174)

Raalsky: So let's remove the `schedule` and `workflow_dispatch` from `dev/1.x` then. This workflow will only be triggered externally or via a `push`.

This will be broken for `push` to `dev/1.x` as `inputs.neptune_ref` will not be provided as far as I can undestand.

szysad: > This will be broken for push to `dev/1.x` as `inputs.neptune_ref` will not be provided as far as I can undestand.
For now `neptune_ref` defaults to `${{ github.event.client_payload.pull_request.head.ref }}` if not provided but I guess it will brake on push

Raalsky: I think that this will not be set for `push` as the `inputs` default was added under "workflow_call". Shouldn't we somehow add a global inputs settings or use some inputs getter with a default?

szysad: Yes I think that is great idea but I'd argue its out of scope for this task

Raalsky (approved):

PR #1776, opened 2024-05-15
Docstring for pop()

normandy7:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Move some `OperationProcessor` implementations to `neptune.core.operation_processors` ([#1760](https://github.com/neptune-ai/neptune-client/pull/1760))
 - Changed handling of too long custom run id ([#1761](https://github.com/neptune-ai/neptune-client/pull/1761))
 - Skip tests that require `GitPython` when not installed ([#1752](https://github.com/neptune-ai/neptune-client/pull/1752))
+- Added docstring for the `pop()` function ([#1776](https://github.com/neptune-ai/neptune-client/pull/1776))
 
 ### Fixes
 - Fixed `tqdm.notebook` import only in Notebook environment ([#1716](https://github.com/neptune-ai/neptune-client/pull/1716))
--- a/src/neptune/handler.py
+++ b/src/neptune/handler.py
@@ -736,6 +736,20 @@ class Handler(SupportsNamespaces):
     @feature_temporarily_unavailable
     @check_protected_paths
     def pop(self, path: str = None, *, wait: bool = False) -> None:
+        """Completely removes the namespace and all associated metadata stored under the path.
+
+        Args:
+            path: Path of the namespace to be removed.
+            wait: By default, logged metadata is sent to the server in the background.
+                With this option set to `True`, Neptune first sends all data before executing the call.
+
+        Example:
+            >>> import neptune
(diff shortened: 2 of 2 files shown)


src/neptune/handler.py
```diff
@@ -736,6 +736,20 @@ def __delitem__(self, path) -> None:
     @feature_temporarily_unavailable
     @check_protected_paths
     def pop(self, path: str = None, *, wait: bool = False) -> None:
+        """Completely removes the namespace and all associated metadata stored under the path.
+
+        Args:
+            path: Path of the namespace to be removed.
+            wait: By default, logged metadata is sent to the server in the background.
+                With this option set to `True`, Neptune first sends all data before executing the call.
```
szaganek, line 744:
```suggestion
                With this option set to `True`, Neptune sends all data before executing the call.
```
Avoids redundancy. Or `..., Neptune first sends all data, then executes the call`.

normandy7: This parameter is used in numerous places, so we have a lot of instances to update :)

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 - Move some `OperationProcessor` implementations to `neptune.core.operation_processors` ([#1760](https://github.com/neptune-ai/neptune-client/pull/1760))
 - Changed handling of too long custom run id ([#1761](https://github.com/neptune-ai/neptune-client/pull/1761))
 - Skip tests that require `GitPython` when not installed ([#1752](https://github.com/neptune-ai/neptune-client/pull/1752))
+- Added docstring for the `pop()` function ([#1776](https://github.com/neptune-ai/neptune-client/pull/1776))
```
Raalsky, line 40: Would you mind to cherry pick this changes to `dev/1.x` as well?

src/neptune/handler.py
```diff
@@ -736,6 +736,20 @@ def __delitem__(self, path) -> None:
     @feature_temporarily_unavailable
     @check_protected_paths
     def pop(self, path: str = None, *, wait: bool = False) -> None:
+        """Completely removes the namespace and all associated metadata stored under the path.
+
+        Args:
+            path: Path of the namespace to be removed.
+            wait: By default, logged metadata is sent to the server in the background.
```
Raalsky, line 743: `wait` seems not to express the effect properly in the context of `pop`. We will wait to all metadata to be removed :wink:

normandy7: Isn't it technically the case that with `wait=True`, any queued operations (which may include earlier logging calls) are completed before continuing? And not just this particular `pop()` operation.

Conversely, if `wait` is used in some call after `pop()`, do we not have the same phrasing problem?

```
run["huge_field"].pop()
run["a"].assign("aaa", wait=True)
```

I.e. in this case we'll technically wait for deletion rather than sending of metadata. Just wondering if we should adjust this wording globally.

Raalsky: From a technical point of view, you're absolutely right 😉 . I've started to feel that it looks like being copy-pasted and doesn't match the actual use-case but I'm good as long as it's ok for you.

Raalsky (dismissed):

src/neptune/handler.py
normandy7: Thanks for confirming 😃 We'll think of some more accurate and helpful phrasing in a separate update.

Raalsky (approved):

PR #1789, opened 2024-05-21
Use environment for e2e and integrations workflow

Raalsky:

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -13,12 +13,12 @@ on:
       - dev/1.x
 
 env:
-  WORKSPACE_NAME: e2e-tests
+  WORKSPACE_NAME: ${{ vars.E2E_WORKSPACE_NAME }}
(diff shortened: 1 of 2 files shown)


szysad (approved):
LGTM!

There are two things that I'm not 100% sure about.

1. in [dev1.x_e2e.yml on master](https://github.com/neptune-ai/neptune-client/blob/master/.github/workflows/dev1.x-e2e.yml) (and other similar files) we have secrets referenced without any envs. I don't have access to secrets but those references should stay valid.
2. again in [dev1.x_e2e.yml on master](https://github.com/neptune-ai/neptune-client/blob/master/.github/workflows/dev1.x-e2e.yml) we have `secrets: inherit` I'm not sure if this will override any new secrets in `dev/1.x` with conficting names

Raalsky:
> There are two thinks that I'm not 100% sure about.

You're right and I was thinking about that. As of Environments introduction it's probably not needed anymore. Let's address that in a separate PR if needed.

PR #1790, opened 2024-05-21
Use Github Environments and remove integrations workflow

Raalsky:


.github/workflows/e2e.yml
```diff
@@ -74,8 +77,6 @@ jobs:
 
       - name: Test
         uses: ./.github/actions/test-e2e
-        env:
-          NEPTUNE_API_TOKEN: ${{secrets.E2E_NEPTUNE_API_TOKEN}}
```
szysad, line 78: why `e2e_management` doesn't need `NEPTUNE_API_TOKEN` all of the sudden?

szysad (approved):
why are we getting rid of `e2e_s3_gcs`, `e2e_s3` jobs and removing `integrations` workflow entirely.

.github/workflows/e2e.yml
Raalsky: It was moved to a common `env` section. And I wanted to made it a way cleaner which API key are we going to use.

PR #1816, opened 2024-06-17
Management functionality removed temporarly

Raalsky:


src/neptune/internal/backends/hosted_neptune_backend.py
```diff
@@ -273,30 +272,22 @@ def get_metadata_container(
...
     ) -> ApiExperiment:
-        try:
-            experiment = (
-                self.leaderboard_client.api.getExperiment(
-                    experimentId=container_id,
-                    **DEFAULT_REQUEST_KWARGS,
-                )
-                .response()
-                .result
+        experiment = (
```
AleksanderWWW, line 275: If I understand correctly, this could now throw a `RuntimeError` - don't we want to catch it and throw the domain exception instead?

Raalsky: As far as I can understand, we've removed management functionality so those errors will never occur so we don't have to worry about them for now.

AleksanderWWW (approved):

PR #1873, opened 2024-09-19
Emit `UnknownField` for unsupported fields

kgodlewski:


src/neptune/api/field_visitor.py
```diff
@@ -39,12 +39,16 @@
     StringField,
     StringSeriesField,
     StringSetField,
+    UnknownField,
 )
 from neptune.exceptions import MetadataInconsistency
 
 
 class FieldToValueVisitor(FieldVisitor[Any]):
```
Raalsky, line 47: `FieldVisitor` should be updated as well

src/neptune/api/models.py
```diff
@@ -150,7 +152,17 @@ def from_model(model: Any) -> Field:
     @staticmethod
     def from_proto(data: Any) -> Field:
         field_type = str(data.type)
-        return Field._registry[field_type].from_proto(data.__getattribute__(f"{camel_to_snake(field_type)}_properties"))
+
+        try:
+            # We might receive a field that is not (fully) supported
+            if field_type not in Field._registry:
```
Raalsky, line 158: Instead of `try..catch` you should extend this if statement with `hasattr` and just return `UnknownField`

src/neptune/api/models.py
```diff
@@ -150,7 +152,17 @@ def from_model(model: Any) -> Field:
...
+                raise AttributeError
+
+            attr = data.__getattribute__(f"{camel_to_snake(field_type)}_properties")
+        except AttributeError:
+            return UnknownField(path=data.name)
```
Raalsky, line 163: We've added `from_proto` method in `UnknownField` already. Something fishy here

src/neptune/api/field_visitor.py
kgodlewski: We have a `FieldVisitor.visit_unknown()`

PR #1897, opened 2024-12-27
PPY-68: Added multithreading support to `neptune sync`

SiddhantSadangi:

- [X] Did you **update the CHANGELOG**? (not for test updates, internal changes/refactors or CI/CD setup)
- [X] Did you **ask the docs owner** to review all the user-facing changes?

### DIFF
--- a/src/neptune/cli/sync.py
+++ b/src/neptune/cli/sync.py
@@ -16,6 +16,7 @@
 
 __all__ = ["SyncRunner"]
 
+import concurrent.futures
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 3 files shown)


src/neptune/cli/sync.py
```diff
@@ -50,60 +51,122 @@
...
+                        backend=backend,
+                        base_path=base_path,
+                        container_names=[selected_offline],
+                        containers=containers.offline_containers,
+                        project_name=project_name,
+                    )
+                    for selected_offline in offline_selected
+                ]
+            for future in concurrent.futures.as_completed(futures):
+                future.result()
```
AleksanderWWW, line 169: I feel that this exact piece of logic (with slightly different parameters each time) is repeated so often it deserves a separate function ;)

src/neptune/cli/sync.py
```diff
@@ -50,60 +51,122 @@
...
+        project_name: Optional[str] = None,
+        num_threads: Optional[int] = None,
+    ) -> None:
         containers = collect_containers(path=base_path, backend=backend)
 
-        project = get_project(project_name_flag=QualifiedName(project_name) if project_name else None, backend=backend)
+        project = get_project(
+            project_name_flag=QualifiedName(project_name) if project_name else None,
+            backend=backend,
+        )
```
AleksanderWWW, line 66: I assume this change is related to automatic formatting, right?

src/neptune/cli/sync.py
```diff
@@ -50,60 +51,122 @@
...
         if not project:
             raise CannotSynchronizeOfflineRunsWithoutProject
 
-        for container in containers.offline_containers:
-            container.sync(base_path=base_path, backend=backend, project=project)
+        with concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as executor:
```
AleksanderWWW, line 70: What if `num_threads` is `None`? I think we should perform an ordinary loop then, right?

AleksanderWWW (commented):
Thing to consider here, no strong opinion:

Python is anyway single-threaded and those operations are IO-bound I think, so perhaps multi-threading approach should be replaced with concurrency and asyncio?
@PatrykGala

src/neptune/cli/sync.py
SiddhantSadangi: yup

src/neptune/cli/sync.py
SiddhantSadangi: `ThreadPoolExecutor` uses [it's defaults](https://getdocs.org/Python/docs/3.10/library/concurrent.futures#:~:text=Changed%20in%20version%203.8%3A%20Default%20value%20of%20max_workers,for%20CPU%20bound%20tasks%20which%20release%20the%20GIL.) if `num_workers` is `None`

src/neptune/cli/sync.py
SiddhantSadangi: Will create a reusable function if we stick with this approach :)

src/neptune/cli/sync.py
AleksanderWWW: So what if a user doesn't want to use threads?

SiddhantSadangi: `neptune sync -n 1` ;)

AleksanderWWW: But it's anyway a design choice that the team should discuss internally. Save for that and the other points LGTM ;)

